Issue #129, opened 2017-01-10
Faster transportation_names


stirringhalo:
In light of https://github.com/openmaptiles/openmaptiles/issues/114 and how it takes 9 hours to rematerialize the transportation_names layer, I went looking for a solution to find an SQL nearly as good looking for much less time.

The work is in https://github.com/openmaptiles/openmaptiles/tree/faster_transportation_name, and initial results suggests massive speed improvements for the import-sql stage, but rendering is slowed (about half speed). I am investigating the rendering end now.

**Switzerland, (note the slow method is into a materialized view, the fast method is into a table, but the performance holds up)**
Slow method:
SELECT 235496
Time: 18775.231 ms

Fast method:
SELECT 119097
Time: 3055.946 ms

The SELECT count difference is due to the fact that the fast method ends up with ST_MultiLineString while the slow method ends up with ST_LineString. I can duplicate that behaviour in the fast method. That may also be a cause of the slow rendering.

It might be possible to get the fast method to handle planet updates in under one hour, I will compare with a larger import and see.

stirringhalo:
For France (with work_mem set to 5GB):
Slow method: 119349.826 ms
Fast method: 6031.045 ms

So massive gains possible when it fits in memory. We really ought to tune the postgis install...

stirringhalo:
@lukasmartinelli 
Question, how critical is transportation_name and can the layer simply be dropped? It feels silly to have nearly the same information, twice, and make adjusting the style based on label preference harder. Like for example, one cannot label railroads as those are not included in transportation_names.

I figure leave the layer out and let mapbox-gl-js sort labels on transportation out. Saves DB update time, tile size and rendering time. Adds slight complexity to the styles.

lukasmartinelli:
> Question, how critical is transportation_name and can the layer simply be dropped? It feels silly to have nearly the same information, twice, and make adjusting the style based on label preference harder. Like for example, one cannot label railroads as those are not included in transportation_names.

The stitching together of the lines is necessary so we get the longest possible lines for roads with the same names. This in turn then helps Mapbox GL JS place the labels with regards to the padding/space between labels. Otherwise it happens that at every turn/crossing there are new linestrings and therefore each road is labelled individually.

..that's not always such a big deal. It really appears when trying to label highway refs - which will result in a forest of labels without stitched together motorways.

[...]

PR #124, opened 2017-01-07
[WIP] imposm3 mapping change to the new reject/require syntax

ImreSamu:
Work in progress ..
- [x] rewrite to the new imposm3 filter syntax  ..
- [x] testing   -   with a  `import-osm:01` :   same record number    :  OK  
- [ ] testing ....

works with:  https://github.com/openmaptiles/import-osm/pull/6

PR #131, opened 2017-01-12
Daily updates

stirringhalo:
This PR enables daily updates when omniscale/imposm3 is used. Right now it doesn't reflect the syntax of https://github.com/openmaptiles/openmaptiles/pull/124 but this PR should not impact current use of OpenMapTiles with our fork of Imposm3.

Once this is merged, I will look into faster refresh (https://github.com/openmaptiles/openmaptiles/issues/129) where it may be possible to get down to hourly with parallel SQL updates.

### DIFF
--- a/layers/housenumber/housenumber_centroid.sql
+++ b/layers/housenumber/housenumber_centroid.sql
@@ -1,4 +1,9 @@
 
 -- etldoc: osm_housenumber_point -> osm_housenumber_point
-UPDATE osm_housenumber_point SET geometry=topoint(geometry)
-WHERE ST_GeometryType(geometry) <> 'ST_Point';
+CREATE FUNCTION convert_housenumber_point() RETURNS VOID AS $$
+BEGIN
+  UPDATE osm_housenumber_point SET geometry=topoint(geometry) WHERE ST_GeometryType(geometry) <> 'ST_Point';
+END;
+$$ LANGUAGE plpgsql;
+
+SELECT convert_housenumber_point();
--- a/layers/housenumber/housenumber_centroid.sql
+++ b/layers/housenumber/housenumber_centroid.sql
@@ -7,3 +7,37 @@ END;
 $$ LANGUAGE plpgsql;
 
 SELECT convert_housenumber_point();
+
+-- Handle updates
+
+CREATE SCHEMA housenumber;
+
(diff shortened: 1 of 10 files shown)


stirringhalo:
It took about 11 hours to do a daily update, most of which was spent refreshing transportation_names layer.

ImreSamu:
@stirringhalo :   Probably I can test at the weekend, not early ..   sorry

lukasmartinelli:
This has been a huge effort from what I see. Respect :clap: 

To generalize:

1. This recreates a trigger each time at insert time
2. Then it creates the derived tables/runs the updates
3. This trigger is bound to each modification of the derived table (INSERT/UPDATE/DELETE) but only executed after each commit?
4. The trigger then refreshes the table when imposm3 imports

So after these PR's we can guarantee the following for derived tables

1. Each time we run `import-sql` the derived tables get created (but only the first time right?)
2. Each time we run import the derived tables now get automatically updated :+1: 

Did I understand this correctly?

stirringhalo:
I should've described how this work!

> To generalize:
>
> 1.    This recreates a trigger each time at insert time
> 2.    Then it creates the derived tables/runs the updates
> 3.    This trigger is bound to each modification of the derived table (INSERT/UPDATE/DELETE) but only executed after each commit?
> 4.    The trigger then refreshes the table when imposm3 imports
>
> So after these PR's we can guarantee the following for derived tables
>
> 1.   Each time we run import-sql the derived tables get created (but only the first time right?)
> 2.   Each time we run import the derived tables now get automatically updated :+1:
>
> Did I understand this correctly?

What you do is run the import-osm, import-sql sequence as normal.

import-sql creates the triggers, flag and update functions as well as the triggers.

Afterwards, you run the update-osm which you have in the  compose_update branch. That runs continuously.

On each INSERT/UPDATE/DELETE, the .flag() function gets executed, which UPSERTs into the .updates table when an update is needed. Only one row gets inserted into .updates per commit.

The .update() function is deferred, which means it gets executed at the end. For each row of of the .updates table, which is just one flag row, the .refresh() function gets executed, which is either a MATERIALIZED VIEW refresh or an UPDATE depending on the layer.  

Every .refresh() raises a notice describing which layer is being refreshed.

You mentioned in one comment a pile of errors for `imposm3 run`? I've had a few, but I have a long running update that I can check on. I use `docker logs` so I can peer back into the logs.

stirringhalo:
Wohoo!

PR #133, opened 2017-01-16
Faster transportation_names

stirringhalo:
This is a test of faster transportation_names, by using ST_Collect in place of ST_Union. It's still going through testing, but the `import-sql` is significantly faster, but still has fast generating of mbtiles (unlike some of the approaches tried in #129 )

I will rebase this if #131 gets merged.

Don't merge yet!!

### DIFF
--- a/layers/transportation_name/merge_highways.sql
+++ b/layers/transportation_name/merge_highways.sql
@@ -20,14 +20,14 @@ CREATE MATERIALIZED VIEW osm_transportation_name_linestring AS (
 		member_osm_ids[0] AS osm_id,
 		member_osm_ids,
 		name,
-        ref,
+        	ref,
 		highway,
 		z_order
 	FROM (
 		SELECT
-			ST_LineMerge(ST_Union(geometry)) AS geometry,
+			ST_LineMerge(ST_Collect(geometry)) AS geometry,
 			name,
-            ref,
+            		ref,
 			highway,
 			min(z_order) AS z_order,
 			array_agg(DISTINCT osm_id) AS member_osm_ids


layers/transportation_name/merge_highways.sql
```diff
@@ -13,50 +13,44 @@ DROP TABLE IF EXISTS osm_transportation_name_linestring_gen3 CASCADE;
...
 		name,
-        ref,
+		ref,
 		highway,
 		z_order
 	FROM (
-		SELECT
-			ST_LineMerge(ST_Union(geometry)) AS geometry,
+            SELECT	                
+			ST_LineMerge(ST_Collect(geometry)) AS geometry,
```
lukasmartinelli: This will just concat into a MultiLineString instead of dissolving into a single LineString.
Does Mapnik than in turn that into a single MVT linestring again?

I am not sure whether that's the same end effect with `ST_Collect`.

stirringhalo: ST_LineMerge(ST_Collect()) first aggregates to MultiLineString and then ST_LineMerge converts the MultiLineString to a LineString. ST_Dump expands any remaining MultiLineStrings into LineStrings.

The row count isn't identical, but the results should be pretty similar. I will do some visual tests to confirm.

lukasmartinelli: Great. Thanks for explaining.

lukasmartinelli:
> Don't merge yet!!

Tell me when to merge :)
Looks good!

stirringhalo:
Over Switzerland, my mbtiles (all the way to zoom 14) is completely identical between ST_Collect and ST_Union (using sqldiff on the 'tiles' table). ST_Collect also takes half the time without any tuning!

I kept member_osm_ids as it may be used in the future.

stirringhalo:
Also, it took about the same amount of time to tile Switzerland, ST_Union was slower by 6s (3min 20 vs 3min 26s) but basically identical timewise.

Ready to merge, please check it out and confirm.

stirringhalo:
Oh, and I didn't visually confirm as the sqldiff was identical. Want me to do a larger scale test?

lukasmartinelli:
Will be tested with a new rendering. It does look fine! Thansk @stirringhalo

Issue #9, opened 2016-11-04
Add ocean and sea labels


lukasmartinelli:
Positron contains labels for seas as well so sea labels should be included in `water_name`.

![basemaps_data_services_ _carto](https://cloud.githubusercontent.com/assets/1288339/20000192/93e22cd4-a276-11e6-8acc-ce568dab87f5.png)

PR #134, opened 2017-01-16
Add ocean and sea labels

lukasmartinelli:
This resolves https://github.com/openmaptiles/openmaptiles/issues/9

@stirringhalo could you take a look whether I implemented the triggers correctly?


stirringhalo:
Yep, the triggers look good but I didn't explicitly test them!

Issue #48, opened 2016-11-18
topoint function


pnorman:
housenumber_centroid.sql contains

```sql
UPDATE osm_housenumber_point SET geometry=topoint(geometry)
WHERE ST_GeometryType(geometry) <> 'ST_Point';
```

topoint is not defined anywhere.

PR #136, opened 2017-01-17
ST_PointOnSurface instead of topoint

stirringhalo:
Covers https://github.com/openmaptiles/openmaptiles/issues/48

Needs review and quick test.


lukasmartinelli:
All `topoint` does additionally is error handling. But in this case there should be no invalid polygons.

Issue #71, opened 2016-11-29
[imposm3] pseudoarea is depricated ( webmerc_area?)


ImreSamu:
See changes:
* imposm3: `add area and webmerc_area field; deprecate pseudoarea`
    *   https://github.com/omniscale/imposm3/commit/9c864952754f678be4480c31d1ee927898bc67ef

We are using now:  
* https://github.com/openmaptiles/openmaptiles/search?utf8=%E2%9C%93&q=pseudoarea

PR #137, opened 2017-01-17
Switch to webmercarea

stirringhalo:
Needs a visual inspection, and I wonder if Imposm3 takes any longer.

Handles #71


stirringhalo:
Takes 4m 3s instead of 3m 58s for Switzerland. Negligible difference.

lukasmartinelli:
> Takes 4m 3s instead of 3m 58s for Switzerland. Negligible difference.

Dude you are on a run :+1:

Issue #128, opened 2017-01-10
Issue with Eiffel tower in 3D


stirringhalo:
See https://github.com/osm2vectortiles/osm2vectortiles/issues/454#issuecomment-271494433

Quick theory: I wonder if it's an issue with units/stories/feet in the base feature of OSM that caused that as I've seen nothing elsewhere (especially over NY) to suggest there's a widespread issue. I have seen some super impossibly tall towers in Mumbai, suggesting it's an issue with how imposm3 imports/converts it.

stirringhalo:
In exploring openstreetmap and the query tool, I argue it's a data issue (see https://www.openstreetmap.org/way/308687739) as it's not tagged as a building:part... thoughts on how to easily include this without painful ST_Contains operations?

stirringhalo:
Another possibly related issue: https://openmaptiles.github.io/klokantech-3d-gl-style/#17.47/48.84638/2.24961/20.8/53

stirringhalo:
Mapping is easily handled, though there is added complexity of handling the relation in examples such as the Eiffel tower. Relations need to be better handled as for example in the case of the Eiffel Tower (https://www.openstreetmap.org/way/5013364) this building feature would overdraw the relation and constituent parts that would better describe the shape. 

There also needs to be better handling of lines as some buildings annoying use them to define the shape.

A query like : `ST_GeometryType(ST_Buffer(geometry,1, 'endcap=flat join=mitre')) from osm_building_polygon_member where ST_GeometryType(geometry) <> 'ST_Point' and ST_IsClosed(geometry) = 'f' and ST_GeometryType(geometry) <> 'ST_Polygon'` could be used to handle these open ways.

Furthermore, closed ways must be converted to polygons with ST_MakePolygon.

So this will be a pretty massive changed which I'm putting temporarily on hold until a good approach can be figured out.

ImreSamu:
@stirringhalo: 

>Furthermore, closed ways must be converted to polygons 

side note:  
* the new imposm3 has a special  `areas:`  metadata. https://imposm.org/docs/imposm3/latest/mapping.html#areas

imho:  it is worth to  test, that adding  `building:part` to the `area_tags` ,  gives a better result or not? 

```yaml
areas:
  area_tags: [ `building`, `building:part` ]
```

stirringhalo:
@ImreSamu 
Awesome! That may prove very very handy in fixing this issue. 

[...]

PR #140, opened 2017-01-19
Support buildings in relations

stirringhalo:
This handles https://github.com/openmaptiles/openmaptiles/issues/128.

Still testing, it imports the relations that defines buildings and drops the outline incase that is problematic.

Some things left to do:

- [ ] Handle multipolygons, as they are not handled for relation_member tables. This will have to be done in SQL. I will have to use a combination of array_agg and ST_MakePolygon to get there.
- [ ] May need to shrink the outline to the lowest of all the parts in the relation. Tricky case that may not be needed. Or, may need extra logic to determine if the outline overlaps other parts and only do so when needed
- [ ] May also need to not select some osm_building_polygons that are already grabbed by osm_building_polygon_member (especially where they are an outline in osm_building_polygon_member)
- [ ] Wide-spread visual tests, need to identify some good test cases and look for regressions.
- [ ] Test ETL Documents

**Don't merge yet!**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ help:
 	@echo "  "
 	@echo "Hints for designers:"
 	@echo "  ....TODO....                         # start Maputnik "
-	@echo "  ....TODO....                         # start Tileserver-gl-light"
+	@echo "  make start-tileserver                # start klokantech/tileserver-gl [ see localhost:8080 ] "
 	@echo "  make start-mapbox-studio             # start Mapbox Studio"
 	@echo "  "
 	@echo "Hints for developers:"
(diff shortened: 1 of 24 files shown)


ImreSamu:
imho : 
* ` load_all: true` :  the imposm3 cache will  be huge,   but I don't have a better idea.  
         *   and creating/testing an impoms3 patch also takes times ... 
*  probably need a performance test.  (  the imposm3 part ( import / diff )   will be slower )

maybe it is worth to add a global  `exclude` list , to minimalize cache size.
*  https://imposm.org/docs/imposm3/latest/mapping.html#tags

``` 
tags:
  load_all: true,
  exclude: [created_by, source, "tiger:*"]
```


some exclude metadatas from other systems:   
*  The ID editor discarded metadata ( https://github.com/openstreetmap/iD/blob/master/data/discarded.json ) 
*   osm2pgsql `delete` tags ( https://github.com/openstreetmap/osm2pgsql/blob/master/default.style )

stirringhalo:
Closing this as I heavily refactored my solution and need to conflict resolve.

PR #157, opened 2017-01-28
Enhanced 3D buildings

stirringhalo:
This handles https://github.com/openmaptiles/openmaptiles/issues/128.

I gutted a lot of the approach of https://github.com/openmaptiles/openmaptiles/pull/140 as it was far too slow and buggy.

This current approach works in nearly all cases, with only some minor artifacting on buildings if you zoom in close (I suspect due to conflicting heights of features that end up merged with each other due to data issues/mapbox-gl-js oddness. May be resolved with changing the sort order) (eg: http://www.openstreetmap.org/way/112452790) and detail is lost where ways tagged "building" completely envelope smaller parts both in 2D and in 3D (another data issue) (eg: http://www.openstreetmap.org/relation/1255782). That second issue would be very difficult to resolve in DB as it would require complex spatial operations and probably multi-pass fixing to resolve.

I added the 'as_numeric' function so that we can have better height imports for buildings where units are included (ie: Burj Khalifa)

Please review, select an area of your choice and see how it looks. I think the mapping can be cleaned up a little, but I added lots of extra tag support to handle poorly tagged buildings.

**Need to get some better performance numbers from my planet import, so don't merge yet.**

### DIFF
--- a/layers/building/mapping.yaml
+++ b/layers/building/mapping.yaml
@@ -1,16 +1,14 @@
-generalized_tables:
+tags:
+  load_all: true
 
+generalized_tables:
   # etldoc: imposm3 -> osm_building_polygon_gen1
   building_polygon_gen1:
     source: building_polygon
     sql_filter: area>1400.0
     tolerance: 10.0
 
-tags:
-  load_all: true
-
 tables:
-
   # etldoc: imposm3 -> osm_building_polygon
   building_polygon:
     fields:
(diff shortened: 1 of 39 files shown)


layers/building/mapping.yaml
```diff
@@ -6,6 +6,9 @@ generalized_tables:
     sql_filter: area>1400.0
     tolerance: 10.0
 
+tags:
```
ImreSamu: ### style  ( not critrical )
It is possible to add the `tags:`  to the  top of the mapping source ?  
like 
```yson
tags:
generalized_tables:
tables:
```

#### performance 
Imho: A big  `exclude`  list is important ..    
But not in the layers definitions 

Maybe we should add a global `exlude`  list with the `imposm-tools`  automatically  ?

```
tags:
  load_all: true,
  exclude: [created_by, source, "tiger:*"]
```

stirringhalo: > It is possible to add the tags: to the top of the mapping source ?

Sure thing, I will handle that

> Maybe we should add a global exlude list with the imposm-tools automatically ?

:+1:

klokan:
Have you guys tried `load_all: true` on the incremental updates?

I am very much afraid of the size of imposm3 caches for whole world for updates...

stirringhalo:
> Have you guys tried load_all: true on the incremental updates?

Not yet, wanted to find the bugs in the mapping/SQL. I will be testing the performance and hard drive footprint.

klokan:
Cool! Thank you a lot for digging into this!

IMHO we should define a process (set of tests) to do before merging to master, so master is always production ready - even for planet rendering.

stirringhalo:
> IMHO we should define a process (set of tests) to do before merging to master, so master is always production ready - even for planet rendering.

Yep, timing tests are critical and hard to identify without at least a day or two for import and export.

Issue #142, opened 2017-01-19
Tools to remove an old layer/Insert a new one


stirringhalo:
In order to avoid complete re-tiling jobs, I'd like to see tools to remove an old layer from the DB and associated MBTiles and then insert a new layer.

Do you think this will be useful?

- Identify tables/functions/views to remove, remove them. This may need to be explicitly listed in the $layer.yaml
- Insert just that layer with import-osm
- Do the associated SQL work with import-sql
- Remove the layer from the MBTiles
- Insert the layer into the MBTiles

ImreSamu:
>Do you think this will be useful?

yes,  

imho: 
* this is solves lot of future problems 
* we have more flexibility in creating layers 

question:  the final  MBTiles size will be equal ?

stirringhalo:
> question: the final MBTiles size will be equal ?

Should be, I don't see why not.

> this is solves lot of future problems
> we have more flexibility in creating layers

I agree, otherwise it gets very painful to do mapping updates as that can be a significant time impact.

Further thoughts:

- Do almost all in a Makefile. Something like `make delete $layer` and `make insert $layer`
- SQL to delete is in a separate file (delete.sql). That way it is trivial to cleanup using a psql based container
- Insert simply runs a stripped down openmaptiles.yaml with just $layer. Then that is imported through normal steps.

- Deleting/Inserting from MBTiles may be a bottleneck
- 

-

ImreSamu:
After the community feedback, probably we can create an **OpenMapTiles RoadMap for 2017**
but this proposal is very good candidate for **a long time future proof architecture.**

As I know,  the current alternative: 
* Reloading every layers - in every ~3 month ...    
   *  `import-water`
   *  `import-natural-earth`
   *  `import-lakelines`
   *  `import-osmborder`
   *  `import-osm`

>Deleting/Inserting from MBTiles may be a bottleneck

This is a new technology ( at least for me );  so it will be a critical factors.
And we have to start to create a minimal  prototype for demonstration and testing ...
As I know, no similar utilities yet https://github.com/mapbox/awesome-vector-tiles#cli-utilities

After we have a prototype ,and some experience  -  we can start re-design the architecture ...  
  *  But I would like more time to design  ... and compare the alternatives ...   ( if possible )

Issue #45, opened 2016-11-18
discuss: OpenMapTiles as a Future-proof Architecture


ImreSamu:
## vision 
My ideal use case 
*  easy importing from 
    * 3 different repos   AND  adding from a local defs  ( similar like in Golang import ) 
         * and expecting NO table/view/function  name collisions 
    * define the layer name for the mbtiles ( `layer=railway` )
    * set the update schedule ( `dataupdate=monthly` or hourly or  never?  )
*  minimal configuraions for job generating 
    *  area=albania ( from a metatable we can set the downloadURL and the BBOX defintions )  
    *  system=K8S  ( Docker-compose , K8S , Docker Swarm, ... )  for the future ...
    *  mbtiles_compress=...    for the future ... like   https://github.com/rastapasta/tileshrink ? 
*  RUN 

like this:
```
openmaptiles clear_all
openmaptiles import_defs github.com/cleartables/omt_layers/cleartables/clearrailway_slim layer=railway   dataupdate=monthly
openmaptiles import_defs github.com/openmaptiles/layers/openmaptiles/building3d          layer=building  dataupdate=weekly 
openmaptiles import_defs github.com/openmaptiles/layers/openmaptiles/landuse             layer=landuse   dataupdate=NO 
openmaptiles import_defs github.com/klokantech/openvectordefs/klokantech/extendedpoi     layer=poi       dataupdate=daily 
openmaptiles import_defs /myprivate_dir/layers/extremehiking                             layer=hiking    dataupdate=daily
openmaptiles generate_jobs area=albania min_zoom=1 max_zoom=10 mbtiles_compress=4 parallelization=8 system=K8S update=yes
openmaptiles schedule_and_run_jobs 
```

And if possible we can use `imposm3` and/or `osm2pgsql` tools in each layer 

## problem

IMHO :  Our FUTURE  Problems ( based on my experience in working on big DW-s ) 
*  How we can avoid SQL table name collisions, when more organisations developing different layers in parallel ,
  * in extreme case they choose same table/view/function names 
     * and creating a very hard to debug problem types .. 
*  Company/organisations want to create a private layers and want to use with a community layers together  
  *  without a problems 
*  ... 

IMHO :  we need to create and adapt some guideline, conventions for this ...

## solutions ?  

As I see we have 4 basic solutions for start of thinking ... 

* V1(PostgreSQL_Schema as a `layer_group_id`) 
   *  create SQLSchema for every organisations, and they can write to only their shema  
[...]

PR #158, opened 2017-01-28
Add and delete layers

stirringhalo:
This contains the `make add` and `make delete` work for https://github.com/openmaptiles/openmaptiles/issues/142.

The MBTiles work will come later.

This also covers part of https://github.com/openmaptiles/openmaptiles/issues/45 and the concerns about name collisions as everything at the SQL level (not imposm3 tables) gets put into schemas.

- [x] add-layer
- [x] delete-layer
- [ ] test!


stirringhalo:
https://github.com/openmaptiles/import-osm/pull/7 is needed for the add-layer functionality to be possible.

stirringhalo:
So is: https://github.com/openmaptiles/openmaptiles-tools/pull/14

stirringhalo:
Add-layer support is complete, now thorough testing is needed. In summary:

1. https://github.com/openmaptiles/openmaptiles-tools/pull/14 adds a script to generate a layer-specific .yaml from an existing openmaptiles.yaml
2. https://github.com/openmaptiles/import-osm/pull/7 rolls all the osc.gz into the PBF for `update-osm`. Same support could/should be added to the single diff updates.
3. Finally, this PR adds the SQL support for .delete() and the Makefile support for add-layer

How this is intended to work:

1. Regular import happens
2. Regular `update-osm` happens
3. Oh no, major mapping changes!
4. Add a .lock to the update-osm container, wait for the latest update to finish and it to pause (no changes will be written to the PBF).
5. Get the changes into the layers directory
6. Run `make add-layer L=poi D=$(pwd)`. This will update the openmaptiles.yaml, clean the build dir, and then add the new layer-specific build.
7. Run `import-osm` again.
8. Remove the .lock file.

I need to test steps 7 and 8 further, but it's looking pretty good to that point.

stirringhalo:
This only works for single layers, it would be nice to have it work for multiple layers too so that major changes can be done all together too.

Issue #172, opened 2017-02-09
'Place' layer text is 'name_en' but text is not English


nouryf:
Hi,

https://editor.openmaptiles.org/?style=https://rawgit.com/openmaptiles/osm-bright-gl-style/master/style.json#9.13/31.5113/-8.2616

Looking at the labels displayed for 'place-town', the text field used is {name_en} but the displayed text is both 'English/French and Arabic'. I thought that its {name} that has a combination of languages in it... and {name_en} would have only English name (latin).

Is this means that osm dataset has in {name_en} a wrong attribute encode or may be its the produced .mbtiles that is storing in {name_en} the name coming from {name} ?

I was trying to work around a limitation for supporting Arabic in text labels to use {name_en} thinking it would have just the English name.

Thanks

ImreSamu:
name_en =    [name:en](http://taginfo.openstreetmap.org/keys/name:en) the english name  (if available, otherwise same as [name](http://taginfo.openstreetmap.org/keys/name))



the `name`  is the local name - and sometimes mixed languages ..
* https://wiki.openstreetmap.org/wiki/Multilingual_names
* Morocco:   https://wiki.openstreetmap.org/wiki/Multilingual_names#Morocco

jirik:
In case of some objects `name_en` is set to `name` if `name:en` is not defined (this is the case of this node: http://www.openstreetmap.org/node/3215639008#map=14/31.7125/-8.3162). However in case of other objects `name_en` is not set if `name:en` is not defined.

This disunity appears also outside of `place` layer. I am not sure if it was done by purpose. Probably not.

jirik:
https://github.com/openmaptiles/openmaptiles/blob/e406537025ff9229d99957cf575d36d8d2666fe5/layers/place/city.sql#L16

jirik:
@nouryf 
> Is this means that osm dataset has in {name_en} a wrong attribute encode or may be its the produced .mbtiles that is storing in {name_en} the name coming from {name} ?

Both can be correct. If the `name:en` is not defined, then `name` is used.

I have unified rules for `name_en`: https://github.com/openmaptiles/openmaptiles/pull/190

PR #190, opened 2017-03-17
Add German names (name_de), unify English names (name_en)

jirik:
German names `name_de` added to layers. It contains value of the first non-empty tag:
1. `name:de`
1. `name`
1. `name:en`

Similarly, `name_en` now contains value of the first non-empty tag:
1. `name:en`
1. `name`

All layers with names were affected:
- mountain_peak
- place
- poi
- transportation_name
- water_name
- waterway

### DIFF
--- a/layers/mountain_peak/layer.sql
+++ b/layers/mountain_peak/layer.sql
@@ -3,11 +3,13 @@
 -- etldoc:     style="rounded,filled", label="layer_mountain_peak | <z7_> z7+" ] ;
 
 CREATE OR REPLACE FUNCTION layer_mountain_peak(bbox geometry, zoom_level integer, pixel_width numeric)
-RETURNS TABLE(osm_id bigint, geometry geometry, name text, name_en text, ele int, ele_ft int, "rank" int) AS $$
+RETURNS TABLE(osm_id bigint, geometry geometry, name text, name_en text, name_de text, ele int, ele_ft int, "rank" int) AS $$
    -- etldoc: osm_peak_point -> layer_mountain_peak:z7_
-   SELECT osm_id, geometry, name, name_en, ele::int, ele_ft::int, rank::int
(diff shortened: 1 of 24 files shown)


jirik:
Related to #172

nicooprat:
Hi,

Do you plan to add other languages? French would be nice :) I was using `https://osm2vectortiles.tileserver.com/v2.json` that supported it but looks like v3 doesn't...

Thanks for your work,
Regards.

jirik:
Franch is already there, together with about 50 languages!
See https://github.com/openmaptiles/openmaptiles/releases/tag/v3.6

nicooprat:
Ok weird my 2nd edit didn't show up. I actually found out that `{name:fr}` was available, but it gave me some disappointing results. Here are my configuration files (only `text-field` changes): https://gist.github.com/nicooprat/5679c48a89400a507c4b157c7142d371

And here are some results (in Corsica, a french Island at South of it): 
* lat: 42.13990692518783
* lng: 9.077687793814448
* zoom: 10

osm

Pretty all city names are not showing up. Any idea?

Thanks.

jirik:
@nicooprat Because they do not have `name:fr` tag in OSM data:
https://www.openstreetmap.org/node/51305031
New `name:*` attributes are meant to be used with fallback layer. Meaning you need to create two layers:
- one for features with `name:fr` attribute (showing `name:fr` as a label)
- second for features without `name:fr` attribute (showing some other name as a label, probably `name_int`)

nicooprat:
Ok, makes sense. I just found this: https://github.com/openmaptiles/klokantech-basic-gl-style/blob/master/lang-fallback.json but I don't understand how to implement it, any hint? Sorry if it looks obvious for you...

jirik:
@nicooprat Sorry for delay. Nevertheles you have found it :) Another usage is here:
https://github.com/klokantech/gl-style-package-spec
I think it is not yet documented feature, but it is not difficult.

PR #193, opened 2017-03-21
Updated versions of docker images in docker-compose

MartinMikita:
- import-osm:0.4
 - import-water:0.3
 - generate-vectortiles:0.1.1
Use same version of import-osm in update-osm and import-osm-diff containers.


jirik:
LGTM

MartinMikita:
Changed version of `import-water` to `0.3.1` docker image, because of a little typo mistake in the source repository.

PR #207, opened 2017-04-07
 osm_ocean_polygon_gen4

stirringhalo:
Don't merge yet, needs testing and make sure the ETL Docs are good.

May propose the use of ZRes() as an alternative instead of hard-coding simplification values.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     networks:
     - postgres_conn
   import-water:
-    image: "openmaptiles/import-water:0.3.1"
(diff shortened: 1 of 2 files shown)


jirik:
Z8.5 before
screenshot_20170413_152336

and after
screenshot_20170413_152811

I think it's sufficient for Z8. There will be probably similar issues for Z7. ETL docs looks good.

@stirringhalo I think you can set import-water to [v0.4](https://github.com/openmaptiles/import-water/releases) in docker-compose.yml and merge.

jirik:
Thanks!

PR #213, opened 2017-04-08
Add primary to zoom 7, and fix index for gen3 & 6

stirringhalo:
Index for gen3 was missing 'primary'. Do we strictly need the partial idx? Doesn't really do anything since no selections are made on the highway type...
Needs testing


MartinMikita:
Yes, partial index decreased time of import-sql step, from 4 hours down to 1.5 hours :)

See [create materialized view SQL](https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/merge_highways.sql#L40-L41)

stirringhalo:
> Yes, partial index decreased time of import-sql step, from 4 hours down to 1.5 hours for the whole planet > :)

:+1:

stirringhalo:
Did a visual test, looks good at z7. Still disjoint at z6 but I'm happy to revisit.

Issue #210, opened 2017-04-07
Insufficient level of detail on  zoom level 7


juho-hanhimaki:
We're having issues with too low amount of detail in some zoom levels. For example the OSM2VT used to provide a lot more detail in similar zoom levels. Also comparing to other map providers such as Google, openmaptiles is missing huge amounts of detail on similar zoom levels.

Please see this screenshot of southern Finland (main part of Finland). One can't really see much useful data on this map compared to OSM2VT which had nice amount of detail.

![openmaptiles](https://cloud.githubusercontent.com/assets/12390182/24806365/47b8c708-1bbd-11e7-8bd8-f9200c0614a0.png)

As far as i know this isn't issue with styles, it's just how the data is. For us this is problematic as we need to show asset locations on map and there needs to be enough visible context/detail for the user.

pathmapper:
Apart from the question what details for which zoom level should be displayed by default, you have the possibility to modify the layer SQL code to fit your needs.

Example:
To have more detailed lakes on lower zoom levels you could adapt [water.sql](https://github.com/openmaptiles/openmaptiles/blob/master/layers/water/water.sql).
Take a look at the [schema](https://openmaptiles.org/schema/#water) and the diagrams to find out which sources are used for which zoom level by default.

klokan:
Zoom level 7 should be improved indeed - roads are tracked under #200 already, more of the place names should be elevated from zoom 8 here too. Pull requests from the community are very welcome - @stirringhalo - are you going to have time for this one?

We have to be very careful about the size of the final tiles and number of edges - an aggressive merging and generalization of the elevated features is required.

stirringhalo:
@klokan Hope so, definitely before v3.6.

stirringhalo:
This is indeed fixed with #214

stirringhalo:
Closed with #214

juho-hanhimaki:
I tried generating tiles with the new version and unfortunately it seems the problem still exists for us. The roads are certainly better now but the insufficient detail of place names are quite terrible for us. Even though Finland is a country of little population I don't think the map should look like empty canvas.

Please compare this image of 3.6 version to the image in original message:
![openmaptiles36](https://user-images.githubusercontent.com/12390182/27121014-50a5328c-50ee-11e7-898f-f3d7ede132c1.png)

stirringhalo:
[...]

Issue #205, opened 2017-04-05
Water layer is over-simplified at zoom level 8


tinoue:
https://openmaptiles.github.io/osm-bright-gl-style/#8.01/43.2660/145.5424

See coastline and road around "Nemuro Peninsula.".
Water layer is relatively over simplified at this zoom level and the Road #44 seems to run over ocean.
Also, It seems to be OK to simplify "landcover class=wood" more.

tinoue:
Or is this geometry mismatch between Natural Earth and OSM ?
North-West coastline seems to unmatch with roads.
https://openmaptiles.github.io/osm-bright-gl-style/#7.01/43.279/141.155

klokan:
Thanks for the report. It is indeed the switch between Natural Earth and OSM which makes the trouble on this zoom level.

It would be good to simplify the OSM water polygons (from "import-water") via SQL in Postgres and use them one level up instead of NE.

See:
https://github.com/openmaptiles/openmaptiles/tree/master/layers/water

Could you please try to do this and make a pull request? It may be a great contribution to the OpenMapTiles open-source project and a lovely first code from a new project team member ;-)

Once you have the first version we can review your code and merge it, if it fine for production use in the next official release! It would be cool @tinoue

stirringhalo:
Here's some proposed changes:
https://github.com/openmaptiles/import-water/pull/1
https://github.com/openmaptiles/openmaptiles/pull/207

Needs testing, but it's a start.

jirik:
@stirringhalo Thanks, I will look at it next week.

stirringhalo:
Closed in https://github.com/openmaptiles/openmaptiles/pull/207

Issue #10, opened 2016-11-04
Feedback on Vector Tile Schema


lukasmartinelli:
The new vector tile schema v3.1 is based on making Positron and Dark Matter and it would be good to have someone read through the layer documentation and give feedback.

PR #214, opened 2017-04-08
Switch to ZRes

stirringhalo:
I'm testing switching to the ZRes function in order to make it easier to modify the tolerance and area filters and make it more clear the scales being used. Testing this now, and if promising (even if values need scaling for tile size issues), I will move the other layers to a similar approach.

This will help with issues similar to https://github.com/openmaptiles/openmaptiles/issues/210 and https://github.com/openmaptiles/openmaptiles/issues/205

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,7 +75,7 @@ services:
     volumes:
      - ./build:/sql
   openmaptiles-tools:
-    image: "openmaptiles/openmaptiles-tools:PIXEL_GEOUNIT"
+    image: "openmaptiles/openmaptiles-tools:latest"
     env_file: .env
     networks:
     - postgres_conn
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ volumes:
   cache:
 services:
   postgres:
-    image: "openmaptiles/postgis"
+    image: "openmaptiles/postgis:2.3"
     volumes:
     - pgdata:/var/lib/postgresql/data
     networks:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ volumes:
   cache:
 services:
   postgres:
-    image: "openmaptiles/postgis:2.3"
+    image: "openmaptiles/postgis"
     volumes:
     - pgdata:/var/lib/postgresql/data
     networks:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,7 +75,7 @@ services:
     volumes:
      - ./build:/sql
   openmaptiles-tools:
-    image: "openmaptiles/openmaptiles-tools:latest"
+    image: "openmaptiles/openmaptiles-tools:PIXEL_GEOUNIT"
     env_file: .env
     networks:
     - postgres_conn
--- a/layers/aeroway/layer.sql
+++ b/layers/aeroway/layer.sql
@@ -9,7 +9,7 @@ RETURNS TABLE(geometry geometry, class text) AS $$
         -- etldoc:  osm_aeroway_linestring -> layer_aeroway:z13
         -- etldoc:  osm_aeroway_linestring -> layer_aeroway:z14_
         SELECT geometry, aeroway
-        FROM osm_aeroway_linestring WHERE zoom_level >= 11
+        FROM osm_aeroway_linestring_gen3 WHERE zoom_level >= 11
         UNION ALL
         -- etldoc:  osm_aeroway_polygon_gen2 -> layer_aeroway:z12
         SELECT geometry, aeroway
--- a/layers/aeroway/layer.sql
+++ b/layers/aeroway/layer.sql
@@ -9,7 +9,7 @@ RETURNS TABLE(geometry geometry, class text) AS $$
         -- etldoc:  osm_aeroway_linestring -> layer_aeroway:z13
         -- etldoc:  osm_aeroway_linestring -> layer_aeroway:z14_
         SELECT geometry, aeroway
-        FROM osm_aeroway_linestring_gen3 WHERE zoom_level >= 11
+        FROM osm_aeroway_linestring WHERE zoom_level >= 11
         UNION ALL
         -- etldoc:  osm_aeroway_polygon_gen2 -> layer_aeroway:z12
         SELECT geometry, aeroway
--- a/layers/aeroway/mapping.yaml
+++ b/layers/aeroway/mapping.yaml
@@ -2,20 +2,20 @@ generalized_tables:
   # etldoc: imposm3 -> osm_aeroway_polygon_gen3
   aeroway_polygon_gen3:
     source: aeroway_polygon_gen2
-    sql_filter: area>power(GEOUNIT11,2)
-    tolerance: GEOUNIT11
+    sql_filter: area>power(ZRES11,2)
+    tolerance: ZRES11
 
   # etldoc: imposm3 -> osm_aeroway_polygon_gen2
   aeroway_polygon_gen2:
     source: aeroway_polygon_gen1
-    sql_filter: area>power(GEOUNIT12,2)
-    tolerance: GEOUNIT12
+    sql_filter: area>power(ZRES12,2)
+    tolerance: ZRES12
 
   # etldoc: imposm3 -> osm_aeroway_polygon_gen1
   aeroway_polygon_gen1:
     source: aeroway_polygon
-    sql_filter: area>power(GEOUNIT13,2)
-    tolerance: GEOUNIT13
+    sql_filter: area>power(ZRES13,2)
+    tolerance: ZRES13
 tables:
   # etldoc: imposm3 -> osm_aeroway_polygon
   aeroway_polygon:
--- a/layers/aeroway/mapping.yaml
+++ b/layers/aeroway/mapping.yaml
@@ -1,15 +1,21 @@
 generalized_tables:
+  # etldoc: imposm3 -> osm_aeroway_polygon_gen3
+  aeroway_polygon_gen3:
+    source: aeroway_polygon_gen2
+    sql_filter: sqrt(area)<ZRes(11)
+    tolerance: ZRes(11)
+
   # etldoc: imposm3 -> osm_aeroway_polygon_gen2
   aeroway_polygon_gen2:
     source: aeroway_polygon_gen1
-    sql_filter: area>240000
-    tolerance: 50.0
+    sql_filter: sqrt(area)<ZRes(12)
+    tolerance: ZRes(12)
 
   # etldoc: imposm3 -> osm_aeroway_polygon_gen1
   aeroway_polygon_gen1:
     source: aeroway_polygon
-    sql_filter: area>60000
-    tolerance: 20.0
+    sql_filter: sqrt(area)<ZRes(13)
+    tolerance: ZRes(13)
 tables:
   # etldoc: imposm3 -> osm_aeroway_polygon
   aeroway_polygon:
--- a/layers/aeroway/mapping.yaml
+++ b/layers/aeroway/mapping.yaml
@@ -2,20 +2,20 @@ generalized_tables:
   # etldoc: imposm3 -> osm_aeroway_polygon_gen3
   aeroway_polygon_gen3:
     source: aeroway_polygon_gen2
-    sql_filter: sqrt(area)<ZRes(11)
-    tolerance: Z11
+    sql_filter: area>power(GEOUNIT11,2)
+    tolerance: GEOUNIT11
 
   # etldoc: imposm3 -> osm_aeroway_polygon_gen2
   aeroway_polygon_gen2:
     source: aeroway_polygon_gen1
-    sql_filter: sqrt(area)<ZRes(12)
-    tolerance: Z12
+    sql_filter: area>power(GEOUNIT12,2)
+    tolerance: GEOUNIT12
 
   # etldoc: imposm3 -> osm_aeroway_polygon_gen1
   aeroway_polygon_gen1:
     source: aeroway_polygon
-    sql_filter: sqrt(area)<ZRes(13)
-    tolerance: Z13
+    sql_filter: area>power(GEOUNIT13,2)
+    tolerance: GEOUNIT13
 tables:
   # etldoc: imposm3 -> osm_aeroway_polygon
   aeroway_polygon:
--- a/layers/aeroway/mapping.yaml
+++ b/layers/aeroway/mapping.yaml
@@ -3,19 +3,19 @@ generalized_tables:
   aeroway_polygon_gen3:
     source: aeroway_polygon_gen2
     sql_filter: sqrt(area)<ZRes(11)
-    tolerance: ZRes(11)
+    tolerance: Z11
 
   # etldoc: imposm3 -> osm_aeroway_polygon_gen2
   aeroway_polygon_gen2:
     source: aeroway_polygon_gen1
     sql_filter: sqrt(area)<ZRes(12)
-    tolerance: ZRes(12)
+    tolerance: Z12
 
   # etldoc: imposm3 -> osm_aeroway_polygon_gen1
   aeroway_polygon_gen1:
     source: aeroway_polygon
     sql_filter: sqrt(area)<ZRes(13)
-    tolerance: ZRes(13)
+    tolerance: Z13
 tables:
   # etldoc: imposm3 -> osm_aeroway_polygon
   aeroway_polygon:
--- a/layers/boundary/mapping.yaml
+++ b/layers/boundary/mapping.yaml
@@ -3,27 +3,27 @@ generalized_tables:
   # etldoc: imposm3 -> osm_boundary_linestring_gen5 
   boundary_linestring_gen5:
     source: boundary_linestring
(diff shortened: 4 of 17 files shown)


ImreSamu:
The idea is good!    

Maybe my dummy questions , but it is good time for rethink some default rules ..

####   256x256 vs. 4096x4096
The   ZRES   is  256x256  ( https://github.com/mapbox/postgis-vt-util/blob/master/src/ZRes.sql )
   *   `Takes a web mercator zoom level and returns the pixel resolution for that scale, assuming 256x256 pixel tiles. Non-integer zoom levels are accepted.`

But as I know  Vector tiles is storing data  `4096x4096`  (default extent )

I know that the ZRES for `4096x4096`  is not for everybody and increasing size, but it would be ideal if we can more easily hack ( and play with )  this settings  with a global parameters ..  
( maybe later with a next refactoring )
for example:
* 256x256
* 512x512
* 1024x1024
* 2048x2048
* 4096x4096
* ...

#### Area filter and donut shaped polygons
Probably we have some polygons with a hole in the middle (a.k.a. donut shaped)  
What is the effect of the `sqrt(area)<ZRes(13)` filter rules for donut shaped polygons  in the map ?

stirringhalo:
I naively thought I could simply drop ZRes into the tolerance and it would all work. In looking at the tolerance field in Imposm3, it's a float64 because it's directly put into GEOS (see https://github.com/omniscale/imposm3/blob/c3f041aa85a54e7107885778dfa766b8ff88298c/geom/geos/ops.go).

So, two alternatives:
1. We add a new field type in Imposm3 and the logic to handle ZRes equivalent. Slowest to get merged.
2. In OpenMapTiles-tools we parse a "tolerance_z: z($float)" function into an actual value and set the tolerance on that.

ImreSamu:
>So, two alternatives:

What about - replacing/evaluating  -  in our  Python pre-processor?  
* https://github.com/openmaptiles/openmaptiles-tools#generate-imposm3-mapping-file

stirringhalo:
> What about - replacing/evaluating - in our Python pre-processor?
> https://github.com/openmaptiles/openmaptiles-tools#generate-imposm3-mapping-file

That's what I'm doing now!

stirringhalo:
@ImreSamu  See https://github.com/openmaptiles/openmaptiles-tools/pull/15 . The final syntax I chose was "Z##".

stirringhalo:
ZRes support added in https://github.com/openmaptiles/openmaptiles-tools/commit/4957a6657dd3fd520bb12ba6d291dabe83efb2d8

ImreSamu:
ideas for brainstorming:

###  name conventions
instead of  _gen1 ; _gen2 ; _gen3 , maybe we can use `_zresXX`  in the table name.
some example:
*  `aeroway_polygon_gen3`  ->  `aeroway_polygon_zres11`
*  `aeroway_polygon_gen2`  ->  `aeroway_polygon_zres12`
*  `aeroway_polygon_gen1`  ->  `aeroway_polygon_zres13`

*  `boundary_linestring_gen5` -> `boundary_linestring_zres09`
*  `boundary_linestring_gen4` -> `boundary_linestring_zres10`
*  `boundary_linestring_gen3` -> `boundary_linestring_zres11`
*  `boundary_linestring_gen2` -> `boundary_linestring_zres12`
*  `boundary_linestring_gen1` -> `boundary_linestring_zres13`

For me -  it is much easier to detect mapping errors in the ETLGRAPH   and less complexity.

stirringhalo:
> instead of _gen1 ; _gen2 ; _gen3 , maybe we can use _zresXX in the table name. some example: For me - it is much easier to detect mapping errors in the ETLGRAPH and less complexity.

I'm considering the same change. Makes it a ton easier to understand what features go where at what detail level. There's some overlap with multiple genXX used for multiple zooms, so we would have to make the decision whether:

1. To include the detail at the highest level that the generalized table will be used at
2. To include the medium detail that the generalized table will be used at
3. Strictly have one generalized table per zoom. Would lead to more disk use, but cleaner ETLs.

I'm leaning towards 2 for the immediate future, perhaps 3 on the next major refactoring.

stirringhalo:
I've done all layers with ZXX and area filtering based on ZRes. Only place I didn't change it was water_name/layer.sql and place/island_rank.sql as that's already nicely set is isn't likely to need tweaking.

Kept the _genX names the same as some generalized tables are used for more than one zoom

stirringhalo:
> [Apr 12 02:17:38] SQL Error: pq: function zres(integer) does not exist in query CREATE TABLE "import"."osm_landuse_polygon_gen1" AS (SELECT "osm_id",

Looks like Imposm3 lower-cases the SQL function string?

jirik:
@stirringhalo This PR seems really promising! I think we will need to do careful visual review and also to compare size of generated tiles before we merge it. Could you please let me know when this PR is ready for these tests? I am looking forward to do the review.

jirik:
> instead of _gen1 ; _gen2 ; _gen3 , maybe we can use _zresXX in the table name.
some example:

I was considering the same change too. However I suggest to do it in another PR after this PR is merged. I have created a special issue for it: https://github.com/openmaptiles/openmaptiles/issues/218

ImreSamu:
@stirringhalo ; @jirik 

for more brainstorming ... 

## extended ideas:

As I see [Mapnik](https://github.com/mapnik/mapnik/wiki/PostGIS)  using the `!pixel_width!*!pixel_height!` tokens for optimalising for printing.
```
!pixel_width! (Mapnik >= 2.1.0) -- width of a pixel in geographical units
!pixel_height! (Mapnik >= 2.1.0) -- height of a pixel in geographical units
```
and  [I see](https://github.com/mapnik/mapnik/issues/2361) a `WHERE area/NULLIF(!pixel_width!*!pixel_height!,0) > 0.1` for filtering 


What about ? ( my proposal for brainstorming ) 
```yaml 
#
#  template for imposm3 mapping file
#  Warning:  this  is only a template. The  !PIXEL_GEOUNIT__!  parameters calculated and replaced by openmaptiles-tools!     
#
generalized_tables:
   aeroway_polygon_z12:
    source: aeroway_polygon_gen1
    sql_filter: area > power( !PIXEL_GEOUNIT12!, 2) 
    tolerance: !PIXEL_GEOUNIT12!
```

*  the  `sql_filter:` and the  `tolerance:` field will be parsed  for ` !PIXEL_GEOUNIT__!` and replaced with the the correct value.
*  we can modify the pixel_scale base unit easily ( 256x256 ; ....  ; 4096x4096 ) in the [openmaptiles.yaml](https://github.com/openmaptiles/openmaptiles/blob/master/openmaptiles.yaml) metadata. 
in my example = 1024   ( 1024x1024 ) 

```
  pixel_scale: 1024    
  maxzoom: 14
  minzoom: 0
```

summary:
my strong wish ( and I can help in the implementation!  )
* imho  the  256x256 is too low for retina quality , for Unity3D /Blender/*  or  for fine vector zooming ( `#10.04`; `#10.37`; `#10.70` )    I would prefer if we can modify this parameter in the `openmaptiles.yaml` metadata.  ( `pixel_scale:` ? )
* it would be better if we can use only 1 python function to conversion  ( now half is python other half is sql ( `ZRES()` )  it is too complex for me and for the new contributors - for modify or debug   ) 
* (documentation) We need to add some comments in the imposm3 mapping files (somewhere), that this is not a standard imposm3 mappings ( only works with the openmaptiles-tools )   it was a lesson about my imposm3 customisations , otherwise people will be ask in the other chanel -  why this is not working with imposm3 ?  ( [example](https://groups.google.com/d/msg/imposm/060bzlX7Ihw/EvjuvrnoAwAJ) )

just an ideas for thinking :
* maybe the power function is faster than the sqrt() , but I am not tested the postgresql implementations.
* the !PIXEL_GEOUNIT12! is similar  name convention than mapnik tokens. no problem if renamed,

stirringhalo:
@ImreSamu Actually, I'm all for using Mapnik style tokens. I was all ready to use the ZRes function in both places until I realized it only works in "sql_filter", not "tolerance". Now, it would be much harder to adjust for pixel resolution.

My plans:

- [ ] Reimplement with PIXEL_GEOUNIT
- [ ] Switch all layers to the new schema

To confirm !PIXEL_GEOUNIT! = (40075016.6855785/(pixel_scale*2^z))

where pixel_scale = 256, 512, 1024, etc...

stirringhalo:
See https://github.com/openmaptiles/openmaptiles-tools/pull/19 , uses the syntax GEOUNIT## for replacing. I will be much more cautious integrating this parsing as opposed to ZRes due to previous issues.

stirringhalo:
Visual tests look great, size comparisons to downloads from OpenMapTiles look great. I think this is ready to go, but I will wait for the signoff on https://github.com/openmaptiles/openmaptiles-tools/pull/19 and here. The previous tile schemas (openmaptiles.yaml) **won't** be compatible as it needs to have the "pixel_scale" result.

jirik:
@stirringhalo Wow, you are running fast! Interesting idea with GEOUNIT, I like it. I have just never heart term **geounit**. Is it used somewhere else? It reminds me term **resolution** from OpenLayers meaning **projection units per pixel**: http://openlayers.org/en/latest/apidoc/ol.View.html 
As far as I understand concept of geounit, it has the same meaning as OpenLayers' resolution. If so, what about using term **resolution** instead of **geounit**? Because it should be more familiar to users.

stirringhalo:
@jirik It was @ImreSamu's request for "PIXEL_GEOUNIT" and in playing with parsing, I ended up shortening it to "GEOUNIT##". Another possible term could be "MAXIMUMRES" given the inputs are the pixel_scale (measure of density) and zoom, which decides how high a resolution you can go until the display + zoom level combination makes no difference.

ImreSamu:
@stirringhalo , @jirik 

:)

*  [naming is not easy](http://hilton.org.uk/blog/why-naming-things-is-hard)
*  As I know 
   *  `GEOUNIT12`  ~  [ZRES(12)](https://github.com/mapbox/postgis-vt-util/blob/master/src/ZRes.sql) 
         * but in the future we need supporting
             * different projections ( ZRES is based on mercator )  
             * different scale ( ZRES fixed for 256x256 )    ( this is resolved via `pixel_scale` ) 
   * `ZRES` ~  **Z**oom level  + **RES**olution 

* Visual test  - We have at least 4 alternative syntax : 

```
aeroway_polygon_gen1:
  source: aeroway_polygon
  sql_filter: area>power(GEOUNIT13,2)
  tolerance: GEOUNIT13
```

```
aeroway_polygon_gen1:
  source: aeroway_polygon
  sql_filter: area>power(MAXIMUMRES13,2)
  tolerance: MAXIMUMRES13
```

```
aeroway_polygon_gen1:
  source: aeroway_polygon
  sql_filter: area>power(ZRES13,2)
  tolerance: ZRES13
```

```
aeroway_polygon_gen1:
  source: aeroway_polygon
  sql_filter: area>power(RESOLUTION13,2)
  tolerance: RESOLUTION13
```


* Or we can create new alternatives ?   `ZRESOLUTION13`   ( this is a  mix of `ZRES` + `RESOLUTION`  )
 
```
aeroway_polygon_gen1:
  source: aeroway_polygon
  sql_filter: area>power(ZRESOLUTION13,2)
  tolerance: ZRESOLUTION13
```

Re-checking my preference ,   anything with starting "Z" letter  (=**Z**oomLevel )  is acceptable for me.
*  ZRESOLUTION13  , ZRES13 , ZMAXIMUMRES13,  ZGEOUNIT13

stirringhalo:
Hmm, in light of possible future alternative projection support, I actually like to keep "GEOUNIT" as it is projection agnostic (ZRes is tied to Mercator with postgis-vt-util). It gets tricky as depending on the projection, the tolerance will be in different units...

jirik:
@ImreSamu Yes, naming matters :) Thanks for the link ot the post and for visual test cases.

Actually I like ZRES## the most, because
- it is already used at least in postgis-vt-util
- it combines zoom level (one of inputs) and resolution (output)
- it is short

ZRESOLUTION or RESOLUTION seem also reasonable to me.

@stirringhalo Yes, [current definition](https://github.com/mapbox/postgis-vt-util#zres) mentions web mercator and 256x256 pixel tiles:

> Takes a web mercator zoom level and returns the pixel resolution for that scale, assuming 256x256 pixel tiles. Non-integer zoom levels are accepted.

I believe the key part of the definition is this: **Takes a zoom level and returns the pixel resolution for that scale**. By discussed changes we are only making the definition more general:
- enabling other pixel scale, i.e. not only 256x256 pixel tiles (already [implemented by you](https://github.com/openmaptiles/openmaptiles-tools/pull/19/files#diff-725b026a6e21f6d524ffff787be6b818R10))
- enabling other SRS (not yet implemented)

So our documentation can say for example
> ZRES## is a constant having value of pixel resolution for given zoom level (##). Values are also affected by pixel scale and SRS that can be changed in `openmaptiles.yaml`. Default pixel scale is 256x256, default SRS is Web Mercator.

jirik:
@stirringhalo Thanks for this improvement! I am looking forward to trying it today. Also thanks for renaming GEOUNIT to ZRES.

One note: I see `openmaptiles/openmaptiles-tools:PIXEL_GEOUNIT` Docker tag in [quickstart.sh](https://github.com/openmaptiles/openmaptiles/pull/214/files#diff-8846dd18c9ee9c09dadeee541156c2b8R169).
Shouldn't we use [latest](https://hub.docker.com/r/openmaptiles/openmaptiles-tools/tags/) instead since [the change is already in master](https://github.com/openmaptiles/openmaptiles-tools/commit/7983dccc9cbf02922c9035771efc6d2824f7186c)?

stirringhalo:
Oops, I forgot to set it back in quickstart.sh (but it was in docker-compose.yml, but that's not used in quickstart.sh). Good catch! I will resolve that now!

jirik:
It seems this PR changed some durations and sizes. I think it was expected. Anyway, here are some numbers for Belgium:

| | 004a6f9eea9b6dd5bc2fc7d0cb47a3eeb4a31f3f | 90690d2a39a50e2e24ae8ec94ddb522166bb7499 | increase |
| --- | --- | --- | --- |
| `import-osm` duration | 3m20s | 4m53s | 46% |
| size of `osm_*_gen*` tables after `import-osm` | 2590 MB | 3332 MB | 28% |

Size was generated through this command:
```sql
SELECT
  pg_size_pretty(sum(size)) AS total_size,
  pg_size_pretty(sum(ext_size)) AS ext_size
FROM (
  SELECT
    relname as "Table",
    pg_total_relation_size(relid) as "size",
    pg_total_relation_size(relid) - pg_relation_size(relid) as "ext_size"
  FROM pg_catalog.pg_statio_user_tables
  WHERE schemaname='public'
    AND relname LIKE 'osm\_%' AND (relname LIKE '%\_gen_' OR relname LIKE '%\_gen__')
) a;
```

stirringhalo:
@jirik A slight size increase isn't entirely a surprise, thanks for the numbers!

jirik:
One surprise, at least for me: size of final .mbtiles decreased about 30% (see extended table above). I wonder where is the difference. I will probably look at it later

jirik:
A have added two more rows into table above. The last one says that size of .mbtiles for Belgium (Z0-12) increased about 84%. It is quite enough. Before rendering the planet for release 3.6, we will probably need to adjust some tolerance/area limits to be more close to original values.

stirringhalo:
Sounds good, though I had compared tile sizes before and after for a region (sorry don't remember the region) and the change was negligible. I wonder which layer was the biggest culprit for the size increase..

klokan:
The increase of the tile size by 84% is unfortunately unacceptable.
We have to optimise this further before release v3.6 - this is urgent. The release is scheduled in two weeks.

jirik:
I made [utility](https://github.com/klokantech/openmaptiles/blob/adjust_zres/qa/osm_gen_sizes.py) for table sizes and checked differences. Biggest differences was in
* osm_building_polygon_gen*
* osm_landcover_polygon_gen*
* osm_landuse_polygon_gen*

Together it was about 99% of total difference. So a have adjusted zres of these layers closer to original values: https://github.com/openmaptiles/openmaptiles/compare/master...klokantech:adjust_zres

With these changes size of Belgium Z0-12 is 38.4 MB, so increase about 11% comparing to the original values. Even with these values 97% of the change make landcover, landuse, and buildings tables.

I am going to make PR and merge [adjust_zres](https://github.com/klokantech/openmaptiles/tree/adjust_zres) branch.

I have in mind one more change: adjusting ZRES also for other layers, specifically for `sql_filter: area>power(ZRES8,2)` cases.
Right now it actually eliminates only polygons that are smaller than 1 pixel for given resolution. But one-pixel feature is usually more like a mess in the map than like holder of information. What i am thinking about is to shift ZRES in area SQL filters about 1 zoom level (meaning polygons with area lower than 4 pixels will be eliminated).

stirringhalo:
@jirik 
Great work identifying the culprits, not too surprised at it either!
I agree about the 1 pixel size, it's hardly a useful feature if it's 1 pixel in size. 4 pixels is probably a good threshold.
We'll have to tweak the names (_genXX) as it will get confusing with differing ZRes tolerance and SQL_Filters for a zoom level. But that's for 3.7+

PR #219, opened 2017-04-13
Move Travis from Precise to Xenial

stirringhalo:
Precise will EOL April 27th! Not sure where else Precise is being used in Travis tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: required
 
-dist: precise
+dist: xenial
 
 language: bash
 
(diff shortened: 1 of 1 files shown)


ImreSamu:
@stirringhalo 
it would be ideal ( for supporting basic technical questions )  -  
if we can test with the minimal system requirements ( now:  docker-engine=1.10  )

So my suggestion  
*  change the minimal system req to  `docker-engine=1.12`
* OR   change the travis to `docker-engine=1.10`

:)    

As I see -  the current travis : 
```
Docker version 1.12.3, build 6b644ec
docker-compose version 1.7.1, build 0a9ab35
```

stirringhalo:
See https://github.com/openmaptiles/openmaptiles/compare/master...xenial2

PR #283, opened 2017-06-17
Issue 210 - Add more detail at z7

stirringhalo:
This pushes landcover down to z7 and tweaks city labelling density. Needs testing. This is for https://github.com/openmaptiles/openmaptiles/issues/210


jirik:
@stirringhalo Thanks, it seems promising. Together with landcover it would be nice to show also osm_ocean_polygon instead of ne_10m_ocean at Z7 (similar to #205).

Issue #225, opened 2017-04-21
Buildings visible on z13 disappear on z14 


klokan:
`#13.99/46.7507/7.6073`

![unnamed 10](https://cloud.githubusercontent.com/assets/59284/25276078/b6fc401e-2698-11e7-9a88-a98451bd3678.png)

`#14.01/46.7507/7.6073`

![unnamed 11](https://cloud.githubusercontent.com/assets/59284/25276081/b949a136-2698-11e7-9fa5-e571a0127d35.png)

This seems to be a bug in the schema. To be fixed in the next release, if possible.

stirringhalo:
Issue might be in https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/building.sql#L18 .  That function is used to properly handle complex building relations like the Eiffel tower. Might also be a bug in the tagging of those buildings.

They are tagged as building="house" or as in http://www.openstreetmap.org/way/246312545, building='yes' but part of the greater relation http://www.openstreetmap.org/relation/3327042 as building='house'. 

Likely related to https://github.com/openmaptiles/openmaptiles/issues/228 .

stirringhalo:
Other more major issue is those buildings are in a relation with the road, and it's tagged as "street", leading to my building code not handling it. Fixing the relation handling will probably break 3D buildings. Thoughts?

https://www.openstreetmap.org/relation/3327042

stirringhalo:
Buildings will need to be better handled, major undertaking. Pushing to 3.7

stirringhalo:
@ImreSamu @klokan @jirik 
The only way I can see to fix this in Imposm3 is to change https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/mapping.yaml#L153 to be any type and then perhaps filter from there. I will try it, I don't expect it to work.

stirringhalo:
Trying again here https://github.com/openmaptiles/openmaptiles/pull/284

stirringhalo:
Fixed, except for some artifacting. Tracking that issue separately.

klokan:
This bug is CRITICAL!

It affects whole towns (such as Winterthur in Switzerland) and it is estimated that over 100k buildings gone missing in vector tiles probably! The original bug report in April sounds like only a few wrongly tagged buildings were wrong...

The bug was introduced in v3.5 during the changes on 3D buildings #128, and it was not fixed in time for the v3.6 release.

[...]

PR #284, opened 2017-06-17
Import street and associatedStreet to get the houses

stirringhalo:
This will need some thorough testing.


stirringhalo:
Some validity issues (st_asmvt/asmvtgeom converts linestrings to conflicting multilinestrings) Going to track that issue separately and close #225

PR #302, opened 2017-07-03
quickstart.sh support macOS

smellman:
My latest fix for macOS ( #164) enable to build tile but quickstart.sh stop before output all messages.

- Check GNU Coreutils stat command and if fail use BSD stat command.
- Fix ls option position


ImreSamu:
@smellman : Thank you!

Looks Good To Me;    ( but only tested on Ubuntu )

PR #326, opened 2017-09-12
Add brunnel to waterway layer

bobwallis:
Can tunnel/bridge tags be added to the waterway layer? This is needed for displaying things like aqueducts.

This PR adds a 'brunnel' column consistent with the transportation layer (coming in from z12 onwards).

### DIFF
--- a/layers/waterway/mapping.yaml
+++ b/layers/waterway/mapping.yaml
@@ -14,6 +14,20 @@ generalized_tables:
     source: waterway_linestring
     sql_filter: waterway IN ('river')
     tolerance: ZRES11
+
+tunnel_field: &tunnel
+  key: tunnel
+  name: is_tunnel
+  type: bool
+bridge_field: &bridge
+  key: bridge
+  name: is_bridge
+  type: bool
+ford_field: &ford
+  key: ford
+  name: is_ford
+  type: bool
+
 tables:
   # etldoc: imposm3 -> osm_waterway_linestring
   waterway_linestring:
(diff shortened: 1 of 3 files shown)


jirik:
@bobwallis Thanks, this PR seems very well. I have two suggestions:
- Rename function `brunnel` to someting else, e.g. `brunnel_waterway`, so that it does not share name with current `brunnel` used in transportation.
- Remove fords from mapping.yaml, osm_waterway_linestring, and waterway.sql. AFAIK `ford=yes` makes sense for `highway`, but not for `waterway`, so it's not needed in this case.

What do you think about it?

bobwallis:
Thanks Jiri. These suggestions are great, I will find some time to make the
changes and resubmit.

On 7 Nov 2017 12:59 pm, "Jiri Kozel" <notifications@github.com> wrote:

@bobwallis <https://github.com/bobwallis> Thanks, this PR seems very well.
I have two suggestions:

   - Rename function brunnel to someting else, e.g. brunnel_waterway, so
   that it does not share name with current brunnel used in transportation.
   - Remove fords from mapping.yaml, osm_waterway_linestring, and
   waterway.sql. AFAIK ford=yes makes sense for highway, but not for
   waterway, so it's not needed in this case.

What do you think about it?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openmaptiles/openmaptiles/pull/326#issuecomment-342474799>,
or mute the thread

klokan:
Done in #393

PR #328, opened 2017-09-25
Added OSM ref tag for aeroway layer

claustres:
So that for instance the runways/taxiways ID can be displayed in styles.

### DIFF
--- a/layers/aeroway/aeroway.yaml
+++ b/layers/aeroway/aeroway.yaml
@@ -6,6 +6,7 @@ layer:
       other airport related polygons can be found in the **aeroway** layer.
   buffer_size: 4
   fields:
+    ref: The OSM [`ref`](http://wiki.openstreetmap.org/wiki/Key:ref) tag of the runway/taxiway.
     class:
       description: |
           The original value of the [`aeroway`](http://wiki.openstreetmap.org/wiki/Key:aeroway) tag.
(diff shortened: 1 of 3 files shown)


claustres:
Example of how it is used in OSM : https://www.openstreetmap.org/#map=15/43.6319/1.3569

jirik:
Thanks @claustres for such clean PR!

claustres:
Thanks, will it be available in the next updates of OpenStreetMap vector tiles on OpenMaptiles.com ?

jirik:
It will be available after next release (not update) at openmaptiles.com. The next release is planned in November.

claustres:
We just downloaded the last version of the OSM MB tiles here https://openmaptiles.com/downloads/tileset/osm/ and the name *osm-2017-11-20-v3.6.2-planet.mbtiles* suggests it used the 3.6.2 release.

The [changelog](https://github.com/openmaptiles/openmaptiles/compare/v3.6.2...master) also suggests that this PR has been integrated. However when inspecting the tiles with TileserverGL we can't see the `ref` field, is it normal ?

Thanks for your return.

jirik:
It will be available in 3.7 release. We are rendering and making test in these days. It should be released next week and deployed the week after, you'll find it in changelogs.

claustres:
OK thanks

PR #332, opened 2017-10-18
Update QUICKSTART.md

hfu:
Fix tile URL served from postserve


hfu:
sorry, please let me withdraw this and let me put a comment in the right place.

PR #339, opened 2017-10-30
Fix class name for rail pois

nlehuby:
For now, the tram and subway stations are not rendered (with OSM Bright and OSM Liberty styles)

The openmaptiles class is `railway`
The maki icon is `rail` 

This PR change the class from `railway` to `rail`

### DIFF
--- a/layers/poi/class.sql
+++ b/layers/poi/class.sql
@@ -4,7 +4,7 @@ RETURNS INT AS $$
         WHEN 'hospital' THEN 20
         WHEN 'park' THEN 25
         WHEN 'cemetery' THEN 30
-        WHEN 'railway' THEN 40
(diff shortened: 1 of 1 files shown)


jirik:
@nlehuby Thanks for pointing this out.

What about changing style instead of schema (I mean renaming the icon)? Because we want to preserve backward schema compatibility if possible.

nlehuby:
I think it would make no sense to change the maki icon name from `rail`to `railway`as it does not represent a railway (this looks much more like a train to me ...). But we can update the styles.

Do you want a pull request for the OSM Bright and OSM Liberty gl styles ?

jirik:
Yes, please, PR for renaming SVG icon in OSM Bright / OSM Liberty repos are welcome. That's what I meant.

PR #340, opened 2017-10-30
fix class for bicycle_rental pois

nlehuby:
This PR changes `bicycle_rental` class to `bicycle_share` because the maki icon has got this name.

Not sure this is the right way to handle this...


jirik:
I prefer to rename the icon, as in case of https://github.com/openmaptiles/openmaptiles/pull/339

nlehuby:
ok : https://github.com/mapbox/maki/pull/398

PR #382, opened 2017-12-07
Handle theme_park as a landuse

nlehuby:


nlehuby:
If this is ok for you, I think we may also add these features as landuse:
* tourism=zoo
* amenity=bus_station

They may be large polygons that shape the territory, and should be seen as polygon from low zoom levels.

jirik:
Naomie, thanks, this PR looks really good. I like your idea to add also
- tourism=zoo
- amenity=bus_station

Would you like to add it to this PR? Or can I merge it now?

nlehuby:
Thank you for the review.
You can merge it now, I won't work again on that topic before a few weeks.

Do you want me to make an issue about the other features we could handle in the same way, so we can add them all in the same PR ?

nlehuby:
Zoo and Bus station added.

jirik:
Thanks! I've merged it and updated documentations (zoo and bus_station added to list of classes).

PR #391, opened 2017-12-21
Add a few missing POIs

nlehuby:
* hackerspace
* escape room
* subway and station entrance
* frozen foods shop


layers/poi/class.sql
```diff
@@ -32,14 +32,15 @@ $$ LANGUAGE SQL IMMUTABLE;
...
         WHEN subclass IN ('townhall','public_building','courthouse','community_centre') THEN 'town_hall'
         WHEN subclass IN ('golf','golf_course','miniature_golf') THEN 'golf'
         WHEN subclass IN ('fast_food','food_court') THEN 'fast_food'
         WHEN subclass IN ('park','bbq') THEN 'park'
         WHEN subclass IN ('bus_stop','bus_station') THEN 'bus'
-        WHEN (subclass='station' AND mapping_key = 'railway') OR subclass IN ('halt', 'tram_stop', 'subway') THEN 'railway'
+        WHEN (subclass='station' AND mapping_key = 'railway') OR subclass IN ('halt', 'tram_stop') THEN 'railway'
```
jirik: Please do not remove `'subway'`, we need it. It's not in the mapping file, but it is created in [update_poi_point.sql](https://github.com/openmaptiles/openmaptiles/blob/3cee3c7f80ce0107f91951b675bdb6da59f59d53/layers/poi/update_poi_point.sql#L7).

jirik:
@nlehuby Thanks for this PR. Please only return back the `'subway'`, and I will merge it.

layers/poi/class.sql
nlehuby: done, sorry !

nlehuby:
ok, done !

Issue #376, opened 2017-11-29
importing is_oneway as bool loses critical information


freeman40:
In the transporation layer is_oneway field is boolean.  It is imported from the oneway key.  Conversion of the oneway key to boolean loses critical information (such as the direction in which the highway is oneway). refer http://wiki.openstreetmap.org/wiki/Key:oneway

freeman40:
See https://www.openstreetmap.org/way/245394629
and https://www.openstreetmap.org/way/245394603
for example of where this causes a problem.

jirik:
This is probably caused by [`oneway=-1`](http://wiki.openstreetmap.org/wiki/Key:oneway) rare cases (0.74% of `oneway` values)

freeman40:
@jirik : indeed that is what causes the problem.

The simplest fix would be to just add the raw oneway key into the transportation layer mapping.yaml file, as follows:  (plus similar edits to layer.sql)

It does feel a bit like a "hack".  An alternative solution is to completely replace the existing "is_oneway" boolean  field with the string typed "oneway" field.
The problem here though is that it will have downstream consequences:  Any styles based on is_oneway will be broken.


```yaml
tables:
# etldoc: imposm3 -> osm_highway_linestring
  highway_linestring:
    type: linestring
    fields:
    - name: osm_id
      type: id
    - name: geometry
      type: geometry
    - name: highway
      key: highway
      type: string
    - name: maxspeed
      key: maxspeed
      type: string
    - *ref
    - *network
    - *z_order
    - *layer
    - *level
    - *indoor
    - *name
    - *name_en
    - *short_name
    - *tunnel
    - *bridge
    - *ramp
    - *ford
    - *oneway
    - name: oneway   }--- add this
      key: oneway    }
      type: string   }
    - *area
    - *service
    - *usage
    - *public_transport
    mapping:
      highway:
      - motorway
      - motorway_link
      - trunk
      - trunk_link
      - primary
      - primary_link
      - secondary
      - secondary_link
      - tertiary
      - tertiary_link
      - unclassified
      - residential
      - road
      - living_street
      - raceway
      - track
      - service
      - path
      - cycleway
      - bridleway
      - footway
      - corridor
      - pedestrian
      - steps
      public_transport:
      - platform
```

frodrigo:
An other solution is to extend the values from [0,1] to [-1, 0, 1]. Even if not take care of all case, it's include the important case of "-1".

jirik:
[...]

PR #392, opened 2017-12-29
Use direction as type for oneway, support oneway-opposite #376

frodrigo:
Fix the issue #376 of losing some oneway directions.


jirik:
Thanks, @frodrigo, I am going to review it today.

jirik:
Looks very good to me. Thanks again, merging.

PR #415, opened 2018-01-22
Add layer for all highway

hanchao:
To fix this problem.
layer

https://www.openstreetmap.org/way/226830432
layer1


jirik:
@hanchao Thank for this PR. In the meantime I've implemented it in a bit more complex way. The `layer` attribute was added for all transportation features on all zoom levels now. See https://github.com/openmaptiles/openmaptiles/pull/441/files for details.

Issue #410, opened 2018-01-16
Why is there no sand or beach


boldtrn:
I think this is a very important tag, but I couldn't find it?

jirik:
It seems reasonable to add it into landcover. Pull request is welcome.

jirik:
Already implemented: https://github.com/openmaptiles/openmaptiles/commit/51d4558ea8c7002be704674180f61f77d215f728#diff-aeed953ebcdeefcc353f5ee25b00fe65

It will be available in the next release.

PR #418, opened 2018-01-28
add sand and beach to landcover

hanchao:
Related to https://github.com/openmaptiles/openmaptiles/issues/410


jirik:
Thanks. Could you please add new class and subclass values also to [documentation](https://github.com/openmaptiles/openmaptiles/blob/master/layers/landcover/landcover.yaml) as a part of this PR? I will merge it then.

jirik:
Thanks, merging

PR #433, opened 2018-03-27
Remove duplicates on shop and art POI class

nlehuby:


jirik:
Thanks, Noémie, merging

PR #452, opened 2018-04-30
add aquarium to the poi layer

nlehuby:
And we already have an icon for those in osm-bright-gl-style ;)


jirik:
Thanks, @nlehuby

PR #470, opened 2018-06-27
add type of sport as subclass for pitch poi

nlehuby:
This PR add the type of sport (from the `sport` tag) as a subclass for pitch poi that have this tag.

This allow for better filtering in the style according to the sport which can be played on the pitch.


jirik:
@nlehuby Thanks. However with this change we are loosing information if feature is a `"pitch"` (class become `"stadium"` and subclass become value of sport tag).

So, what about removing `"pitch"` from `"stadium"` class to make it standalone class?
https://github.com/openmaptiles/openmaptiles/blob/acd489f04b4592c1af702b86da984993768c4226/layers/poi/class.sql#L61

Also, I suggest to leave `"pitch"` value in subclass if sport is empty string/NULL.

jirik:
This was implemented in #532

PR #525, opened 2018-11-07
Remove park_label layer, add labels to park layer

### LINKED ISSUE #513: No national park names
Hi,

On all the world map, each national park and natural reserve doesn't have any name. I checked with the inspect mode and I can confirm that the names are not in the schema/tiles data (at least the point to localize the name is not there), so it's not a style problem. Also, those names appear correctly on OpenStreetMap (you can check for any national park in the world).

It's a huge miss, especially when we want to use the map for tourism... :-(

Any plan to add the national park and natural reserve names in the tiles/schema? On my side, I don't have the skill to tweak openmaptiles script to add them. I able to run it however on our server and edit the code files, if someone can tell me which one to edit.

Thank you in advance,

Frankkky

jirik:
Fix #513


jirik:
Preview:
image

typebrook:
good to see avoiding one more layer, but does this means `aerodrome_label` could be merged into `aeroway`? 

Or maybe it is not a good idea since `aerodrome_label` is a subset and needs more description?

Issue #420, opened 2018-01-30
Missing volcanos


ladislavhorky:
Hello,

when using the tiles through tilehosting.com, we noticed that `mountain_peak` layer contains only OSM features with `natural:peak` tag and not with `natural:volcano` tag. As noted in OSM [volcano definition](https://wiki.openstreetmap.org/wiki/Tag:natural=volcano?uselang=en-US#Relation_to_natural.3Dpeak), there is a certain distinction on how volcano vs peak features are placed but the way it is currently done in OSM database causes many very significant peaks (e.g. [Etna](https://www.tilehosting.com/#topo//vector/12.58/14.99829/37.75015)) to have no label if only `natural:peak` features are used.

Proposed solution is to either add volcanos to `mountain_peak` layer and possibly add a tag to distinguish them from normal peaks, or create a completely new layer with volcanos. Seeing that other layers contain whole groups of features, the first option could be more appropriate.

jirik:
+1 for adding volcanoes to `mountain_peak` layer. PR is welcome

ladislavhorky:
Hi @jirik ,

do you know if there has been any progress regarding this issue?

jirik:
I don't know about any progress on this issue, we have currently other priorities. Anyway, pull request is welcome.

PR #535, opened 2018-11-17
Adding tag natural=volcano to the layer mountain_peak; issue #420

evka2830:
* https://github.com/openmaptiles/openmaptiles/issues/420
* adding tag natural=volcano into mountain_peak layer
* tested on data of Italian islands (isole-latest.osm.pbf)
-- before: 3278 records in osm_peak_point
-- after: 3327 records
-- e.g. 
```
SELECT * FROM import.osm_peak_point
WHERE name = 'Etna' OR name = 'Stromboli';
```
gives results now.

### DIFF
--- a/layers/mountain_peak/mapping.yaml
+++ b/layers/mountain_peak/mapping.yaml
@@ -28,3 +28,4 @@ tables:
     mapping:
       natural:
       - peak
+      - volcano


evka2830:
Basic style with layer mountain_peak_volcano in Maputnik json for a check:
[klokantech-basic.json.txt](https://github.com/openmaptiles/openmaptiles/files/2597071/klokantech-basic.json.txt)

MartinMikita:
Thanks for your PR, but it only solves part of this issue.
It is required to add a new logic to `distinguish them from normal peaks`, so they could be styled separately.

If you are interested, create a new PR.

Issue #405, opened 2018-01-11
including trees


musicformellons:
OSM contains trees. Would OMT consider including these in its schema?

klokan:
Why not. Make a pull request ;-)

frodrigo:
Back from my pull request to add trees #554 
Should add a new layer for tree? or include them in existing one, maybe poi?

MartinMikita:
Hi @frodrigo,

Thanks for your PR and contribution.

However, as pointed by @typebrook, the new layer is not a great idea. You should import trees into one of the existing layers, probably `landcover`, `landuse`, or `park`.
You would need to create a new table with nodes, not include them into the existing tables, as they are polygons.

Please, could you also compare, what effect on the sizes would this PR means? Size of MBTiles, size of Database, Time of importing and rendering tiles. Before and after the change.
According to OSM, there are 11 million nodes of Trees...

Thanks for your contribution!

cbeddow:
@frodrigo @MartinMikita it seems `landcover`, `landuse`, and `park` are all polygons so adding a point type to them is a bit out of place, and a tree is not really any of these.

I see `mountain peak` as its own data layer. What if we merge this with tree, to have a `landforms` dataclass? There are plenty of future tags that can go in this:

- coral reef
- valley
- boulder
- shrub

Could we attempt this?

PR #554, opened 2018-12-20
Support trees from natural=tree #405

frodrigo:
Add support of natural=tree, without any specific tags.

**NOTE: I do not include the mapping_diagram.png, how to get it ?**

image


daliborjanak:
Thanks for PR. It does not make sense to add it as a new layer. It is not often used thing and not especially used in base maps. It is a candidate for special external layer.

frodrigo:
What are external layer ? some kind of contrib layer ?

typebrook:
> What are external layer ? some kind of contrib layer ?

@frodrigo 
Means a vector tiles source **not in openmaptiles**. Consider tile size, we should only include layers for **general user cases**. Of course we still can make other layers in additional tile source.

For example, see [contour](https://github.com/openmaptiles/klokantech-terrain-gl-style/blob/014d553f2d7358c921a897590e43b04cfe4745eb/style.json#L29-L31) used in `klokantech-terrain-gl-style`

eva-j:
Please create fork for this topic, if you want. We can't merge this into general OMT schema, because the tile size would increase a lot and not many people would use it...

cbeddow:
Made a pull request here, but it may need some extra specification of the styling because an icon does not exist

https://github.com/openmaptiles/openmaptiles/pull/1545

PR #555, opened 2018-12-21
Add colour to buildings

frodrigo:
Add colour to building from `building:colour` tag and colour derived from `building:material` tags as fallback.
3dcolour


frodrigo:
Thank.
Interested on PR to support colour in klokantech-3d-gl-style ?

Issue #558, opened 2018-12-22
Use same docker version of "import-osm"


typebrook:
Since `import-osm:0.5` is introduced in #356 for docker-compose service `import-osm`, it is weird to keep `import-osm:0.4` as another docker image for services `import-osm-diff` and `update-osm`. For now it is fine to use newer version without any bug.

MartinMikita:
Yes, thanks for the notice. We have merged your PR #559.

Thanks for your contribution!

PR #559, opened 2018-12-22
Update older import-osm docker image

typebrook:
#558 
Remove `import-osm:0.4` since newer version is here.


MartinMikita:
Thanks for your PR!

Issue #565, opened 2019-01-04
Bicycle and motorcycle parkings


Phyks:
Hi,

It seems that currently both bicycle and motorcycle parkings are ignored from [the amenities included in the tiles](https://github.com/openmaptiles/openmaptiles/blob/master/layers/poi/mapping.yaml#L10).

Having them would be super useful for cycle-oriented render, but as well for regular renders (they are shown on numerous renders including the one on osm.org).

Could OMT include the `amenity=bicycle_parking` and `amenity=motorcycle_parking` tags? I could make a PR for this.

Best,

Phyks:
I have a branch ready for this [here](https://github.com/phyks/openmaptiles/tree/parking).

On Albania, resulting `tiles.mbtiles` (at zoom level 14) is 106369024 bytes (versus 106508288 for the current `master`).

MartinMikita:
Hi Phyks,

could you compare the size of Switzerland and Czechia (Czech republic) regions, with zooms 0-14?
It seems, these tags should not increase it a lot, but your comparison could help us to include it earlier.

Prepare a PR, please.

Thanks for your contribution!

francois2metz:
This issue can now be closed.

PR #567, opened 2019-01-07
Add parking, bicycle_parking and motorcycle_parking as POIs

Phyks:
Hi,

As discussed in https://github.com/openmaptiles/openmaptiles/issues/565, this PR adds support for `bicycle_parking` and `motorcycle_parking` as amenities.

On Albania, resulting tiles.mbtiles (at zoom level 14) is 106369024 bytes (versus 106508288 for the current master).

I'm trying to get numbers on Switzerland and Czechia at the moment.

Best,

### DIFF
--- a/layers/poi/mapping.yaml
+++ b/layers/poi/mapping.yaml
@@ -39,6 +39,7 @@ def_poi_mapping_amenity: &poi_mapping_amenity
   - motorcycle_parking
   - nightclub
   - nursing_home
+  - parking
   - pharmacy
   - place_of_worship
   - police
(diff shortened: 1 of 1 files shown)


Phyks:
@MartinMikita on Switzerland, I get 428 kB difference (total tiles size of 268 MB).

My server is not very poweful and processing is a bit long. I'll try to get some numbers for Czechia though.

Phyks:
For Czech Republic, I get 372 kB difference (total tiles size of 519 MB).

MartinMikita:
Thanks for your PR and simple benchmark.
We would like to add also normal `parking` POI, see #592

Issue #422, opened 2018-02-03
path, footway, cycleway ambiguity


chen-ye:
There's still a lot of ambiguity in OSM about how different kinds of paths should be coded (see: https://wiki.openstreetmap.org/wiki/Path_controversy). 

There's also a wide variety of features coded as "paths" that should probably be differentiated from each other, especially on non-automotive maps (maps for cycling, hiking, etc): https://wiki.openstreetmap.org/wiki/Path_examples.

To differentiate these, could OMT include the bicycle/foot/horse and surface: tags from OSM?

klokan:
Yes it could - make a proposal for modification of the schema please.

Phyks:
Hi,

I am interested in this issue as well, and this would likely solve https://github.com/openmaptiles/openmaptiles/issues/512.

Concerning the surface, I would add a `surface` field to the transportation layer. I can make a PR for this.

Is adding the bicycle/foot/horse fields as well fine? I guess these should be added only on path highways?

Thanks

MartinMikita:
Hello @Phyks,

Yes, you can add a new attribute `surface` to distinguish these types in the `transportation` layer. Bicycle, foot, horse should be enough for now.

Please, prepare a PR, and we would appreciate, if you could compare what would be the final effect - the size of MBTiles, size of the database, time of importing and rendering (before and after your change).
You could start with smaller states like Albania, and probably prepare and compare the results on Switzerland or Czech Republic, these have a lot of bicycle ways.

This could help us to decide if we would include it in the official OpenMapTiles scheme in the next release (v3.10).

zstadler:
The approach for handling `bicycle = designated` implemented in https://github.com/openmaptiles/openmaptiles/pull/573 is different than the approach requested in issue #512. 

As @nnhubbard [noted](https://github.com/openmaptiles/openmaptiles/issues/512#issuecomment-430982350), the different values of the bicycle tag should be preserved so they can be styled differently if needed.

Issue #389, opened 2017-12-15
Add surface for highways


pertile:
Hi, I think it would be great getting surface key in the tiles. In non-developed countries is a very useful tag to make a difference between paved and unpaved ways.

jirik:
It seems reasonable. We can generalize it into two values only: paved and unpaved.

Pull request is welcome.

TomPohys:
close by https://github.com/openmaptiles/openmaptiles/pull/570

PR #570, opened 2019-01-09
Add surface field for highways

Phyks:
Hi,

This PR keeps surface field from OSM on highways, generalize it to two values:
"paved" and "unpaved".

This is a fix for #389 and a partial fix for #422.

Here are some benchmarks (for tiles up to zoom 14):

## Albania

Total tile size with `master` branch: 102MB.
Size increase with this PR: 60kB.

## Czech Republic

Total tile size with `master` branch: 519MB.
Size increase with this PR: 1.19MB.

## Switzerland

Total tile size with `master` branch: 268MB.
Size increase with this PR: 1.25MB.

Best,

### DIFF
--- a/layers/transportation/class.sql
+++ b/layers/transportation/class.sql
@@ -42,3 +42,13 @@ CREATE OR REPLACE FUNCTION service_value(service TEXT) RETURNS TEXT AS $$
         ELSE NULL
     END;
 $$ LANGUAGE SQL IMMUTABLE STRICT;
+
+-- Limit surface to only the most important values to ensure
+-- we always know the values of surface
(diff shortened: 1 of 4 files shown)


nnhubbard:
I just tested this. Is it just me or are there not many roads that have the `surface` tagging? Or is it just my area? I was looking at Washington, USA.

Phyks:
From my experience, surface is not often mapped in urban areas where there is asphalt everywhere and contributors are not very eager to add the same surface again and again. You can check with an Overpass query such as https://overpass-turbo.eu/s/F9o.

eva-j:
I've just updated the tag lists according to Wiki https://wiki.openstreetmap.org/wiki/Key:surface.

PR #573, opened 2019-01-12
Add bicycle/foot/horse tags for highways=path|cycleway|footway

Phyks:
Hi,

This is a fix proposal for
https://github.com/openmaptiles/openmaptiles/issues/422.

Here are some benchmarks on tiles up to zoom 14.

## Albania

No noticeable tile size increase.

## Switzerland

Before this PR, 267.8MB.
After this PR, 268.7MB.
Increase of 848kB.

## Czech Republic

Before this PR, 518.7MB.
After this PR, 519.3MB.
Increase of 596kB.

Best,

### DIFF
--- a/layers/transportation/class.sql
+++ b/layers/transportation/class.sql
@@ -42,3 +42,30 @@ CREATE OR REPLACE FUNCTION service_value(service TEXT) RETURNS TEXT AS $$
         ELSE NULL
     END;
 $$ LANGUAGE SQL IMMUTABLE STRICT;
+
+-- Limit bicycle to only the most important positive values
+CREATE OR REPLACE FUNCTION bicycle_value(bicycle TEXT) RETURNS BOOL AS $$
+    SELECT CASE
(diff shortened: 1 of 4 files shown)


zstadler:
In bike maps, it would be useful to keep the original value of the `bicycle` tag rather than simplifying into Boolean value. This will enable different rendering for `yes`, `designated`, `dismount`, etc.
When simplification is needed, it can be done in the rules of the rendering style.

ref:  [bicycle restrtictions](https://wiki.openstreetmap.org/wiki/Bicycle#Bicycle_Restrictions) wiki page

Phyks:
Hi,

I agree with your remark @zstadler. I however am no longer using OpenMapTiles schemes and will likely not have much time for the back and forth rounds of reviews. Feel free to take over this PR if you think this is worth requesting a merge :)

Best,

zstadler:
Hi @Phyks,

Following you proposal, I've create PR https://github.com/openmaptiles/openmaptiles/pull/604 that hopefully could take over this PR.

PR #578, opened 2019-01-27
Add ski resorts to poi and landuse layers

nlehuby:
This PR adds the ski resorts :skier:  to the `poi` layer. As they usually occupy much spaces, they also have been added to the `landuse` layer, so they can have some special colour from low zooms.

https://wiki.openstreetmap.org/wiki/Tag:landuse=winter_sports

Please let me know if you also want a PR to display them in the styles :wink:

### DIFF
--- a/layers/landuse/landuse.yaml
+++ b/layers/landuse/landuse.yaml
@@ -32,6 +32,7 @@ layer:
       - playground
       - theme_park
       - bus_station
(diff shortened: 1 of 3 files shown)


daliborjanak:
Thanks for PR. We decide to add ski resorts only as POIs, not a polygon. We will merge the only this.

Issue #548, opened 2018-12-18
make Wikidata optional


typebrook:
@jirik @MartinMikita 

According to description in https://github.com/openmaptiles/openmaptiles/pull/358, importing Wikidata should be optional,  I made a simple PR at openmaptiles/import-sql to achieve this. Here is the link: https://github.com/openmaptiles/import-sql/pull/8

For now, if user doesn't run script like `docker-compose run --rm import-wikidata`
docker-compose service `import-sql` always fails at https://github.com/openmaptiles/import-sql/blob/f9a84094179b585889bc9a86d45cddc726925fd6/language.sql#L122-L135 because table for WIkidata doesn't exist.

The newly added line just create table if it doesn't exitst, it also fix this issue about Wikidata https://github.com/openmaptiles/openmaptiles/issues/454
If this PR is accepted, I'll add description of Wikidata in READEME for both using it or not  (for now it is 0...), and upgrade docker image version in `docker-compose.yml`

typebrook:
@MartinMikita 
Since https://github.com/openmaptiles/import-sql/pull/8 is merged, please make a new docker image for [`openmaptiles/import-sql`](https://hub.docker.com/r/openmaptiles/import-sql/tags) so I can continue working on this issue, thanks!

nyurik:
update: import-sql has been merged with [tools](https://github.com/openmaptiles/openmaptiles-tools#importing-into-postgres). The original repo's issues should be migrated, and the repo archived.

TomPohys:
Import-wikidata are optional and can be import into PostgreSQL from backup (resolved by https://github.com/openmaptiles/openmaptiles-tools/commit/9e011179ba03bee6eaa544a1fffa689b10bdd729 and https://github.com/openmaptiles/openmaptiles-tools/commit/4ed919770276a1aad949c7ddfe4bbfa0b68e3196)

PR #588, opened 2019-02-28
import-sql won't work without import-wikidata

iruoy:


typebrook:
Just because there is no table for wikidata used in sql script you imported, make a new empty one in database will solved.
`docker-compose run --rm import-osm /usr/src/app/psql.sh -c "CREATE TABLE IF NOT EXISTS wd_names(id varchar(20), page varchar(200), labels hstore);"`

I already opened issue #548 for this and have one line fix PR in https://github.com/openmaptiles/import-sql/pull/8

@MartinMikita @jirik @daliborjanak 
Could you guys handled this first? Should be a simple quick fix with no side effect

MartinMikita:
Thanks @typebrook, we have merged your request, which solves the problem with optional import of wikidata.

MartinMikita:
There is now no need to change the basic documentation.

PR #600, opened 2019-04-01
Optional wikidata (need import-sql:0.9)

typebrook:
### [Caution] This PR should only be merged when image of [`import-sql:0.9`](https://github.com/openmaptiles/import-sql/releases/tag/v0.9) is on [DockerHub](https://hub.docker.com/r/openmaptiles/import-sql/tags)

#548 

- Use `import-sql:0.9`, which allows no Wikidata in psql
- Add option `--no-wikidata` for `quickstart.sh`, to skip importing Wikidata.
- Also add description about Wikidata in `README.md` and new suggestion in `QUICKSTART.md`.

Latest Wikidata is about 50GB, which means it is not **quick** for newcomers who just want to try OpenMapTiles. So excluding it from `quickstart.sh` is reasonable. But since travis CI use `quickstart.sh` to do build check, I think one more option is the simplest way I can do.

By the way, add option `--wikidata` and skip importing by default is also fine in this case, any idea?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -68,7 +68,7 @@ services:
      - ./build:/mapping
      - cache:/cache
   import-sql:
-    image: "openmaptiles/import-sql:0.8"
+    image: "openmaptiles/import-sql:0.9"
     env_file: .env
     networks:
     - postgres_conn
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -68,7 +68,7 @@ services:
      - ./build:/mapping
      - cache:/cache
   import-sql:
-    image: "openmaptiles/import-sql:0.9"
+    image: "openmaptiles/import-sql:1.0"
     env_file: .env
     networks:
     - postgres_conn
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -39,9 +39,10 @@ Important:  The ./quickstart.sh is for small extracts - not optimal for a Planet
 ```bash
 git clone https://github.com/openmaptiles/openmaptiles.git
 cd openmaptiles
-./quickstart.sh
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -68,7 +68,7 @@ services:
      - ./build:/mapping
      - cache:/cache
   import-sql:
-    image: "openmaptiles/import-sql:0.8"
+    image: "openmaptiles/import-sql:0.9"
```
ldgeo: it seems that the docker hub has been updated with the latest import-sql code (https://hub.docker.com/r/openmaptiles/import-sql/tags) but the version jumped to `1.0` instead of `0.9`

Tested with the change above and it worked well for me

```suggestion
    image: "openmaptiles/import-sql:1.0"
```

typebrook:
> it seems that the docker hub has been updated with the latest import-sql code (https://hub.docker.com/r/openmaptiles/import-sql/tags) but the version jumped to `1.0` instead of `0.9`
> 
> Tested with the change above and it worked well for me

Thanks, already added a new commit for that.
@ldgeo @MartinMikita 
Then it is fine to merge it now

typebrook:
Wow, I was wrong.  In `quickstart.sh`:
```
docker-compose run --rm import-wikidata
```
will skip import when dump file doesn't exist, see
https://github.com/openmaptiles/import-wikidata/blob/1074ce6060aaf68ce99120d7fe9daec2d1fdecf5/bin/import-wikidata#L28-L42

So, `quickstart.sh` doesn't perform download for Wikidata dump. 
Add one more option makes no sense, I'll just close this PR and make a new one to just upgrade `import-sql`

PR #601, opened 2019-04-03
Adding volcanoes

eva-j:
- https://github.com/openmaptiles/openmaptiles/issues/420
- "natural" added to osm_peak_point - loading now 'peak' and 'volcano'
- edit mountain_peak layer in Topo style and add mountain_volcano layer; layers should then look like
[styl_mountain_peak_volcano.txt](https://github.com/openmaptiles/openmaptiles/files/3039347/styl_mountain_peak_volcano.txt)

### DIFF
--- a/layers/mountain_peak/layer.sql
+++ b/layers/mountain_peak/layer.sql
@@ -3,14 +3,14 @@
 -- etldoc:     style="rounded,filled", label="layer_mountain_peak | <z7_> z7+" ] ;
 
 CREATE OR REPLACE FUNCTION layer_mountain_peak(bbox geometry, zoom_level integer, pixel_width numeric)
-RETURNS TABLE(osm_id bigint, geometry geometry, name text, name_en text, name_de text, natural text, tags hstore, ele int, ele_ft int, "rank" int) AS $$
+RETURNS TABLE(osm_id bigint, geometry geometry, name text, name_en text, name_de text, "natural" text, tags hstore, ele int, ele_ft int, "rank" int) AS $$
    -- etldoc: osm_peak_point -> layer_mountain_peak:z7_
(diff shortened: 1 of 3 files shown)


typebrook:
Duplicate to #553, the difference is about name of new field in `mountain_peak` layer: `class`(my PR) and `natural`(This one).

eva-j:
You are right, we can now decide witch version we want to merge. Was there any reason for renaming field `natural` to `class`? I would prefer to keep consistency with OSM tagging and approach in other layers - it is used e.g. in water, landcover.

typebrook:
As my opinion, the use case of this layer should be includes markable natural landscape that user can recognize them from a long distance.

> Was there any reason for renaming field `natural` to `class`?

Because I am thinking maybe we need to include some man-made OSM features into this layer in the future, to show natural landscape which is not in OSM but markable. In this case, `class` would be a general field to use.

Some countries like Taiwan or Japan have `man_made=survey_point` for topography use case on unnamed peaks. For example [node 5696179414](https://www.openstreetmap.org/node/5696179414) only tagged as `man_made=survey_point` but in Google Map it is shown as mountain peak called [奥聖岳](https://goo.gl/maps/gRDMmskihSA2)

And according to [Mapbox Street v7](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v7/#mountain_peak_label), it is reasonable to includes `hilltop` inside this layer. So maybe things like `power lines` or `radio tower` on the top would be the better choice to be included.

Of course things above need to have some discussion before being part of OpenMapTiles schema, but I just think a general field name would be a better choice.

eva-j:
OK, this makes sense, I am closing this.

PR #611, opened 2019-04-17
Avoid long repeated text in poi mapping file

zstadler:
The `fields` and `mapping` sections for `poi_point` and `poi_polygon` are the same, and must be the same for the SQL to work properly.

Using definitions avoids repetition and the need to make each change in these sections twice, as was already done for the subsections of `mapping`.


nyurik:
Rebased onto the newest changes, and re-created a separate pull request in https://github.com/openmaptiles/openmaptiles/pull/689 , closing this one.

Issue #610, opened 2019-04-17
wrong admin_level 4 for France


nbozon:
The former admin_level 4 boundaries (old french regions) are still in use in OpenMapTiles, and that affects every style. The new regions (since 2017) available in OpenStreetMap should be used instead !

| <img width="299" alt="old-admin4-fr" src="https://user-images.githubusercontent.com/7135374/56273832-db13a180-60fd-11e9-9c22-b4a3198be1d1.png">  Old admin 4 | <img width="299" alt="new-admin4-fr" src="https://user-images.githubusercontent.com/7135374/56273916-09917c80-60fe-11e9-9015-4e565a98d2ef.png"> New admin 4 |
|---|---|

New geometries visible here: 
http://umap.openstreetmap.fr/en/map/contours-geographiques-des-nouvelles-regions-metro_43093#6/46.593/2.771
Details available here: 
https://en.wikipedia.org/wiki/Regions_of_France

eva-j:
Suggestions for update of zoom 4:

1. update attributes of lines in France in NE source ne_10m_admin_1_states_provinces_lines.shp:
https://github.com/nvkelso/natural-earth-vector/tree/master/10m_cultural
according to https://en.wikipedia.org/wiki/Regions_of_France#Reform_and_mergers_of_regions

2. use NE data with `min_zoom <= 5` only (not `7`) here:
https://github.com/openmaptiles/openmaptiles/blob/master/layers/boundary/boundary.sql#L46
this way regions in France (and other countries) are not displayed - only these countries are devided: USA, Canada, Brasil, Russia, China, India, Australia, Indonesia, South Africa

I vote for the second version, updates of NE are not really organized, as described in https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-1-states-provinces/, so there can be problems in other countries as well.
____

Another problematic zoom level is 5. I suggest to use the same content like in zoom 6 (not using NE input). This way there won't be too much data in z5 and there will not be too big difference between reduced z4 and z5. This solves problems in tickets https://github.com/openmaptiles/openmaptiles/issues/594 and https://github.com/openmaptiles/klokantech-basic-gl-style/issues/21.

frodrigo:
I think it could be closed. It is now solved.

Issue #594, opened 2019-03-22
admin_level 6 present only on zoom level 5


frodrigo:
On may countries the data for admin_level=6 are present, but only on zoom level 5, look at Spain, France, Greece... 

Zoom 5, then Zoom 6
![image](https://user-images.githubusercontent.com/1785486/54834781-284c5100-4cc1-11e9-88f4-6dd63f73a98d.png)

pathmapper:
This has been reported in https://github.com/openmaptiles/klokantech-basic-gl-style/issues/21, see also https://github.com/openmaptiles/openmaptiles/issues/586.

eva-j:
Solved in https://github.com/openmaptiles/openmaptiles/pull/613.

Issue #586, opened 2019-02-21
Mapping NE boundaries -> OSM admin level


pathmapper:
With the current [mapping](https://github.com/openmaptiles/openmaptiles/blob/master/layers/boundary/boundary.sql) of NE boundaries  -> OSM `admin_level`, it happens that boundaries of the same OSM `admin_level` are not available for the same zoom levels.

Example:

Zoom level 4
There are boundaries `admin_level=4` visible for Nigeria, but not for neighbouring countries like e.g. Niger, Chad or Cameroun:

![auswahl_536](https://user-images.githubusercontent.com/20856381/53191349-e5d51d00-360b-11e9-98c7-eca12019f424.png)

Zoom level 5
For these countries the boundaries `admin_level=4` are available starting with zoom level 5:

![auswahl_537](https://user-images.githubusercontent.com/20856381/53191640-a0651f80-360c-11e9-8caa-5fd5006e0e39.png)

Potentially related issues: 
https://github.com/openmaptiles/openmaptiles/issues/500
https://github.com/openmaptiles/klokantech-basic-gl-style/issues/21
https://github.com/openmaptiles/openmaptiles/issues/549

There are also other issues related to the `scalerank` which prevent updating from NE 4.0.0 -> 4.1.0, related issue:
https://github.com/openmaptiles/openmaptiles/issues/365

@nvkelso suggested in https://github.com/nvkelso/natural-earth-vector/issues/282:

> you should consider switching to the `min_zoom` and the `min_label` columns, they are curated for web mapping.

From the [NE 4.0.0 changelog](https://github.com/nvkelso/natural-earth-vector/releases/tag/v4.0.0):

> The v4 series introduces web map optimized min_zoom, min_label, and max_label properties that compliment the existing scalerank property that was originally curated for print cartography in proper projection glory. The min_zoom property is optimized for the web Mercator projection we are collectively hallucinating, and the min_label and max_label properties are optional on some themes when the label should be displayed later than the feature's geometry (or should stop showing).

Is this the way to go, a mapping based on `min_zoom` and `min_label` (and not on `scalerank` anymore)?

Or do you want to keep using also `scalerank`?

In any case, the mapping needs as revision.

JulianSSS:
I have a similar problem on the area of Cyprus. The island Cyprus is sparated into north (Turkish) and south (Cyprus).
[...]

Issue #500, opened 2018-09-24
Data missing in 4th zoom


Rajavelu:
Hi,
1. At 4th zoom, States name data not showing for South India. 
2. Telangana state boundary showing from 6th zoom only, but other state boundaries are showing from 4th zoom.

![sindiadata_telangana_border_issue](https://user-images.githubusercontent.com/3033600/45940192-26951f80-bff5-11e8-9d7d-5b9a20faa513.gif)

eva-j:
Solved in https://github.com/openmaptiles/openmaptiles/pull/613.

PR #613, opened 2019-05-03
Boundary - changes in z4 and z5

eva-j:
https://github.com/openmaptiles/openmaptiles/issues/610, https://github.com/openmaptiles/openmaptiles/issues/594, https://github.com/openmaptiles/openmaptiles/issues/586, https://github.com/openmaptiles/klokantech-basic-gl-style/issues/21, partly concerning https://github.com/openmaptiles/openmaptiles/issues/500

zoom 4:
- use NE data with `min_zoom <= 5` (not `7`)
- this way regions in France (and other countries) are not displayed
- countries with smaller admin. units are then only: USA, Canada, Brasil, Russia, China, India, Australia, Indonesia, South Africa
- updates of NE are not really organized, as described in https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-1-states-provinces/, so there can be more problematic countries (not only France)

zoom 5:
- use the same content like in zoom 6 - only OSM data, no NE input
- this way there won't be too much data in z5, it will be more up to date, there will not be too big difference between reduced z4 and z5 and no dismissing data when zooming to the z6

Comments are welcome!


pathmapper:
Currently [OMT uses NE 4.0](https://github.com/openmaptiles/openmaptiles/blob/master/docker-compose.yml#L16) and (for me) it's not clear if a switch to NE 4.1 would affect what should be achieved with this PR.

What about upgrading to NE 4.1 in the scope of this PR to have the data from the latest NE release available and be prepared for future NE releases?

I think this would mean not to consider NE `scalerank` anymore and use `min_zoom` and `min_label` instead (see https://github.com/openmaptiles/openmaptiles/issues/586 and https://github.com/openmaptiles/openmaptiles/issues/365#issuecomment-442979739) 

> countries with smaller admin. units are then only: USA, Canada, Brasil, Russia, China, India, Australia, Indonesia, South Africa

Pakistan is missing from that list.

eva-j:
I did some more tests and found out that we can avoid using scalerank quite easily. It was used in zoom 1, where condition `WHERE scalerank <= 2` did no difference, and in updated zoom 4, where could be used `WHERE min_zoom <= 5` only. I've just committed this edits.

Anyway, I am quite confused by the NE version used. I am using docker-compose, but it seems I have NE 4.1 imported (even in the fresh database). I checked 
https://github.com/openmaptiles/import-natural-earth/blob/master/Dockerfile
and
https://github.com/openmaptiles/import-natural-earth/blob/v1.4/Dockerfile
plus other versions of import-natural-earth and it seems like it is getting data from the same link:
http://naciscdn.org/naturalearth/packages/natural_earth_vector.sqlite.zip - it seems to be v4.1

Please correct me if I am wrong, but aren't we using v4.1 already (despite of the documentation and ticket https://github.com/openmaptiles/openmaptiles/issues/365#issuecomment-442823991)?

As z4 and z5 are reduced by this PR and z5 now include the same boundaries as z6 - z8, I plan to add more data into those zoom levels (https://github.com/openmaptiles/openmaptiles/issues/615).

pathmapper:
> we can avoid using scalerank quite easily

`scalerank` is also used in other layers, e.g. https://github.com/openmaptiles/openmaptiles/blob/master/layers/place/update_country_point.sql

> aren't we using v4.1 already (despite of the documentation and ticket #365 (comment))?

The v1.4 docker image was created on November 16, 2017:

https://hub.docker.com/r/openmaptiles/import-natural-earth/tags

Bildschirmfoto von »2019-05-06 19-55-23«

The NE release 4.1. was on May 23, 2018 and the 4.0 release on October 27, 2017:
https://github.com/nvkelso/natural-earth-vector/releases

So despite of the same link (which has no version) the docker v1.4 docker image should contain NE 4.0.

> I plan to add more data into those zoom levels

Sounds good!

PR #645, opened 2019-07-16
Remove the area condition on highway polygon.

francois2metz:
For some reason highway multipolygon where not rendered. We use them a lot in Montrouge, France as we map sidewalk with multipolygon.

An example of multipolygon rendered with openmaptiles after this patch:

Screenshot_2019-07-16 OSMontrouge

- the relation: https://www.openstreetmap.org/relation/9792499
- the live map: https://osmontrouge.fr/@48.818765,2.320661,18.36

### DIFF
--- a/layers/transportation/layer.sql
+++ b/layers/transportation/layer.sql
@@ -366,7 +366,7 @@ indoor INT, bicycle TEXT, foot TEXT, horse TEXT, mtb_scale TEXT, surface TEXT) A
         WHERE zoom_level >= 13
             AND (
                   man_made IN ('bridge', 'pier')
-                  OR (is_area AND COALESCE(layer, 0) >= 0)
+                  OR (ST_GeometryType(geometry) IN ('ST_Polygon','ST_MultiPolygon') AND COALESCE(layer, 0) >= 0)
             )
     ) AS zoom_levels
     WHERE geometry && bbox


eva-j:
Hi @francois2metz, is this still relevant problem? If I look at the maps generated using OMT now, I see all the paths: https://www.maptiler.com/maps/#streets//vector/17.59/2.320521/48.818777
Do you have more location examples to we can test it?

francois2metz:
@eva-j Hi, the paths are here but not the sidewalks. The paths were added to ease the routing. 

If you look around the maptiler's preview, you can see some sidewalks, but not all of them. For instance this sidewalk (https://www.openstreetmap.org/relation/9453420) is a multipolygon with a hole inside is not rendered (https://www.maptiler.com/maps/#streets//vector/17.74/2.320664/48.817895).

layers/transportation/layer.sql
```diff
@@ -343,7 +343,7 @@ indoor INT, surface TEXT) AS $$
         WHERE zoom_level >= 13
             AND (
                   man_made IN ('bridge', 'pier')
-                  OR (is_area AND COALESCE(layer, 0) >= 0)
```
eva-j: `OR (ST_GeometryType(geometry) IN ('ST_Polygon','ST_MultiPolygon') AND COALESCE(layer, 0) >= 0)`

To avoid printing polygons in the places where are closed line features (see Note above), I suggest to take only polygon and multipolygon features. Testing area (Ile-de-France) passed this way, I will be happy if you edit this PR and test it too.

francois2metz: I cannot see your note, but I'll try what you propose.

eva-j: I meant this one https://github.com/openmaptiles/openmaptiles/pull/645/files#diff-508d01732fb3c79a172704135f511452R325-R328

francois2metz: I updated the query and the tiles on osmontrouge.fr. I haven't seen any differences.

eva-j: OK, thank you!

eva-j:
I am sorry, we need to revert this change because of cases like this one https://github.com/openmaptiles/openmaptiles/issues/724. I was wrong in my testing, checking if geometry is 'ST_Polygon' or 'ST_MultiPolygon' doesn't eliminate closed lines imported as polygons.

francois2metz:
Thanks. I'll take a look to find another solution.

PR #646, opened 2019-07-28
Add a subclass column from the mapping_value to clean the landcover layer

francois2metz:
I wanted to add some new mapping in the landcover layer, but I figured out that the code was difficult to update. 
I added a subclass column of type `mapping_value` to make things easier to maintain.


eva-j:
Thanks for sorting it in subclasses first, @francois2metz, it works good.

francois2metz:
Thanks for merging this @eva-j  !

PR #655, opened 2019-08-12
Remove duplicate POI class match on chocolate and confectionery

frodrigo:
I am afraid we must remove some chocolate tiles.

`chocolate` and `confectionery` POI subclass are duplicated:
* https://github.com/openmaptiles/openmaptiles/blob/master/layers/poi/class.sql#L33
* https://github.com/openmaptiles/openmaptiles/blob/master/layers/poi/class.sql#L48


nyurik:
good catch, makes sense. In general I think we should be moving from the generic "shop" into a more specific category.

PR #658, opened 2019-08-26
close #657: limit mountain_peak's 'ele' tag to 10000

### LINKED ISSUE #657: Prevent OSM vandalism on natural=peak:ele
Like in the https://github.com/openmaptiles/openmaptiles/issues/315 I came across a mistake or vandalism:
https://www.openstreetmap.org/node/6716725790/history
that caused:
```
Error: Postgis Plugin: ERROR:  value "11111111111118" is out of range for type integer
```
because that peak had elevation tag equal to 11111111111118.

Like in the issue with buildings we can prevent this by filtering out peaks with elevation >= 10000.

golubev:
An attempt to avoid "out of range" crashes, fixes https://github.com/openmaptiles/openmaptiles/issues/657.

In short - this peak:
https://www.openstreetmap.org/node/6716725790/history
was causing:
```
Error: Postgis Plugin: ERROR:  value "11111111111118" is out of range for type integer
```

So I tried to do a limitation like in the #316.


nyurik:
@golubev thanks, but why not just fix the regex?  e.g. change `AND ele ~ E'^-?\\d+'` to `AND ele ~ E'^-?\\d{1,4}$'` would limit the number to `-9999..9999` -- would that suffice?

golubev:
Hi @nyurik,

I think that the solution you proposed is really more elegant than mine.

It will also fix an issue with peaks that have non-integer `ele` values. There are some objects in the OSM having non-integer `ele` tag value:
https://taginfo.openstreetmap.org/keys/ele#values

The old regex `E'^-?\\d+'` captures only integer values and perhaps is losing some peaks.

I agree that if the regex test yields `false` on empty `ele` then we can remove the redundant `is not null` check.

nyurik:
@golubev ok, please update your patch. Also, I think we should allow up to 5 digits -- highest mountains are under 30,000 ft (just in case someone uses ft) -- should still solve the parsing problem. Thanks!

nyurik:
Actually never mind, the wiki page specifies it to always be in meters, so 4 digits should be enough.  Please update and i will merge. thanks!

golubev:
Hi @nyurik,

I've updated the branch.

Also, I have left the `is not null` condition. I decided to do so because the regular expression operator yields `NULL` if the first argument is `NULL`. I thought that it is more correct and strict to have proper types in the condition. Otherwise, the whole logical condition would yield `NULL` instead of `True` or `False`.

nyurik:
@golubev WRT null vs false -- not a big deal really :).  You do realize that by following your logic we should rewrite half of SQL? E.g. `geometry && bbox` would also need to be changed to `geometry is not null AND geometry && bbox` ? :D

golubev:
Okay, @nyurik, I've got it! )

Thanks! :hand:

PR #662, opened 2019-08-31
Support natural=dune aside of beach and sand

frodrigo:
Add `natural=dune` in the sand landcover class.

There is not lot of `natural=dune` but there are major objects of the landscape, and can be simply mapped the `sand` class.

https://taginfo.openstreetmap.org/tags/natural=dune#overview


eva-j:
Thanks for adding this!

PR #663, opened 2019-08-31
Map fell, heath, scrub and tundra like grassland, as 'grass'

### LINKED ISSUE #660: Scrub is not included in landcover
Scrub is currently not included in landcover.sql

I've made a fork that includes scrub to landcover/landcover.sql, landcover/mapping.yaml and landcover/landcover.yaml

frodrigo:
Add more natural value mapped to `grass` class.

Some empty areas are on maps because some natural kinds are not mapped. Just map this small vegetation fields to grass.

Fix #660

### DIFF
--- a/layers/landcover/landcover.sql
+++ b/layers/landcover/landcover.sql
@@ -15,7 +15,7 @@ CREATE OR REPLACE FUNCTION landcover_class(landuse VARCHAR, "natural" VARCHAR, l
         WHEN "natural" IN ('glacier', 'ice_shelf') THEN 'ice'
         WHEN "natural"='wood' OR landuse IN ('forest') THEN 'wood'
         WHEN "natural" IN ('bare_rock', 'scree') THEN 'rock'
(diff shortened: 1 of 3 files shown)


eva-j:
Thanks for this PR.
Could you please add information about size of tiles before and after this change for some test countries?
Can you list a few places for which it make sense to include?

frodrigo:
Alpage
https://www.openstreetmap.org/#map=13/42.9292/-0.1263
image
image

Beach
https://www.openstreetmap.org/#map=15/43.5875/-1.4826
image
image

frodrigo:
I test on Aquitaine, France (PBF 214Mb):

* Original mbtiles size: 614.400
* With this patch : 622.592

+1.3%

sqldiff give no output. Not sure if it is normal or not.

Do you want test on larger area? but I am not sure to be able to do it.

eva-j:
Thanks a lot for testing, it looks good, @frodrigo

PR #664, opened 2019-09-01
Upgrade import-sql from 0.8 to 1.0

typebrook:
- In `import-sql:0.9`, it allow allows no Wikidata in psql, solve #548 
- Add description about Wikidata in `README`

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,16 @@ docker-compose run import-lakelines
 docker-compose run import-osmborder
 ```
 
+**[Optional]**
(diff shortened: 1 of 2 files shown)


nyurik:
import_sql repo has been merged to [tools](https://github.com/openmaptiles/openmaptiles-tools#importing-into-postgres) - we should migrate to use that instead.

typebrook:
@nyurik 
Thanks for this great work! I'll take a look at it later.

PR #667, opened 2019-09-03
Remove pixel_scale

petrsloup:
There is `pixel_scale: 256` in the metadata, but I don't think there's any practical reason for this.

- Vector tiles are usually rasterized into 512x512 (so why 256?)
- I couldn't find any library using this value for anything

### DIFF
--- a/openmaptiles.yaml
+++ b/openmaptiles.yaml
@@ -25,7 +25,6 @@ tileset:
   bounds: [-180.0,-85.0511,180.0,85.0511]
   maxzoom: 14
   minzoom: 0
-  pixel_scale: 256
   languages:
     - ar # Arabic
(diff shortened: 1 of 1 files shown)


sfkeller:
AFAIK vector tiles are usually rasterized to 256.
I BTW found this which perhaps gives some more background info:
https://support.maptiler.com/i413-does-512-pixel-scale-result-in-more-request-and-high-data-consumption-also-how-to-fetch-512-pixel-tiles

MartinMikita:
This value is now required for generating `tm2source`, in [openmaptiles-tools](https://github.com/openmaptiles/openmaptiles-tools/blob/master/openmaptiles/tmsource.py#L26)

PR #675, opened 2019-09-28
Fix err in transportation update_gbr_route_members

nyurik:
The query was producing 3 columns - `(member, ref, network)`,
whereas osm_id is a required field (otherwise it throws null constraint).

I suspect this was never triggered before because no data has ever matched
the filter, thus always producing null results, and hence being skipped.

### DIFF
--- a/layers/transportation_name/update_route_member.sql
+++ b/layers/transportation_name/update_route_member.sql
@@ -8,7 +8,7 @@ BEGIN
   select st_buffer(geometry, 10000) into gbr_geom from ne_10m_admin_0_countries where iso_a2 = 'GB';
   delete from osm_route_member where network IN('omt-gb-motorway', 'omt-gb-trunk');
 
-  insert into osm_route_member (member, ref, network)
+  insert into osm_route_member (osm_id, ref, network)
     (
       SELECT hw.osm_id, substring(hw.ref from E'^[AM][0-9AM()]+'), 'omt-gb-motorway'
       from osm_highway_linestring hw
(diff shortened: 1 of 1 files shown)


MartinMikita:
Have you tested this change? Because the column `member` is used in the select query, which will be NULL after your change.
You need to duplicate the value of `member` into `osm_id` column.
The actually used version of `imposm` creates this table with a NULL allowed column `member`.

amatissart:
Maybe we could simply remove the `osm_id` column from `route_member` table ? Is it used anywhere ?

```diff
diff --git a/layers/transportation/mapping.yaml b/layers/transportation/mapping.yaml
index 8819329..12a4af2 100644
--- a/layers/transportation/mapping.yaml
+++ b/layers/transportation/mapping.yaml
@@ -344,8 +350,6 @@ tables:
   route_member:
     type: relation_member
     columns:
-    - name: osm_id
-      type: id
     - name: member
       type: member_id
     - name: role
```

nyurik:
@MartinMikita TBH I do not know if the query now does what it was originally intending to do, but yes, it does execute without errors after my change.  I do not know why switching to the newest imposm has triggered this error  - I'm guessing now it creates all tables with non-null osm_id. Should I add the proposed change by @amatissart ?

klokan:
Unique OSM ID is important and used by some users - for example, Carto's Mobile SDK (Nutiteq) rendering engine has behaved better on features with preserved OSM IDs as unique identifiers.

Picking and query into external OSM-related APIs for requesting additional info about a clicked feature of given OSM ID is another use-case. So removing OSM ID is in general not a good idea.
 
Not sure to what extent this PR (or suggested removal of osm-id at this place in code) breaks the compatibility and these functions - and I have no time to dig deeper into this right now...

nyurik:
@klokan thx for the feedback.  The issue here is that the newer imposm version marks `osm_id` column as `non null`, thus the old query no longer works - because it used to produce `null` for the `osm_id` column.  If `osm_id` and `member` columns should have identical data (kinda strange but ok), we could change the query to do just that.

nyurik:
Could we merge this change?  It simply replaces `null` with `0` in the `osm_id` column.

eva-j:
Thanks for solving this, nyurik!

Issue #321, opened 2017-08-22
Proposed Road & Under Contruction Road


joaokho:
There is any tag for proposed road and under contraction road on openmaptiles styles

jirik:
Neither proposed roads nor under construction roads are handled rin the schema right now. Related tags:
- https://wiki.openstreetmap.org/wiki/Key:construction
- https://wiki.openstreetmap.org/wiki/Tag:highway%3Dproposed
- https://wiki.openstreetmap.org/wiki/Key:proposed

MartinMikita:
see [description of lifecycle](https://wiki.openstreetmap.org/wiki/Key:highway?uselang=en-US#Lifecycle).

An example: [The bridge `Honk-Kong-Zhuhai-Macao`](https://www.openstreetmap.org/way/572611768/history) has been tagged incorrectly as completed, but fixed later, now it is not visible in OSM vector tiles.

It could be probably as a separate layer, with other "proposed" or "under construction" items in OSM.
To be researched.

Pull requests are welcome, if you are able to compare the sizes and counts, @joaokho

eva-j:
https://github.com/openmaptiles/openmaptiles/pull/677 merged, thanks.

golubev:
Thanks, @eva-j

golubev:
Can someone update docs and schemas, please? That PR has brought some changes.

Also, it would be great if someone could check whether the docs had been updated correctly in the https://github.com/openmaptiles/openmaptiles/pull/677.

/cc @eva-j @nyurik

zvukoper:
Thank you, @golubev !!!

nyurik:
@golubev `make generate-devdoc` is your friend :) Just make sure you run `docker pull openmaptiles/openmaptiles-tools` to get the latest docker image.  Pushed updated images as https://github.com/openmaptiles/openmaptiles/commit/73c7290893255ae49c122a158083944f5964d658 

We should somehow make validation of the docs as part of the make file (should be fairly straightforward)

golubev:
Thanks, @nyurik! See https://github.com/openmaptiles/openmaptiles/pull/700.

PR #677, opened 2019-10-08
Add roads under construction

golubev:
This PR adds roads under construction to the `transportation` and `transportation_name` layers. It is related to the issue https://github.com/openmaptiles/openmaptiles/issues/321.

I decided not to add a new property, e.g. boolean `is_under_construction`, because it would probably break backward compatibility in most of the styles. In such a case, all roads under construction would appear as normal ones. If we add a road with `class=motorway` and `is_under_construction=1` it would pass filters like `["==", "class", "motorway"]` and it would be painted as an ordinary motorway.

In my solution I created new classes `xxx_construction` for each original class of roads depending on the `construction` tag:
- `motorway_construction`
- `trunk_construction`
- `primary_construction`
- `secondary_construction`
- `tertiary_construction`
- `minor_construction`
- `service_construction`
- `track_construction`
- `raceway_construction`
- `path_construction`

There are quite many cases when roads with `highway=construction` tags are not given the `construction` tags to determine their classes. Such roads fall under the `minor_construction` class as a "generic road under construction".

The main motivation of this PR was to remove the emptiness and sudden breaks of roads in places where roads are being repaired.

This is how it looks before the roads under construction were added:

Screenshot from 2019-10-08 11-24-13

And after (some pieces of the road lack the `construction` tag, so they are painted with white color as "a generic road under construction"):

Screenshot from 2019-10-08 11-25-10

Some more examples (we draw roads under construction as dashed lines):

Screenshot from 2019-10-08 11-32-30
Screenshot from 2019-10-08 11-31-55

### DIFF
--- a/layers/transportation/class.sql
+++ b/layers/transportation/class.sql
@@ -17,9 +17,8 @@ CREATE OR REPLACE FUNCTION highway_class(highway TEXT, public_transport TEXT) RE
         WHEN highway IN ('secondary', 'secondary_link') THEN 'secondary'
         WHEN highway IN ('tertiary', 'tertiary_link') THEN 'tertiary'
         WHEN highway IN ('unclassified', 'residential', 'living_street', 'road') THEN 'minor'
-        WHEN highway IN ('service', 'track') THEN highway
         WHEN highway IN ('pedestrian', 'path', 'footway', 'cycleway', 'steps', 'bridleway', 'corridor') OR public_transport IN ('platform') THEN 'path'
-        WHEN highway = 'raceway' THEN 'raceway'
+        WHEN highway IN ('service', 'track', 'raceway') THEN highway
         ELSE NULL
     END;
 $$ LANGUAGE SQL IMMUTABLE;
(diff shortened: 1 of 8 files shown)


nnhubbard:
This seems quite complicated. In my branch I have just added the `construction` as a mapping type for `highway`. It works great and so far I have not ran into any issues styling. Does anyone see why this simple implementation would be a problem?

golubev:
Hi @nnhubbard,

Thanks for your comment!

Initially, I've done the same patch as you proposed - https://github.com/openmaptiles/openmaptiles/pull/677/commits/45a50a5bd09d4a09d9d811581e9fafd42d9c74b4.

But in our case, that did run into some styling issues. In our style, major roads (`motorway`, `trunk`, `primary`) are yellow and minor roads are white. I thought that it would be nice to be able to differentiate constructed roads by classes.

Let's suppose we have some part of a major road (e.g. `motorway` or `trunk`) being repaired. At lower zooms when only major roads are visible one will like a repaired piece of a major road to be visible too, have the same color and width as the original major road. Now let's imagine that in a nearby town there are some `tertiary` or `minor` roads being constructed or repaired. Those roads are most likely to have higher `minzoom`, less width and a different color than the major roads have. It seems to me that it is almost impossible to find a common style for those two examples of road classes. I think that if one will try to use a common style for all roads under construction then he will run into some visible roads issues. E.g. a repaired part of the road would disappear while zooming, or it will differ by width or color, etc.

That was the main motivation for me to differentiate roads under construction by classes.

But I can also suppose that destroyed or repaired roads are common issues for only the post-USSR countries. And other countries don't have to deal much with repaired roads while mapping :joy:

eva-j:
I've just tested difference in tile size on Switzerland (262.1 MB --> 262.3 MB) and Lithuania - it is minimal.
Thanks you for this PR!

PR #679, opened 2019-10-11
add brunnel info for water

GuillaumeGomez:
The "Canal Saint-Martin" in Paris is a tunnel for a long time so we need to extend the brunnel feature to water as well. This PR does it.

cc @amatissart


GuillaumeGomez:
ping @eva-j

eva-j:
This is really rare case...
https://taginfo.openstreetmap.org/keys/?key=tunnel#combinations
https://taginfo.openstreetmap.org/keys/?key=bridge#combinations

But it looks ok, tile size is basically not changing.
Thanks for adding it, @GuillaumeGomez

GuillaumeGomez:
Thanks!

PR #685, opened 2019-10-21
Generate gettile.sql in a function form

eyusupov:
This allows running versions of postserve that use it (like https://github.com/openmaptiles/postserve/pull/15)


eyusupov:
SQL file with gettile function has to be imported after `tileset.yaml`, so I renamed it to `tileset_gettile.sql`

nyurik:
@eyusupov `postserve` should not rely on pre-generated files. Instead, it should generate what it needs dynamically, directly from the layer configuration.  Of course there could be an override, but I don't think we should be automating it tbh.

eyusupov:
You're right - we actually don't need that in https://github.com/openmaptiles/openmaptiles-tools/pull/62 anymore

PR #688, opened 2019-10-23
Avoid long repeated text in poi mapping file

nyurik:
Manual merge of the original #611

The fields and mapping sections for poi_point and poi_polygon are the same, and must be the same for the SQL to work properly.

Using definitions avoids repetition and the need to make each change in these sections twice, as was already done for the subsections of mapping.


nyurik:
sorry, messed up with history, redoing...

PR #691, opened 2019-10-25
layers/poi: new weight function to consider page views

remi-dupre:
The rank of a POI is computed through a weight function now primarily based on page views of its corresponding Wikipedia page, if any, otherwise it will use a similar ranking to the previous method.


remi-dupre:
This PR is a mistake: it should be commited to the QwantResearch/openmaptiles fork as it relies on external tables.

PR #693, opened 2019-10-26
Manage field mapping in SQL declaratively

nyurik:
Simplify some of the OSM->OMT field value mappings using declarative syntax.

This approach is not for all cases, but in many it removes
the need of storing the same field in both the .yaml and .sql files.

Another PR will support more complex AND/OR cases

### DIFF
--- a/layers/aerodrome_label/aerodrome_label.yaml
+++ b/layers/aerodrome_label/aerodrome_label.yaml
@@ -15,12 +15,23 @@ layer:
           [`aerodrome`](http://wiki.openstreetmap.org/wiki/Proposed_features/Aerodrome)
           and `aerodrome:type` tags.
       values:
-      - international
-      - public
-      - regional
-      - military
-      - private
-      - other
+        international:
+          aerodrome: 'international'
+          aerodrome_type: 'international'
+        public:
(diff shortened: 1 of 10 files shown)


kukulich:
@nyurik Is somewhere a fork with the upgrade to newer openmaptiles-tools ?

nyurik:
> @nyurik Is somewhere a fork with the upgrade to newer openmaptiles-tools ?

@kukulich I have been mostly working in https://github.com/sophox/openmaptiles , but there is no current openmaptiles that is using the latest tools -- mostly because I use the tools commands directly in most of my projects, rather than using docker-compose or quickstart. The current tools can use the latest openmaptiles layers without any changes.

frodrigo:
This PR can now be merged. It would be nice!

nyurik:
yes, OMT is now using latest tools, so this can be merged.

nyurik:
thx for reviewing!  BTW, not a biggie, but usually i think it is better for the original contributor to merge the code (when they have the rights for that) -- because it keeps original contributor more responsible for the change (per our [code review process](https://github.com/openmaptiles/openmaptiles/blob/master/CONTRIBUTING.md#code-review-process))

PR #700, opened 2019-11-04
update mapping diagram for the `transportation_name`

golubev:
Updated transportation name mapping diagram after the https://github.com/openmaptiles/openmaptiles/pull/677.


eva-j:
Thanks for this update, @golubev

PR #706, opened 2019-11-15
Disputed borders - adding disputed_view attribute

eva-j:
This is an option how to handle disputed borders, if we want to have a chance to show them differently in style, according to the point of view.

After discussion in team, we decided it would be fine to have an attribute containing an identifier of disputed area and a code of the country that likes to see the boundary line. This way we can make then different combinations in style and show the map from the view of many different countries, not just a few preset.

Here is an example of Crimea. The new attribute `disputed_view` gets value of `crimea_ru` or `crimea_ua`. You can see the changes in the style in red for now:

Current map, showing disputed boundaries in dashed line:
Screenshot from 2019-11-15 11-31-06

We can then display the borders according to the Russian view:
Screenshot from 2019-11-15 11-30-05

or the Ukrainian:
Screenshot from 2019-11-15 11-29-15

As there are no columns in input data, that we could easily use to find and name the conflicting borders, in this version I just used `osm_id` for OSM input and `ogc_fid` for NE input.
This means quite a lot of work on keeping the problematic parts up to date, I will be happy if anybody finds a better way to fill `disputed_view` column.
From import-osmborder we get this set of attributes: https://github.com/pnorman/osmborder#output. Then we use 3 different tables from NaturalEarth import - `ne_110m_admin_0_boundary_lines_land`, `ne_50m_admin_0_boundary_lines_land` and `ne_10m_admin_0_boundary_lines_land` to get `admin_level=2` which applies disputed borders, so I've set a function for each of them.

As ID's of lines are not changing really often, I hope this approach can work at least for the most problematic areas (I am going to add Kashmir and a few other Asian places into this PR). I would like to keep it up to date correspondingly to every new release of import-osmborder and import-natural-earth.

Please, write your comments and suggestions.

### DIFF
--- a/layers/boundary/boundary.sql
+++ b/layers/boundary/boundary.sql
@@ -1,11 +1,48 @@
+CREATE OR REPLACE FUNCTION osm_disp_view(disputed bool, osm_id bigint) RETURNS TEXT AS $$
+    SELECT CASE
+        WHEN disputed THEN CASE
+            WHEN osm_id IN (130072455, 216249910, 216249912, 130072435, 310894006, 310894007, 130072449, 320568448, 130207737, 130072456, 130207714, 330695989, 330696042, 330696028, 330696000, 330695990, 130207721, 171255708, 30555413) THEN 'crimea_ru'
+            WHEN osm_id IN (238797482) THEN 'crimea_ua'
+            ELSE NULL
+        END
+        ELSE NULL
+    END;
+$$ LANGUAGE SQL IMMUTABLE;
+
+
+CREATE OR REPLACE FUNCTION ne10_disp_view(ogc_fid int) RETURNS TEXT AS $$
+    SELECT CASE
+        WHEN ogc_fid IN (462) THEN 'crimea_ru'
+        ELSE NULL
+    END;
+$$ LANGUAGE SQL IMMUTABLE;
+
+
+CREATE OR REPLACE FUNCTION ne50_disp_view(ogc_fid int) RETURNS TEXT AS $$
+    SELECT CASE
+        WHEN ogc_fid IN (361) THEN 'crimea_ru'
+        ELSE NULL
+    END;
+$$ LANGUAGE SQL IMMUTABLE;
+
+
+CREATE OR REPLACE FUNCTION ne110_disp_view(ogc_fid int) RETURNS TEXT AS $$
+    SELECT CASE
+        WHEN ogc_fid IN (186) THEN 'crimea_ru'
+        ELSE NULL
+    END;
+$$ LANGUAGE SQL IMMUTABLE;
+
+
 
 
 -- etldoc: ne_110m_admin_0_boundary_lines_land  -> boundary_z0
 
 CREATE OR REPLACE VIEW boundary_z0 AS (
-    SELECT geometry,
+    SELECT 0 AS osm_id, geometry,
         2 AS admin_level,
         (CASE WHEN featurecla LIKE 'Disputed%' THEN true ELSE false END) AS disputed,
+        ne110_disp_view(ogc_fid) AS disputed_view,
         false AS maritime
     FROM ne_110m_admin_0_boundary_lines_land
 );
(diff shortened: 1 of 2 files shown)


daliborjanak:
Not ideal to do that with IDs but there is no better way now. We can refactor this behaviour after OSM community will say how to store that clearly
https://wiki.openstreetmap.org/wiki/Proposed_features/Mapping_disputed_boundaries

nyurik:
This method is almost guaranteed to break within months if not weeks, as any of those ways are edited. I think a slightly more stable approach might be some sort of a geofencing -- e.g. set up a materialized view that contains a list of way IDs that are part of all disputed borders. That mat view could be generated by a hard-coded query, something like  `find all ways that have admin_level=... and are located within bbox/geoshape ...`, where the bbox will be hard-coded over the entire line of the conflict?  We could even hard-code this query as part of the tile generation itself, if it proves to be as fast as a table join, and assuming it is comparable to the `in (id1, id2, ...)` approach proposed by this PR.

eva-j:
Thanks a lot for your comment, @nyurik!

As I went through the data more in detail yesterday, I am getting to the point that we should completely rewrite boundary layer processing in OMT.

In Kashmir area, there is this problematic line https://www.openstreetmap.org/way/340291900
In OSM it shows `admin_level=3`, in OMT `admin_level=7` so for us it is visible starting in z11 (and it is what India likes to see as country boundary).

Screenshot from 2019-12-03 12-19-47

Screenshot from 2019-12-03 12-20-19

This is caused by the settings of osmborder project - it takes `admin_level` from relation, not the way https://github.com/pnorman/osmborder#admin_level

Then there are some boundaries missing or not detected as disputed. E.g. missing boundaries in Arunachal Pradesh area for `disputed_view=arunachalpradesh_cn`

Screenshot from 2019-12-03 20-15-36

So even if we detect disputed area with bbox/geoshape as you suggested, there will always be gaps because of not correctly specified admin levels or missing geometry from import...

Then the question is, if we want to include this PR into OMT v3.11 to have at least something for now, or better wait till we find a better solution.
I would like to completely rewrite import-osmborder or test generating boundary layer directly from OSM data. Then we could maybe use relations (https://www.openstreetmap.org/relation/4611980, https://www.openstreetmap.org/relation/4610574, https://www.openstreetmap.org/relation/4611981).
Anyway it would be good to get more into discussion with OSM community and decide together about mapping this problematic (as https://wiki.openstreetmap.org/wiki/Proposed_features/Mapping_disputed_boundaries was rejected).

nyurik:
@eva-j thanks for such a detailed reply!

WRT `admin_level=3` vs `admin_level=7` -- I think we can handle that with bbox method, as long as the data is actually in the table, e.g. admin_level >= 7.

I agree that if the data is not there, we won't find anything, but in that case listing ways by their IDs won't work either.  If on the other hand the ID listing works, we could try to find a `where` condition to capture them with a geo filter (i.e. bbox).  In other words my thought goes like this:  if you can filter lines based on their IDs, you can also filter those same lines on some other method like admin_level and geometry, but that filtering could be anything. As long as the filter is a bit more generic than a list of IDs.

Judging by my past experiences with OSM community, making any kind of change and arriving at an agreement there will be a looooong process.  I wonder if @nvkelso might want to collaborate on this here?

nvkelso:
Hi @nyurik, @eva-j, and @daliborjanak. Great idea to add "point of view" or "worldview" to the boundaries layer! I was focused on the same earlier this year tagging in OSM and Natural Earth and posted two diary entrees about this:

- [March 2019] **Disputed boundary tagging sprint:** https://www.openstreetmap.org/user/nvk/diary/47890
- [August 2019] **disputed boundaries update:**  https://www.openstreetmap.org/user/nvk/diary/390496

The relations are available in OSM with that tagging for the cases you mention, it's better to rely on relations and transformed tags into OMTv3 like this as they are more resistant to OSM feature churn. 

There's a pre-release version of Natural Earth v5 available in the Tilezen PRs and assets that match that. It'd be great to collaborate on this moving forward :) 

_Caveat, my second kid is due any day now so my response times may be delayed._

eva-j:
@nvkelso, could you please share a link to the pre-release version of Natural Earth v5?

frodrigo:
I also try an other approach, using "bit fields",  each bit is on if match the country point of view

https://www.openstreetmap.org/user/frodrigo/diary/390502
https://github.com/frodrigo/osmborder/commits/disputed_claim
https://gist.github.com/frodrigo/afbb8e89da7dde0861c1485036d5b893#file-index-html-L157-L169

eva-j:
Thank you all for your comments, I am closing this PR. We need to use OSM relations to detect country view points on disputed boundaries.

nvkelso:
Natural Earth v5.0.0-pre5 has the disputed properties added. The file is in this comment, and linked PR has the schema and ETL:

https://github.com/tilezen/vector-datasource/issues/1840#issuecomment-497583012

PR #707, opened 2019-11-19
Use unified tools version for all images

nyurik:
* TOOLS_VERSION file controls which tools version will be used
* Remove `import-sql` docker-compose image
* Use `openmaptiles-tools` instead of `import-sql` and `postserve` docker images
   (`import-sql` script and `postserve` server have been migrated to tools)
* Add MVT test to quickstart


MartinMikita:
The problem with tests:
```
[Dec  3 19:57:02] IO error: /cache/coords/LOCK: No such file or directory
The command "./quickstart.sh" exited with 1.
```

I think, the problem is with DC_OPTS for travis - volume belongs to root.
Could we set default DC_OPTS for travis to be as before? I am not familiar with travis CI.

Also, docker-compose.yml is missing `import-sql` service, which is used by quickstart, tested by Travis CI.

nyurik:
@MartinMikita sorry, still working on this. Root ownership is a massive pain to work with (all files created by docker will be owned by root, and hard to use/delete afterwards).  Note that this patch relies on a noop patch #726 , which should be merged first.

nyurik:
P.S. `import-sql` was deleted on purpose -- all of its code is now part of the tools itself.

MartinMikita:
Thanks for changes and your refactor of OMT tools @nyurik 

We are thinking, wouldn't be good to only change `import-osm` to use `tools` docker image, instead of removing?

```
import-sql:
    image: "openmaptiles/openmaptiles-tools:${TOOLS_VERSION}"
    command: import-sql
```

This would let to use `docker-compose run import-sql` without change.

What do you think?

nyurik:
@MartinMikita sure, that could also work for now, but I think in the future we should remove duplicate docker compose targets if they are identical - unless they need to run at the same time.

PR #709, opened 2019-11-19
Fix missing description for the new brunnel field

nyurik:
Fixes metadata for query added in https://github.com/openmaptiles/openmaptiles/pull/679

cc: @GuillaumeGomez


GuillaumeGomez:
Nice! :+1:

PR #718, opened 2019-11-29
Add leisure=golf_course, map it to grass

frodrigo:
`leisure=golf_course` are not present on tiles data, giving large empty area with only a golf POI.

Map golf polygon to grass.


eva-j:
Do you please have a comparison of tile size in some testing areas?

frodrigo:
I not planned to do one. As it only add golfs. But I can make one.

frodrigo:
Original mbtiles size of Aquitaine, France: 258256896

With golfs: 258527232 -> +0,01%

Issue #703, opened 2019-11-07
Check continent and ocean name against OSM vandalism


eva-j:
Check continent and ocean names - compare them with NaturalEarth data.
Same logic as is already done for countries.

eva-j:
This ticket was started mostly because of names like "Sudameriko", "Norda Atlantiko" in OMT maps:

![Screenshot from 2019-12-02 11-07-30](https://user-images.githubusercontent.com/8367628/69950733-30b58680-14f4-11ea-9800-bb693cf11aae.png)
![Screenshot from 2019-12-02 11-07-58](https://user-images.githubusercontent.com/8367628/69950731-30b58680-14f4-11ea-95ba-f194680f38ef.png)

There was Esperanto as default language for some time.

Natural Earth data is used to compare and set ranks for countries, but it is not so easy to use for name comparison of continents and waters.
PR https://github.com/openmaptiles/openmaptiles/pull/719 solves the problem - it is setting English where Esperanto is as default.

Anyway the changes in OSM data were reverted and seems to be more controlled in future:
https://www.openstreetmap.org/changeset/77742097 ("revert Esperanto names, see https://www.openstreetmap.org/user_blocks/3294")
This PR now loses its relevance.

PR #719, opened 2019-12-02
Avoid default Esperanto in names

eva-j:
Concerning names of continents (`place` layer) and marine points (`water_name` layer).
Commented in https://github.com/openmaptiles/openmaptiles/issues/703.


eva-j:
We can use this later when it's needed, for now the problem is solved on data side.

PR #720, opened 2019-12-06
Add garages

golubev:
Garages areas are quite common for post-soviet countries - https://wiki.openstreetmap.org/wiki/Tag:landuse=garages.

They can cover quite big areas. They are very similar to industrial zones (for a map user) and can be styled in the same way.

See e.g.
https://www.openstreetmap.org/way/297028765


golubev:
Here is a screenshot from the overpass-turbo. It shows how much `way({{bbox}})[landuse=garages];` there are for Kiev -

Screenshot from 2019-12-06 19-53-32

eva-j:
Thanks for this PR. Could you please make a comparison of tile size in some testing areas?

golubev:
Okay, I'll try.

nyurik:
@golubev you may want to import all your data into postgres, and run the `test-perf` utility for all tests just for zoom 14, and record the output (using the `-r` option).  Afterwards, modify your layer definition to actually use the new garage data, and re-run the `test-perf` with the same parameters, except instead of `-r` use `-c` to compare.

eva-j:
I've just tested it on Ukraine and Switzerland. There will be a lot of countries with no difference, so I suggest to approve this PR.

Area | import-osm | generate-vectortiles | size
--- | --- | --- | ---
Ukraine | 2m53.154s --> 2m57.470s | 2m46.556s --> 2m46.79s | 58.0 MB --> 58.7 MB
Switzerland | 1m32.230s --> 1m33.502s | 14.91s --> 14.856s | 10.2 MB --> 10.1 MB

golubev:
Many thanks, @eva-j and @nyurik!

Sorry, I was drowned in my tasks at work and put this off every day (

PR #721, opened 2019-12-06
Filter building on polygons

frodrigo:
The table osm_building_relation, osm_building_associatedstreet and osm_building_street are built from relation members and not filtered on polygons.

Thus tables contain polygons but also point and line, not desirable for buildings layer.
And can not be filtered at imposm import step.

This table are filled with more or less lines and points, depending if addresses are set using relation on building polygons or using dedicated address nodes.

This check can be done once at indexing time using a where clause on the index.
The SQL change on index have no measurable impact on creation time on my tests. But the index is now smallest.

Original, without filter on building geometry. Test made on Aquitaine, France, with z14 only

* Time to generate MBTiles : 54m44.543s
* MBtiles size : 179 871 744

Modified, with filter on building geometry

* Time to generate MBTiles : 54m27.697s
* MBtiles size : 178 827 264 (-0.58%)

Since Aquitaine is in France. And in France AssociatedStreert relations are used more than in other part of the world the gain is probably better.

### DIFF
--- a/layers/building/building.sql
+++ b/layers/building/building.sql
@@ -13,10 +13,10 @@ END;
 $$ STRICT
 LANGUAGE plpgsql IMMUTABLE;
 
-CREATE INDEX IF NOT EXISTS osm_building_relation_building_idx ON osm_building_relation(building);
+CREATE INDEX IF NOT EXISTS osm_building_relation_building_idx ON osm_building_relation(building) WHERE ST_GeometryType(geometry) = 'ST_Polygon';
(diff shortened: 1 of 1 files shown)


eva-j:
Thanks a lot for this PR, it looks good!
Just one thought, aren't there also MultiPolygons that will be excluded this way?

frodrigo:
As I understand multipolygons are present only in table `osm_building_polygon`, with polygons. This table is not filtered.

PR #725, opened 2019-12-10
[WIP] Cluster buildings

frodrigo:
The buildings layer on z14 is so far the heaviest one. It is particularly the case on dense urban area where the median time to generate tiles is about 5 000 ms.

Using materialized view of building reduce this median time to 3 200 ms. But still impracticable for real time tiles production.

Clustering the buildings using a grid highly reduce the number of objects to be fetched from database in dense area. The clustering is made using the other attributes to keep them.

Nevertheless clustering buildings requite a precomputation step and a storage to a table with index on geometry.

On Aquitaine, France

| Step | Original | With Clustering |
|-|-:|-:|
| Import time (import-osm + import-sql) | 4m42 | 10m27 |
| mbtiles export z0-z14 | 57m16 | 51m18 |
| Total | 61m58 | 61m45 |
| 1 day Database update avg | 44s | 58s |

On Europe

| Step | Original | With Clustering |
|-|-:|-:|
| Import time (import-osm + import-sql) | 11h25 | 16h19 |
| mbtiles export of z14  | na | na |
| Total | na | na |
| 1 day Database update avg | 3h00 | 4h30 |

The clustering take more time at preprocessing step, but less at tiles generation.

But the big change is on median time to produce z14 tiles in dense urban area, is now only at 225 ms.

At counter part, we lost the `osm_id` on clustering.

The table of buildings cluster take time to build and can be improved using differential update. Updating only cluster of changed buildings.

This is a work in progress.

### DIFF
--- a/layers/building/building.sql
+++ b/layers/building/building.sql
@@ -59,8 +59,8 @@ FROM osm_buildings_standalone
 DROP TABLE IF EXISTS osm_all_buildings_mat CASCADE;
 CREATE TABLE osm_all_buildings_mat AS (
     SELECT
-        --max(osm_id) AS osm_id,
-        ST_Collect(geometry) AS geometry,
+        array_agg(osm_id ORDER BY osm_id)    AS osm_ids,
+        ST_Collect(geometry ORDER BY osm_id) AS geometry,
         height,
         min_height,
         levels,
@@ -140,8 +140,8 @@ FROM (
                                   material,
                                   colour,
                                   hide_3d
-         FROM (SELECT NULL::bigint             AS osm_id,
-                      (ST_Dump(geometry)).geom AS geometry,
+         FROM (SELECT unnest(osm_ids) AS osm_id,
+                      (ST_Dump(geometry)).geom AS geometry, -- Set Returning Functions are dumped in parallel
                       height,
                       min_height,
                       levels,
@@ -240,7 +240,8 @@ BEGIN
 
     -- Seach for clusters of changed buildings
     CREATE TEMP TABLE impacted_clusters AS
-    SELECT osm_all_buildings_mat.geometry,
+    SELECT osm_ids,
+           osm_all_buildings_mat.geometry,
            height,
            min_height,
            levels,
@@ -281,7 +282,8 @@ BEGIN
 
     -- Unpack impacted clusters
     CREATE TEMP VIEW unclustered_buildings AS
-    SELECT (ST_Dump(geometry)).geom AS geometry,
+    SELECT unnest(osm_ids) AS osm_id,
+           (ST_Dump(geometry)).geom AS geometry, -- Set Returning Functions are dumped in parallel
            height,
            min_height,
            levels,
@@ -293,7 +295,8 @@ BEGIN
 
     -- Discart old buildings from clusters
     CREATE TEMP VIEW untouched_buildings AS
-    SELECT unclustered_buildings.geometry,
+    SELECT osm_id,
+           unclustered_buildings.geometry,
            height,
            min_height,
            levels,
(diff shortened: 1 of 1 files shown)


nyurik:
needs rebase

daliborjanak:
It can break existing applications if we lose osm_id. I would prefer to not do that at least on Z14.

frodrigo:
This PR is only on z14.

klokan:
Is this a pure performance optimization - described at 
https://makina-corpus.com/blog/metier/2020/optimisation-de-la-production-de-tuiles-vectorielles-de-fond-de-carte-openstreetmap-pour-un-service-a-la-demande
Correct? @frodrigo 

Is the pain on upper zoom levels (12, 13?) which is being solved here?

We wanted to have something similar for upper zoom levels (13 and perhaps 12) - with "blocks of buildings" in town centers merged into a polygon - this means modifying the geometry - not only doing a single MULTI-POLYGON from multiple polygons within a 100x100 meters areas.
Wouldn't that solve the problem in a more proper / elegant way?

Losing the osm_id on deepest zoom levels on buildings is not really a great advantage of this approach.

klokan:
Let's discuss on the next community meeting.

frodrigo:
Yes. English version here : https://medium.com/@frederic.rodrigo/optimization-of-openstreetmap-background-vector-tile-production-for-an-on-request-service-makina-5f6dba98a232

On Z13 only large building are already on tiles. This approach does not match this existing zoom level.
https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/mapping.yaml#L3

My test does not have issue on z12 or z13 like on z14 for dense urban area.
```
  2.0 tiles/s at z12,  20 tiles in 0:00:09.7
  7.0 tiles/s at z13,  80 tiles in 0:00:11.3
  0.9 tiles/s at z14, 252 tiles in 0:04:40.0
```

Merge geometries cost more than just collect. My approach was to lost less attribute as possible on buildings. The grouping of building is done with all other attributes. So it is able to keep the height for 3D render eg. And un-grouping the cluster keep the same result as before, except for osm id.
The main idea on this PR is to be able to server tiles on demand.

Merging polygons for z12 or z13 will have major impact on perf.

Maybe Z12 and z13 should have a different strategy than z14. Keeping all detail on z14 (with my approach or not).

frodrigo:
Nevertheless, my other PRs are more elementaries and should be instructed to simply the buildings layer.

nyurik:
If this is still a `WIP`, please click `convert to draft` link in the `Reviewers` section on the right top.

frodrigo:
@daliborjanak @klokan About the osm_id I make a prof of concept and I am able to do it.
Nevertheless, currently the feature_id of the building layer is not set to the osm_id.

We can switch it to osm_id, but it is an other issue.

klokan:
Cool! Well done - if you can now preserve the feature_ids @frodrigo !

Let's check the implication on the updating process and tile generating performance & size of the produced tiles on this (pretty cool) PR.

frodrigo:
@klokan but do you not prefer first we switch the feature_id of building layer to osm_id alone in a PR?

klokan:
Yes, it could make a great separate PR. Where we also need to verify the implication on the size of the produced tiles. It should run with CI thanks to the effort from @TomPohys and @nyurik IMHO

frodrigo:
The proof of concept for keeping osm_id is on this extra branch : https://github.com/frodrigo/openmaptiles/commits/building_cluster_with_osm_id
It not deals with update yet.

Waiting for update on switching to osm_id for feature_id or not before moving on. (#827)

nyurik:
@frodrigo I merged this branch with master, resolving a few conflicts, let me know if changes are ok

frodrigo:
I updated with the support of the osm_id of the building.
While building are clustered, I group osm_id in array.
On read, I extract ST_Collection and array on parallel.

TomPohys:
Thank @frodrigo for updating this PR. I need to look closer into this PR.

Issue #495, opened 2018-09-03
Customize country boundaries


Rajavelu:
Hi Guys,
              
I am using OpenMapTiles for our vector tiles server.

**Default Boundary of India :**
 
<img width="685" alt="screen shot 2018-09-06 at 12 46 19 pm" src="https://user-images.githubusercontent.com/19775829/45150445-844dfd00-b1e9-11e8-8886-2a95119e4f7a.png">

**Proper boundary we need :** 

<img width="573" alt="screen shot 2018-09-06 at 12 46 03 pm" src="https://user-images.githubusercontent.com/19775829/45150503-99c32700-b1e9-11e8-88b5-495192136736.png">

- By default, it shows incorrect India boundary. So i have created a layer in style.json using geojson data of all countries boundary data(with updated India boundary).

- But it's taking too much time to render boundary line. I've also tried trimming the geojson data(1.6mb), but still its rendering very slowly.

- I also tried to render country boundary using OpenMapTiles boundary data for all countries except India, then add layer using geojson for India boundary only.
But i can't disable rendering the default boundary for India using OpenMapTiles, as they are not providing country code or name with boundary data.

Any alternate options ?

_**Is there an option to change the boundary data of select countries in the mbtiles itself ?**_

Thanks.

eva-j:
We are trying to find a solution for these cases - https://github.com/openmaptiles/openmaptiles/pull/706.

somm-doe:
I have a similar issue; currently you cannot use the `name=` attribute to select the borders manually.

PR #727, opened 2019-12-11
Disputed borders - adding relations with claimed_by

eva-j:
Adding attributes `disputed_name` and `claimed_by` to boundary lines that are part of OSM boundary relation. This enables to show borders from the point of view of one country or filtering according to the name of disputed area.

Example of displaying Chinese point of view:
Screenshot from 2019-12-11 11-57-45

In zoom levels 1 - 4, relations from OSM source were just added to the lines from NE. If you want to show a general map without any specific point of view, you can dismiss them in style using `claimed_by !has`. We will update this after NE v5 is officially released.
Screenshot from 2019-12-11 11-48-00

There is still a lot of work that needs to be done on data side - in OSM.

Previous discussion in https://github.com/openmaptiles/openmaptiles/pull/706.
Important for https://github.com/openmaptiles/openmaptiles/issues/495.


nyurik:
This PR introduced a SQL bug -- https://github.com/openmaptiles/openmaptiles/issues/742

PR #728, opened 2019-12-11
Parallel capability to layer functions

jsanz:
This PR allows queries to be parallelized on recent versions of Postgres. The `PARALLEL SAFE` modifier has been added to the layer functions and a PLPGSQL function to convert strings into number has been replaced.

`PARALLEL SAFE` is a modifier for `CREATE FUNCTION` available since Postgres 9.6, so this change does not break current OpenMapTiles supported database version. More details about this topic [here](https://www.postgresql.org/docs/current/parallel-safety.html) and at the reference documentation for [`CREATE FUNCTION`](https://www.postgresql.org/docs/current/sql-createfunction.html).

### Testing procedure

The procedure to test this was:

* Imported `spain.pbf` in a clean environment
* Dumped the OpenMapTiles database from the Postgres Docker image
* Created a clean Postgres 12 database using the default Docker image
* Installed `postgis` 3 from the default Debian package and `osml10n` 2.5.8 from the repository (`make`, etc.)
* Restored the dump
* Lowered the postgres planner parameters for triggering parallel plans:
```sql
set parallel_setup_cost = 5;
set parallel_tuple_cost = 0.005;
```
* Manually added the `PARALLEL SAFE` modifier to each function involved in layer queries (not on updates or inserting functions).
* For each layer, run a testing query to confirm parallel workers were created, something like this:
```sql
explain analyze 
select * from layer_aerodrome_label(tilebbox(8,128,95),10,null)
union all
select * from layer_aerodrome_label(tilebbox(8,128,97),10,null);
```
* After all the layers were processed and confirmed to start parallel executions, a more complete example was run. This example just retrieves the geometries for all the layers from the same tile but without using any MVT related function.

<details><summary>Testing query</summary>

```sql
-- Using the function layer_landuse
explain analyze 
select geometry from layer_water(tilebbox(14,8020,6178),14)
union all
select geometry from layer_waterway(tilebbox(14,8020,6178),14)
union all
select geometry from layer_landcover(tilebbox(14,8020,6178),14)
union all
select geometry from layer_landuse(tilebbox(14,8020,6178),14)
union all
select geometry from layer_mountain_peak(tilebbox(14,8020,6178),14)
union all
select geometry from layer_park(tilebbox(14,8020,6178),14)
union all
select geometry from layer_boundary(tilebbox(14,8020,6178),14)
union all
select geometry from layer_aeroway(tilebbox(14,8020,6178),14)
union all
select geometry from layer_transportation(tilebbox(14,8020,6178),14)
union all
select geometry from layer_building(tilebbox(14,8020,6178),14)
union all
select geometry from layer_water_name(tilebbox(14,8020,6178),14)
union all
select geometry from layer_transportation_name(tilebbox(14,8020,6178),14)
union all
select geometry from layer_place(tilebbox(14,8020,6178),14)
union all
select geometry from layer_housenumber(tilebbox(14,8020,6178),14)
union all
select geometry from layer_poi(tilebbox(14,8020,6178),14)
union all
select geometry from layer_aerodrome_label(tilebbox(14,8020,6178),14);
```
</details>

You can inspect the execution plan and results on [this page](https://explain.dalibo.com/plan/3z). Also [attaching](https://github.com/openmaptiles/openmaptiles/files/3951822/explain-tile-simple.tar.gz) the query and JSON output for future reference. The website gives a ton of details, but you may want to search for nodes mentioning `workers` or `parallel` like in this area referring to `osm_border` or `osm_aeroway_linestring` entities

image

### Next steps

Since the execution plan is not showing a parallel append at the top level, meaning it's not running each layer individually, I want to continue experimenting with parameters and queries to see if it's possible to even parallelize more the request.

I will post my finding here, even no change in the code should happen.

cc. @nyurik


layers/building/building.sql
```diff
@@ -1,17 +1,17 @@
...
+CREATE OR REPLACE FUNCTION as_numeric(text)
+RETURNS NUMERIC AS                                                                
+$$                                
+SELECT CAST(
+        (REGEXP_MATCH(
+            TRIM($1),
+            '^(([-+]?[0-9]+(\.[0-9]+)?(e[-+]?[0-9]+)?)|([-+]?\.[0-9]+)(e[-+]?[0-9]+)?)$')
+        )[1] AS NUMERIC);
+$$
+LANGUAGE SQL
```
nyurik: * We might want to add `COST`, e.g. `COST 5` or `COST 20` -- this function seems like it is quicker than the usual functions (cost=100), but slightly slower than a C++ ones (cost=1)?  Might need to tune it a bit.
* TRIM + Regex is not very efficient - might as well include optional spaces inside the regex, i.e. `'^ *((...) *$'`.  Please add those to the test (btw, at some point we should orchestrate all these tests too, but probably not as part of this PR)
* for some reason I was unable to get `REGEXP_MATCH` to run in the `postgis` docker, need to double check why its broken.

nyurik: update - REGEXP_MATCH was not available in 9.6, not REGEXP_MATCHES.

jsanz: Yep I saw the CI failing, I will change that into using `REGEXP_MATCHES` :+1:

jsanz: addressed at https://github.com/openmaptiles/openmaptiles/pull/728/commits/980190105a2f28d2ec6d5cb946f5a2343b2335c2

jsanz:
https://github.com/openmaptiles/openmaptiles/pull/728/commits/980190105a2f28d2ec6d5cb946f5a2343b2335c2 changes `as_numeric` to support 9.6 to 12. I've removed the `trim` call  and added `COST`. Also changed the function to return `-1` as the original function for any non valid value.

layers/building/building.sql
```diff
@@ -1,17 +1,24 @@
...
-DECLARE test NUMERIC;
-BEGIN
-     test = $1::NUMERIC;
-     RETURN test;
-EXCEPTION WHEN others THEN
-     RETURN -1;
-END;
-$$ STRICT
-LANGUAGE plpgsql IMMUTABLE;
+CREATE OR REPLACE FUNCTION as_numeric(text)
```
nyurik: the above comments should be moved down to the index creation.

layers/building/building.sql
```diff
@@ -1,17 +1,24 @@
...
+$$
+SELECT CASE 
+            WHEN TEST IS NULL THEN -1
+            WHEN TEST[1] IN ('','.') THEN -1
+            ELSE CAST(TEST[1] AS NUMERIC)
+        END AS RESULT
+FROM (
+    SELECT ARRAY_AGG(i) AS TEST
+    FROM (
+        SELECT (REGEXP_MATCHES($1,'^[\ ]*?([-+]?[0-9]*\.?[0-9]*?(e[-+]?[0-9]+)?)[\ ]*?$','i'))[1] i
```
nyurik: we really need a bunch of unit tests for this, i'm afraid someone will start modifying it to fix some new case, and break the existing usages. Actually maybe we should move this to tools repo as part of the utilities functions? This seems too generic and possibly useful for other cases?

jsanz: this is definitely generic, I understand we could add this to a new `.sql` file inside the `openmaptiles-tools/sql` folder and [this script](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-sql) would import it automatically, right?

regarding testing, what would be your approach? creating a new type of test execution at the `Makefile` or coding them in Python as a new testcase?

layers/building/building.sql
jsanz: Do you mean about [the first two lines](https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/building.sql?#L1-L2) starting with `--etldoc:`? I understood they referred to the whole file and they were already at that position before this PR :thinking:

layers/building/building.sql
nyurik: @jsanz yes, adding a new file to `/sql` in tools would solve everything nicely, thanks!
WRT testing, I guess the simplest way would be to use Makefile (even though I do prefer Python in a general case, Makefile is currently the standard to run simple commands... maybe eventually we will change that). So the Makefile target would do these steps as part of the `build-tests` target (probably a separate sub-target)
- create a new postgis docker container (using the same docker image as we use for everything else)
- run `psql` to import all files in the `/sql` dir.
- run `psql` once for each "testing" sql file, i.e. in `test-sql/*.sql`, and save each output into `/build/`
- place expected results into the `testdata/expected/` dir.
- makefile already has the needed code to compare the results in the `/build` with the `testdata/expected`, so you will just reuse that.

layers/building/building.sql
nyurik: hmm.. never mind, ignore, i think we may want to clean it up separately.

nyurik:
@jsanz i implemented unit testing framework for the tools as https://github.com/openmaptiles/openmaptiles-tools/pull/155 . Please add `as_numeric` as a new file there, together with the relevant unit tests. Once added and we can see that the exact copy works, I will +1 this PR.  Note that we could also wait for the new version of the tools to be released, so that we can upgrade this repo to the latest tools version and remove `as_numeric` completely. Also, I think we should rename `as_numeric` to `omt_as_numeric` to avoid name collisions. Thanks!

frodrigo:
Can we at least merge the `PARALLEL SAFE` part? I can create an other PR to do that.

nyurik:
@frodrigo I would prefer to remove `as_numeric` changes from this PR, and merge it afterwards (this way all of our discussion can be kept in one place. The `as_numeric` could then be merged into the tools repo (as `omt_as_numeric`), and once we publish a new tools release, we can just delete it here and use the new version.

jsanz:
@nyurik @frodrigo I've reverted the `as_numeric` function to its original state and updated the description of the PR accordingly.

layers/aerodrome_label/layer.sql
```diff
@@ -36,5 +36,6 @@ $$
     round(substring(ele from E'^(-?\\d+)(\\D|$)')::int*3.2808399)::int AS ele_ft
   FROM osm_aerodrome_label_point
   WHERE geometry && bbox AND zoom_level >= 10;
-
-$$ LANGUAGE SQL IMMUTABLE;
+$$ 
+LANGUAGE SQL
```
frodrigo:
```suggestion
$$
LANGUAGE SQL
```

Please can you remove trailing spaces.

nyurik: done

PR #736, opened 2019-12-13
Fixed docker-compose commands.

MartinMikita:
This PR should fix running simply `docker-compose` commands, that are addressed in README.md.


MartinMikita:
@nyurik I think, that we should use only this `TOOLS_VERSION` in `.env` file, instead of parsing and using `TOOLS_VERSION`.

This variable could be overridden by the environment variable, which keeps `quickstart.sh` and `Makefile`.

I could add another commit to clean it, if you agree.

nyurik:
Sure, I'm ok with any approach as long as the version number is only stored in one place. TBH I think at some point we should move to k8s instead of compose, but clearly this is a long way off

PR #738, opened 2019-12-13
Remove unused import-sql image

nyurik:
import-sql image is not used anywhere in OMT, deleting.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -67,17 +67,6 @@ services:
       - ./data:/import
       - ./build:/mapping
       - cache:/cache
-  import-sql:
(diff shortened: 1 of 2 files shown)


eva-j:
This will need a change also in readme (documentation) I guess.

nyurik:
@eva-j readme does not reference `import-sql` image. It only mentions `import-sql` command.

eva-j:
Oh, I see... Instead of 
`docker-compose run import-sql`
will be used
`docker-compose run openmaptiles-tools import-sql`
but in Readme there is just `make import-sql` which stays the same.

PR #740, opened 2019-12-14
Create boundary materialized views

nyurik:
Create all the needed materialized views as part of OMT, not part of the tools.

This is similar #739


MartinMikita:
Thanks for this PR! It would be definitely better for future or custom changes.

Would be possible to format these SQLs, so they are more readable? Probably to have selected columns in `SELECT` directive on a separate line? Or do you have a better suggestion to improve the readability of it?

nyurik:
@MartinMikita i broke it up into multiple lines, thx, good call. Note that the first two lines of each MAT VIEW will be deleted very soon, once we drop them in the tools repo, making it even more readable.

Issue #742, opened 2019-12-14
Disputed borders PR added warnings to SQL


nyurik:
Beginning with TravisCI [build 774](https://travis-ci.org/openmaptiles/openmaptiles/builds/624268832?utm_medium=notification&utm_source=github_status) triggered by the merge of https://github.com/openmaptiles/openmaptiles/commit/d82ede81ce6adba82f5efa7cd637831142f8d0b3 by PR https://github.com/openmaptiles/openmaptiles/pull/727 a warning began to appearing in the logs:

```
psql:/sql/tileset.sql:1115: NOTICE:  Processing layer boundary
DO
Time: 0.755 ms
CREATE FUNCTION
Time: 1.593 ms
psql:/sql/tileset.sql:1135: WARNING:  column "disputed_name" has type "unknown"
DETAIL:  Proceeding with relation creation anyway.
psql:/sql/tileset.sql:1135: WARNING:  column "claimed_by" has type "unknown"
DETAIL:  Proceeding with relation creation anyway.
CREATE VIEW
```

Once this bug is resolved, merge #743 to avoid it in the future.

MartinMikita:
Note: using PostgreSQL 11.4 DB with OMT-tools 3.1.0 does not generate this warning, and the lines in tileset.sql are moved (probably expected)?

eva-j:
Thanks for detecting this, @nyurik!

PR #743, opened 2019-12-14
Detect sql warnings and abort quickstart

nyurik:
If quickstart generates an SQL `WARNING` while running
the import-sql script, abort.

This should be merged after https://github.com/openmaptiles/openmaptiles/issues/742 is resolved. TravisCI correctly detected that there is currently a problem:
image


eva-j:
Thanks for this correction, @nyurik

PR #752, opened 2019-12-22
remove useless sql selection

ldgeo:
Hi team, 
here is a small modification which removes unnecessary sql selection on landuses and parks layers.

- in landuse.sql: `scalerank` is not used nor returned in function `layer_landuse`
- in park.sql: `scalerank` is only used for the point subselection


nyurik:
@ldgeo please join us tomorrow at 10am US Eastern Time for the OpenMapTiles community sync up using https://meet.google.com/wsi-ebva-dhm . I also sent you an invite directly to your email, but if that's wrong, contact me at YuriAstrakhan@gmail.com and I will add you to the proper invite in the future. Thanks!

TomPohys:
Just for Info: If there is already a view landuse_z6 - z14 in PostgreSQL database and was created with scalerank column, after this PR the views must be deleted manually. Otherwise, it will raise an error - cannot drop columns from view

PR #755, opened 2020-01-02
Micro SQL improvement of building layer

frodrigo:
Micro change of the SQL to filter osm_building_relation early in the query. I test and assert it give the same result.


layers/building/building.sql
```diff
@@ -82,10 +82,10 @@ CREATE OR REPLACE VIEW osm_all_buildings AS (
                   COALESCE(nullif(as_numeric(obp.min_level),-1),nullif(as_numeric(obp.buildingmin_level),-1)) as min_level,
                   nullif(obp.material, '') AS material,
                   nullif(obp.colour, '') AS colour,
-                  CASE WHEN obr.role='outline' THEN TRUE ELSE FALSE END as hide_3d
+                  obr.role IS NOT NULL as hide_3d
          FROM
          osm_building_polygon obp
-           LEFT JOIN osm_building_relation obr ON (obr.member = obp.osm_id)
+           LEFT JOIN osm_building_relation obr ON obr.member = obp.osm_id AND obr.role = 'outline'
```
nyurik: I would prefer to add non-join-related conditions inside the `WHERE` clause for clarity (unless there is some major performance difference?).  So for this just move the `obr.role = 'outline'` to `WHERE`

frodrigo: No. Since we are in a `LEFT JOIN` it is not the same. Here `obr.role = 'outline'` only apply to joined record.

nyurik: ah my mistake, i was overthinking it. Thx for explaining.

frodrigo:
This PR is part of set to simply this. Thank you any way for your time.

frodrigo:
Rebased.

frodrigo:
Rebased, again.

daliborjanak:
Thanks for the improvement! But I'm missing the clear basic information about this PR (and we need to improve the templates for pull requests):

Why are you doing this? Ticket? - this is just code cleaning? 
How it will change the schema? - no change
How ti will affect the size of the output file? - no increase?
How it will affect rendering time? - no increase?

frodrigo:
> Why are you doing this? Ticket? - this is just code cleaning?

Code cleaning found in implementing #725

> How it will change the schema? - no change

No change

> How ti will affect the size of the output file? - no increase?

No change

> How it will affect rendering time? - no increase?

Theoretically better. Filter data early. But maybe the same perf.

PR #756, opened 2020-01-02
Remove table building_multipolygon

frodrigo:
Buildings from ways and multipolygons are loaded in table `osm_building_polygon`. But a table for `osm_building_multipolygon` is also loaded, the content is not used except to ensure an `osm_id` is from a multipolygon. To check if the object is from a multipolygon we have only to check if `osm_id` is negative. It is the counter part of https://github.com/openmaptiles/openmaptiles/blob/e0c8ece375348a17463aba24ec8b65780941b7cd/layers/building/building.sql#L89

I checked the objects are the same after this change.


frodrigo:
Rebased.

PR #758, opened 2020-01-03
Filter osm_building_polygon on polygon

frodrigo:
The table `osm_building_polygon` contents way and multipolygon. This table aim to contain polygon but also contains some lines and point as unsolved geometry from multipolygon.

Filtering on geometry type avoid to output this lines and points in building layer.

### DIFF
--- a/layers/building/building.sql
+++ b/layers/building/building.sql
@@ -72,7 +72,8 @@ CREATE OR REPLACE VIEW osm_all_buildings AS (
                   FALSE as hide_3d
          FROM
          osm_building_polygon obp
-         WHERE osm_id < 0
+         -- OSM mulipolygons once imported can give unique postgis polygons with holes, or multi parts polygons
+         WHERE osm_id < 0 AND ST_GeometryType(geometry) IN ('ST_Polygon', 'ST_MultiPolygon')
 
          UNION ALL
          -- etldoc: osm_building_polygon -> layer_building:z14_
@@ -88,7 +89,8 @@ CREATE OR REPLACE VIEW osm_all_buildings AS (
          FROM
          osm_building_polygon obp
            LEFT JOIN osm_building_relation obr ON (obr.member = obp.osm_id)
-         WHERE obp.osm_id >= 0
+         -- Only check for ST_Polygon as we exclude buildings from relations keeping only positive ids
+         WHERE obp.osm_id >= 0 AND ST_GeometryType(obp.geometry) = 'ST_Polygon'
 );
 
 CREATE OR REPLACE FUNCTION layer_building(bbox geometry, zoom_level int)


layers/building/building.sql
```diff
@@ -71,7 +71,7 @@ CREATE OR REPLACE VIEW osm_all_buildings AS (
                   nullif(colour, '') AS colour,
                   FALSE as hide_3d
          FROM
-         osm_building_polygon obp WHERE EXISTS (SELECT 1 FROM osm_building_multipolygon obm WHERE obp.osm_id = obm.osm_id)
+         osm_building_polygon obp WHERE EXISTS (SELECT 1 FROM osm_building_multipolygon obm WHERE obp.osm_id = obm.osm_id) AND ST_GeometryType(geometry) = 'ST_Polygon'
```
nyurik: This might be a small nit -- I usually try to put the "faster" execution path before the slower ones in a `AND` condition -- i.e.  `WHERE  <something fast> AND <something slow>`.  The subquery is probably slower to execute than a simple geometry type check. Otherwise you rely on PostgreSQL optimizer to catch these, and TBH i don't know if it will handle that.

frodrigo: You are right, especially before postgres 12.

frodrigo: Even on 11, plan dit not change when changing the order.

frodrigo:
Yes I miss the `ST_MultiPolygon`. Rebeased and commit amended.

Only `ST_MultiPolygon` on the first query, since it is the only one dealing with OSM relations.

nyurik:
@frodrigo could you add a comment to both changes to explain why you do it, esp why the second case only has `ST_Polygon` and not `ST_MultiPolygon` -- i'm sure someone else might look at it later and think it might be a bug.

frodrigo:
Done to comply and have a stable state. But note I rework and simplify this code in my others PR.

layers/building/building.sql
nyurik: Does the execution plan show constant filters like this?  Very possible that PostgreSQL optimizes for that, at least in the later versions (we still use 9.6 for some prod btw). For me its just a matter of practice, but hopefully most modern systems catch these things when they can (in theory you cannot have compiler solve it in an arbitrary complex code). Regardless, its probably too minor of a point to worry about.

nyurik:
needs rebase

frodrigo:
Rebased

layers/building/building.sql
frodrigo: Since rebase, there is no more question here.

nyurik:
@TomPohys or @MartinMikita or someone else from MapTiler, please review

TomPohys:
There is no performance and size difference. Tested on the Czech Republic and finished successfully.

PR #770, opened 2020-01-23
Use import-borders script instead of docker image

nyurik:
Switch to the new import-borders script, works on the same data
as used for importing, and makes more streamlined build process.

This can be merged at any time, without waiting for OMT tools
release. Later we can switch from omt-tools-latest to omt-tools.


nyurik:
oops, dup of #750

PR #771, opened 2020-01-24
Narrow osm_building_relation_building_idx with condition building = ''

frodrigo:
Following the query https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/building.sql#L33
also add `building = ''` to the index.


frodrigo:
Thank you !

PR #785, opened 2020-02-19
Upgrade to tools v5 - rm import-osm, new downloader...

nyurik:
Update to tools v5. See https://github.com/openmaptiles/openmaptiles-tools/releases/tag/v5.0.0 for the list of all changes. Other OMT-repo specific changes:

* removes `import-osm` docker usage, replacing it with `openmaptiles-tools`
* quickstart builds faster because it uses postgres with preloaded water, natural earth, and lake centerlines tables.

### Makefile targets
* `tools-dev` will open a shell in a docker to experiment and debug (instead of `import-sql-dev` and `import-osm-dev`)
* separated `start-maputnik` from `start-postserve`
* renamed `clean-docker` into `db-destroy` to make it more explicit
* cleaner `db-start`, `db-stop`, `db-destroy` targets
* `db-start-preloaded` is the same as `db-start`, except that it uses `postgis-preloaded` -- an image with preloaded water, natural-earth, and lake centerline data
* `db-start` will not recreate the container if it already exists -- this way if it was started as preloaded, it will not be rebuilt.
* better output messages

### Quickstart
* uses `postgis-preloaded` image by default to make quickstart quicker.  To start with a clean db, pass 2 parameters to quickstart, e.g. `./quickstart.sh albania empty`


QUICKSTART.md
```diff
@@ -426,8 +426,7 @@ Hints for developers:
   make psql-analyze                    # PostgreSQL: ANALYZE
   make generate-qareports              # generate reports [./build/qareports]
   make generate-devdoc                 # generate devdoc  [./build/devdoc]
-  make import-sql-dev                  # start import-sql  /bin/bash terminal
-  make import-osm-dev                  # start import-osm  /bin/bash terminal (imposm3)
+  make tools-dev                  # start import-sql  /bin/bash terminal
```
frodrigo:
```suggestion
  make tools-dev                       # start import-sql  /bin/bash terminal
```

TomPohys:
Thank you @nyurik for this PR. @MartinMikita and I need to check and verify the new tool v5 and OMT v5. Testing on the whole planet with diff and updates could take a week. We are working on it.

docker-compose.yml
```diff
@@ -1,86 +1,49 @@
...
-      - postgres_conn
-    volumes:
-      - ./data:/import
-      - ./build:/mapping
-      - ./cache:/cache
-  update-osm:
-    image: "openmaptiles/import-osm:${TOOLS_VERSION}"
-    env_file: .env
-    environment:
-      DIFF_MODE: ${DIFF_MODE}
```
MartinMikita: This config needs to be in OMT-tools docker service, because it is not possible to change DIFF_MODE from shell, without changing `.env` file
See [this document](https://vsupalov.com/override-docker-compose-dot-env/)

Makefile
```diff
@@ -53,7 +55,7 @@ help:
 
 .PHONY: init-dirs
 init-dirs:
-	mkdir -p build && mkdir -p data && mkdir -p cache
+	mkdir -p build/sql && mkdir -p data && mkdir -p cache
```
MartinMikita: Wouldn't be better to split these 3 directories into separate lines?
If we are already cleaning and improving Makefile...

Makefile
```diff
@@ -193,73 +232,60 @@ generate-qareports:
 .PHONY: generate-devdoc
 generate-devdoc: init-dirs
 	mkdir -p ./build/devdoc && \
-	docker-compose run $(DC_OPTS) openmaptiles-tools-latest sh -c \
+	docker-compose run $(DC_OPTS) -e TOOLS_VERSION=latest openmaptiles-tools sh -c \
```
MartinMikita: Why is this using `latest` version of tools?

docker-compose.yml
nyurik: thanks, makes sense, fixed

Makefile
nyurik: I used latest because tools had newer code to generate documentation images (removed dups, etc). This was relevant for older version of OMT. Now it doesn't matter any more, removed.

MartinMikita:
Hi @nyurik,
We found another issue with our testing process for the whole planet.

The step `make import-wikidata` fails, with the message "asyncpg.exceptions.UndefinedTableError: relation "wd_names" does not exist", related to the code in OMT tools - `await conn.execute(f'TRUNCATE TABLE {storage_table}')`.

From further testing and reviewing the whole code, we found the difference between README (with older approach) and `quickstart.sh`: **Import wikidata** step has been used **before import SQL**.

Also, all older `quickstart.sh` were executing firstly `import-wikidata`, then `import-sql`. Moreover, `import-wikidata` step was optional, therefore there is a single SQL query, which creates the table, if it does not exists.

The previous `import-wikidata` tool (step) was creating this table at the beginning - https://github.com/openmaptiles/openmaptiles-tools/blob/v3.1.0/docker/import-wikidata/wikidata/etl.py#L10..L21

This should stay the same with the new `import-wikidata` tool.

MartinMikita:
Another related question is - is there any logical reason to have `import-wikidata` AFTER `import-sql` command? I was in the impression, that there are (were) some SQL functions/queries, that could modify imported wikidata, therefore the `import-sql` step was the last one in the processing queue.

nyurik:
@MartinMikita thx for the review. We can do the import-wikidata step before -- that's totally fine.  I will update `import-wikidata` tool to automatically create the `wd_names` if missing.

MartinMikita:
Hi @nyurik,

Why we have switched to CleanNumeric, instead of using `omt_as_numeric` function, defined and verified in [tools](https://github.com/openmaptiles/openmaptiles-tools/pull/157)?

TomPohys:
https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/building.sql#L4
`nullif(as_numeric(height),-1` return null if `as_numeric` was `-1`, 

https://github.com/openmaptiles/openmaptiles-tools/blob/master/sql/numeric.sql
`omt_as_numeric` returns `-1` if `variable` is not number. The same as `as_numeric`

https://github.com/openmaptiles/postgis-vt-util/blob/master/src/CleanNumeric.sql
`cleanNumeric` returns `null`.

Using cleanNumeric make more sense to get rid of `nullif`. `omt_as_numeric` using `cleanNumeric` but instead of `null` returns `-1`

nyurik:
@MartinMikita as @TomPohys says - we remove redundant  "badvalue => null => -1 => null" conversion.  as numeric actually doesn't do anything else beyond that.  We probably should add a few tests to vt utils, or perhaps maybe even migrate away from vt-utils entirely (once we are fully switched to using AsMVT), and just have OMT-defined functions.

sfkeller:
Given so much efforts you put into this nice function and and given the value it, I'd suggest following:

* Agree first on a set of tests.
* Putting the suggested code on Stackoverflow https://stackoverflow.com/questions/tagged/postgresql 
* Evaluate and benchmark the solutions proposed on Stackoverflow.
* Publish finally the best one on PG Wiki https://wiki.postgresql.org/wiki/Category:Snippets

nyurik:
@sfkeller are you referring to the [postgis-vt-util](https://github.com/openmaptiles/postgis-vt-util) library? Or the few extra functions defined in [openmaptiles-tools](https://github.com/openmaptiles/openmaptiles-tools/tree/v5.0-dev/sql) ?

sfkeller:
I'm referrinf mainly to CleanInt and to CleanNumeric https://github.com/openmaptiles/postgis-vt-util/blob/master/src/CleanNumeric.sql .

BTW: I would rename them to `to_int_clean()` and `to_numeric_clean()` - in order to be closer to the well-known built-in `to_number()` and to stick to the lower-case SQL style (guide), see e.g. https://www.sqlstyle.guide/#stored-procedures .

nyurik:
@sfkeller these functions have existed for many years as part of [mapbox/postgis-vt-util](https://github.com/mapbox/postgis-vt-util).  At this point the hope is to introduce as few disruptions as possible to migrate to v5, but in the later PRs.  I created an related issue in https://github.com/openmaptiles/openmaptiles/issues/794

sfkeller:
@nyurik et al. : I've asked about a review of ClearInt on PG IRC channel, and Rhodium Toad quickly answered (while wondering about some issues of ClearInt) with a solutions which is 3 times faster and more clean:

````
--drop function to_int_clear(text);
create or replace function to_int_clear(t text) 
returns integer 
as $body$ 
declare n numeric := substring(t from '^\s*([-+]?(?=\d|\.\d)\d*(?:\.\d*)?(?:[Ee][-+]?\d+)?)\s*$'); 
begin 
  if n not between -2147483648 and 2147483647 
    then return null; 
    else return n::float8::integer; 
  end if; 
end; 
$body$ language plpgsql strict immutable parallel safe cost 20;

-- Test
select to_int_clear(null) is null as test_true;
select to_int_clear('.') is null as test_true;
select to_int_clear('') is null as test_true;
select to_int_clear('foobar') is null as test_true;
select to_int_clear('9999999999') is null as test_true;         -- out of range, returns null
select to_int_clear('-2147483649') is null as test_true;        -- one less than the smallest possible int
select to_int_clear('2147483647') = 2147483647 as test_true;    -- largest possible int
select to_int_clear('-2147483648') = -2147483648 as test_true;  -- smallest possible int
select to_int_clear('123.456') = 123 as test_true;              -- round down
select to_int_clear('456.789') = 457 as test_true;              -- round up
select to_int_clear('  456.789  ') = 457 as test_true;          -- round up with trimming
select to_int_clear('0') = 0 as test_true;                      -- just testing...
select to_int_clear('-1') = -1 as test_true;                    -- just testing...

--- Benchmark (about 3 times faster than CleanInt)
select sum(to_int_clear(t))
from (
	select row_number() over() as r, (random()*1000)::text as t from generate_series(1,1000000)
) as tmp;
````

**Comment by nyurik** [comment]
@sfkeller i'm all for it, and thanks for the great info, but lets move it to #794 and a separate PR?  This PR is already pretty substantial :) BTW, it doesn't mean we must release this PR as v5 - we could add a few more independent PRs to master before releasing.

**Comment by MartinMikita** [comment]
Hi @nyurik,

After 10 hours running `import-sql` step, it failed with the following error:

```
psql:/sql/parallel/011.sql:410: ERROR:  function highway_class(character varying, unknown, character varying) does not exist^M
LINE 77:             AND highway_class(highway, '', construction) NOT...^M
                         ^^M
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.^M
Time: 2.265 ms^M
```

This was using your newly introduced feature - parallel SQL.

This function `highway_class` is introduced in `transportation` layer, and used in `transportation` and `transportation_name` layers.

**Comment by MartinMikita** [comment]
Note - agreed via hangout to revert the default behavior back to using non-parallel import-sql steps by @nyurik

**Comment by nyurik** [comment]
@MartinMikita I reverted the `--dir` param usage (parallel sql) until #796 is resolved.

**Comment by MartinMikita** [comment]
Thanks for the revert.
Have you tested your changes?

Because, import SQL step has changed in your v5 tools:
 - https://github.com/openmaptiles/openmaptiles-tools/blob/v5.0-dev/bin/import-sql#L83
 - https://github.com/openmaptiles/openmaptiles-tools/blob/v5.0-dev/Dockerfile#L73
 - https://github.com/nyurik/openmaptiles/blob/upgrade-v5/docker-compose.yml#L45

This means, import-sql is now looking for `tileset.sql` in the folder `build/sql/`, instead of `build/` as it was now.

**Comment by nyurik** [comment]
@MartinMikita I was waiting to push the above after the tools docker images would get built. Should pass the build now.

**Comment by nyurik** [comment]
the current code [has passed](https://travis-ci.org/github/openmaptiles/openmaptiles/builds/665891311?utm_medium=notification&utm_source=github_status), but for some reason the checkmark did not post back to this PR.

**Comment by nyurik** [comment]
Update: this PR would need https://github.com/openmaptiles/postgis-vt-util/pull/4 to be merged in order to fix an edge case I discovered with about 15 tiles -- they used `-` as a numeric value, which crashed the current implementation, thus resulting in a missing tile.  Other than that, everything imported and rendered OK.  

My servers:
* PostgreSQL server - image was created with a 100GB RAM, 16 CPU machine, controlled from another machine that ran imposm (similar specs) -- took less than a day for full planet (I forgot to optimize DB, hence took much longer than it should). I used Kubernetes for import orchestration. The server was a direct GCP VM, not a docker image (per PG dev recommendation). I used Debian 10, upgraded to Debian sid, and installed 12 + PostGIS 3.0.1 + other required extensions.
* Once imported, I started up 3x 64CPU 100GB RAM PostgreSQL servers from the same image.
* tile generation was done from a coordinating server (2 CPU, 200GB SSD) with `docker run --rm -it -v $PWD:/data nyurik/tilelive-pgquery bash` and executed the script below. The docker image code is in [nyurik/tilelive-pgquery](https://github.com/nyurik/tilelive-pgquery).  Note that z13 and z14 were not generated entirely, but instead I used z12 to look for empty tiles (identical tiles that occur in hundreds of places were assumed to be empty), and copied z12->z13. It generated a list of needed tiles (the non-empty ones), and generated. The same was done z13->z14.

```bash
#!/usr/bin/env bash

set -euo pipefail
set -x

MBTILES_ROOT=/data/2020-03  # location of the data results
MBTILES_FILE="$MBTILES_ROOT/result.mbtiles"

PGHOSTS_LIST="host=pg-3"  # &host=...&host=...
HOST_COUNT=1  # number of postgres servers
CPU=2  # number of CPUs per postgres server
MAXCONNECTIONS=$(( CPU + CPU / 10 ))
ALL_STREAMS=$(( MAXCONNECTIONS * HOST_COUNT ))

PGPASSWORD=openmaptiles
PGUSER=openmaptiles
PGDATABASE=openmaptiles

# Generate entire zoom levels until $MID_ZOOM. Afterwards only generate for those tiles that exist in the lower zoom.
START_ZOOM=0
MID_ZOOM=9
END_ZOOM=9

PGQUERY="pgquery://?database=$PGDATABASE&$PGHOSTS_LIST&username=$PGUSER&password=$PGPASSWORD&funcZXY=getmvt&name=OpenMapTiles&errorsAsEmpty=1&maxpool=$MAXCONNECTIONS&key=1&nogzip=1"

mkdir -p "$MBTILES_ROOT"
echo "Generating zoom $START_ZOOM..$MID_ZOOM from $HOST_COUNT servers, using $MAXCONNECTIONS connections per server, $ALL_STREAMS streams"
tilelive-copy --exit --minzoom=$START_ZOOM --maxzoom=$MID_ZOOM --slow=300000 --retry=2 --concurrency=$ALL_STREAMS --timeout=5400000 \
    "$PGQUERY" "$MBTILES_FILE"

for (( ZOOM=MID_ZOOM+1; ZOOM<=END_ZOOM; ZOOM++ )); do
  LIST_FILE="$MBTILES_ROOT/tiles_$ZOOM.txt"
  mbtiles-tools impute "$MBTILES_FILE" --zoom $ZOOM --output "$LIST_FILE" --verbose
  tilelive-copy --exit --scheme=list "--list=$LIST_FILE" --slow=30000 --retry=2 --concurrency=$ALL_STREAMS --timeout=300000 \
      "$PGQUERY" "$MBTILES_FILE"
done
```

z14 took about 7 hours (it could be faster now because of https://github.com/openmaptiles/postgis-vt-util/pull/4 ).  Lower zooms alltogether took about 4 hours (hard to count as I was running a few experiments).

Interesting discovery: the MVT produced by PostgreSQL directly (rather than via Mapnik) is not deterministic, possibly due to layer generation being done in parallel. I will modify the SQL to at least order by layer index, but individual features may still be produced in arbitrary order. This means that in some cases MVT tile and its ETAG may be different without the content actually being different.

docker-compose.yml
```diff
@@ -1,86 +1,51 @@
...
+
   openmaptiles-tools:
     image: "openmaptiles/openmaptiles-tools:${TOOLS_VERSION}"
     env_file: .env
+    environment:
+      # Must match the version of this file (first line)
+      # download-osm will use it when generating a composer file
+      MAKE_DC_VERSION: "2.3"
+      # Allow DIFF_MODE to be overwritten from shell
+      DIFF_MODE: ${DIFF_MODE}
```
frodrigo, line 39: Since there is already `env_file: .env`, explicit `DIFF_MODE` is not required.

zetafr:
> Update: this PR would need [openmaptiles/postgis-vt-util#4](https://github.com/openmaptiles/postgis-vt-util/pull/4) to be merged in order to fix an edge case I discovered with about 15 tiles -- they used `-` as a numeric value, which crashed the current implementation, thus resulting in a missing tile. Other than that, everything imported and rendered OK.
> 
> My servers:
> 
> * PostgreSQL server - image was created with a 100GB RAM, 16 CPU machine, controlled from another machine that ran imposm (similar specs) -- took less than a day for full planet (I forgot to optimize DB, hence took much longer than it should). I used Kubernetes for import orchestration. The server was a direct GCP VM, not a docker image (per PG dev recommendation). I used Debian 10, upgraded to Debian sid, and installed 12 + PostGIS 3.0.1 + other required extensions.
> * Once imported, I started up 3x 64CPU 100GB RAM PostgreSQL servers from the same image.
> * tile generation was done from a coordinating server (2 CPU, 200GB SSD) with `docker run --rm -it -v $PWD:/data nyurik/tilelive-pgquery bash` and executed the script below. The docker image code is in [nyurik/tilelive-pgquery](https://github.com/nyurik/tilelive-pgquery).  Note that z13 and z14 were not generated entirely, but instead I used z12 to look for empty tiles (identical tiles that occur in hundreds of places were assumed to be empty), and copied z12->z13. It generated a list of needed tiles (the non-empty ones), and generated. The same was done z13->z14.
> 
> ```shell
> #!/usr/bin/env bash
> 
> set -euo pipefail
> set -x
> 
> MBTILES_ROOT=/data/2020-03  # location of the data results
> MBTILES_FILE="$MBTILES_ROOT/result.mbtiles"
> 
> PGHOSTS_LIST="host=pg-3"  # &host=...&host=...
> HOST_COUNT=1  # number of postgres servers
> CPU=2  # number of CPUs per postgres server
> MAXCONNECTIONS=$(( CPU + CPU / 10 ))
> ALL_STREAMS=$(( MAXCONNECTIONS * HOST_COUNT ))
> 
> PGPASSWORD=openmaptiles
> PGUSER=openmaptiles
> PGDATABASE=openmaptiles
> 
> # Generate entire zoom levels until $MID_ZOOM. Afterwards only generate for those tiles that exist in the lower zoom.
> START_ZOOM=0
> MID_ZOOM=12
> END_ZOOM=14
> 
> PGQUERY="pgquery://?database=$PGDATABASE&$PGHOSTS_LIST&username=$PGUSER&password=$PGPASSWORD&funcZXY=getmvt&name=OpenMapTiles&errorsAsEmpty=1&maxpool=$MAXCONNECTIONS&key=1&nogzip=1"
> 
> mkdir -p "$MBTILES_ROOT"
> echo "Generating zoom $START_ZOOM..$MID_ZOOM from $HOST_COUNT servers, using $MAXCONNECTIONS connections per server, $ALL_STREAMS streams"
> tilelive-copy --exit --minzoom=$START_ZOOM --maxzoom=$MID_ZOOM --slow=300000 --retry=2 --concurrency=$ALL_STREAMS --timeout=5400000 \
>     "$PGQUERY" "$MBTILES_FILE"
> 
> for (( ZOOM=MID_ZOOM+1; ZOOM<=END_ZOOM; ZOOM++ )); do
>   LIST_FILE="$MBTILES_ROOT/tiles_$ZOOM.txt"
>   mbtiles-tools impute "$MBTILES_FILE" --zoom $ZOOM --output "$LIST_FILE" --verbose
>   tilelive-copy --exit --scheme=list "--list=$LIST_FILE" --slow=30000 --retry=2 --concurrency=$ALL_STREAMS --timeout=300000 \
>       "$PGQUERY" "$MBTILES_FILE"
> done
> ```
> 
> z14 took about 7 hours (it could be faster now because of [openmaptiles/postgis-vt-util#4](https://github.com/openmaptiles/postgis-vt-util/pull/4) ). Lower zooms alltogether took about 4 hours (hard to count as I was running a few experiments).
> 
> Interesting discovery: the MVT produced by PostgreSQL directly (rather than via Mapnik) is not deterministic, possibly due to layer generation being done in parallel. I will modify the SQL to at least order by layer index, but individual features may still be produced in arbitrary order. This means that in some cases MVT tile and its ETAG may be different without the content actually being different.
> 
> PostgreSQL CPU utilization (all 3 machines had near identical profile, and near zero IO usage).
> image

Hello ! 

I'm trying to generate tiles with your script but i get an error:

`
+ MBTILES_ROOT=/data/output
+ MBTILES_FILE=/data/output/result.mbtiles
+ PGHOSTS_LIST=host=postgres
+ HOST_COUNT=1
+ CPU=12
+ MAXCONNECTIONS=13
+ ALL_STREAMS=13
+ PGPASSWORD=openmaptiles
+ PGUSER=openmaptiles
+ PGDATABASE=openmaptiles
+ START_ZOOM=0
+ MID_ZOOM=12
+ END_ZOOM=14
+ PGQUERY='pgquery://?database=openmaptiles&host=postgres&username=openmaptiles&password=openmaptiles&funcZXY=getmvt&name=OpenMapTiles&errorsAsEmpty=1&maxpool=13&key=1&nogzip=1'
+ mkdir -p /data/output
+ echo 'Generating zoom 0..12 from 1 servers, using 13 connections per server, 13 streams'
Generating zoom 0..12 from 1 servers, using 13 connections per server, 13 streams
+ tilelive-copy --exit --minzoom=0 --maxzoom=12 --slow=300000 --retry=2 --concurrency=13 --timeout=5400000 'pgquery://?database=openmaptiles&host=postgres&username=openmaptiles&password=openmaptiles&funcZXY=getmvt&name=OpenMapTiles&errorsAsEmpty=1&maxpool=13&key=1&nogzip=1' /data/output/result.mbtiles
PostgreSQL connection was configured from 
tilelive-pgquery v0.7.0
Server information for postgres:5432:
                 version() = PostgreSQL 11.2 (Debian 11.2-1.pgdg90+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit
    postgis_full_version() = POSTGIS="2.5.2 r17328" [EXTENSION] PGSQL="110" GEOS="3.5.1-CAPI-1.9.1 r4246" PROJ="Rel. 4.9.3, 15 August 2016" GDAL="GDAL 2.1.2, released 2016/10/24" LIBXML="2.9.4" LIBJSON="0.12.1" LIBPROTOBUF="1.2.1" TOPOLOGY RASTER
                       jit = off
            shared_buffers = 9461MB
                  work_mem = 128MB
      maintenance_work_mem = 256MB
      effective_cache_size = 9461MB
  effective_io_concurrency = 500
           max_connections = 70
      max_worker_processes = 8
      max_parallel_workers = 8
  max_parallel_workers_per_gather = 0
               wal_buffers = 16MB
              min_wal_size = 80MB
              max_wal_size = 1GB
          random_page_cost = 1
  default_statistics_target = 100
  checkpoint_completion_target = 0.8

Verifying pgquery data source by retrieving a tile at [14,9268,3575] from postgres:5432...
Failed to get a tile at [14,9268,3575] from postgres:5432 (in 2ms), aborting tilelive-pgquery initialization:
error: function getmvt(unknown, unknown, unknown) does not exist
/usr/lib/node_modules/@mapbox/tilelive/bin/tilelive-copy:100
        if (err) throw err;
                 ^

error: function getmvt(unknown, unknown, unknown) does not exist
    at Connection.parseE (/usr/lib/node_modules/tilelive-pgquery/node_modules/pg/lib/connection.js:614:13)
    at Connection.parseMessage (/usr/lib/node_modules/tilelive-pgquery/node_modules/pg/lib/connection.js:413:19)
    at Socket.<anonymous> (/usr/lib/node_modules/tilelive-pgquery/node_modules/pg/lib/connection.js:129:22)
    at Socket.emit (events.js:311:20)
    at addChunk (_stream_readable.js:294:12)
    at readableAddChunk (_stream_readable.js:275:11)
    at Socket.Readable.push (_stream_readable.js:209:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23) {
  name: 'error',
  length: 226,
  severity: 'ERROR',
  code: '42883',
  detail: undefined,
  hint: 'No function matches the given name and argument types. You might need to add explicit type casts.',
  position: '34',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'parse_func.c',
  line: '621',
  routine: 'ParseFuncOrColumn'
}
` 


What is `function getmvt(unknown, unknown, unknown)` exactly ?

Thx

>

frodrigo:
I already use some part of tools v5 and this PR. And I have a regression. But you have the same here.

```
                                   0.0 tiles/s -67407.2% at z5, 5 tiles in 0:13:37.2 
```
z5 take minutes.

I make test on the same setup with master and this PR on faroe-islands.
The issue come from the `water` layer.

z5 on master
```
Per-layer generation speed (longer is better)
######################################################################################################################
                                                          3.9 tiles/s at layer water, 6 tiles in 0:00:01.4            
█████████████████                                       867.1 tiles/s at layer waterway, 6 tiles in 0:00:00           
███                                                     171.5 tiles/s at layer landcover, 6 tiles in 0:00:00          
█████████████████                                       844.6 tiles/s at layer landuse, 6 tiles in 0:00:00            
█████████████████████████████████████████████████████ 2,594.0 tiles/s at layer mountain_peak, 6 tiles in 0:00:00      
█████████████████████                                 1,058.2 tiles/s at layer park, 6 tiles in 0:00:00               
███████                                                 343.5 tiles/s at layer boundary, 6 tiles in 0:00:00           
████████████████████████████████████████████          2,201.8 tiles/s at layer aeroway, 6 tiles in 0:00:00            
██████                                                  314.9 tiles/s at layer transportation, 6 tiles in 0:00:00     
███████████████████████                               1,128.9 tiles/s at layer building, 6 tiles in 0:00:00           
██████████████████                                      905.9 tiles/s at layer water_name, 6 tiles in 0:00:00         
████████████████████                                    989.4 tiles/s at layer transportation_name, 6 tiles in 0:00:00
███████████                                             550.7 tiles/s at layer place, 6 tiles in 0:00:00              
████████████████████████████████████████              1,996.0 tiles/s at layer housenumber, 6 tiles in 0:00:00        
███████████████████                                     949.7 tiles/s at layer poi, 6 tiles in 0:00:00                
██████████████████████████████████                    1,675.0 tiles/s at layer aerodrome_label, 6 tiles in 0:00:00    
```

z5 on this PR
```
Per-layer generation speed (longer is better)
######################################################################################################################
                                                          0.0 tiles/s at layer water, 6 tiles in 0:13:46.6            
██████████████                                          511.7 tiles/s at layer waterway, 6 tiles in 0:00:00           
                                                          4.1 tiles/s at layer landcover, 6 tiles in 0:00:01.4        
██████████████████████                                  799.1 tiles/s at layer landuse, 6 tiles in 0:00:00            
█████████████████████████████████████████████████████ 1,867.4 tiles/s at layer mountain_peak, 6 tiles in 0:00:00      
████████████████████                                    732.2 tiles/s at layer park, 6 tiles in 0:00:00               
████████                                                285.5 tiles/s at layer boundary, 6 tiles in 0:00:00           
█████████████████████                                   744.8 tiles/s at layer aeroway, 6 tiles in 0:00:00            
██████████                                              352.4 tiles/s at layer transportation, 6 tiles in 0:00:00     
████████████████████████████████████████              1,426.9 tiles/s at layer building, 6 tiles in 0:00:00           
███████████                                             402.0 tiles/s at layer water_name, 6 tiles in 0:00:00         
██████████████                                          518.4 tiles/s at layer transportation_name, 6 tiles in 0:00:00
██████████                                              365.3 tiles/s at layer place, 6 tiles in 0:00:00              
                                                         18.7 tiles/s at layer housenumber, 6 tiles in 0:00:00.3      
███████████████                                         530.1 tiles/s at layer poi, 6 tiles in 0:00:00                
███████████████████████████████████                   1,253.1 tiles/s at layer aerodrome_label, 6 tiles in 0:00:00    
```

The issue come from the usage of `ne_10m_ocean` only used at z5. But I still not spotted why.

nyurik:
Update: there was a bug backporting preloaded data image, which has now been fixed. Everything seems to be building ok.

nyurik:
@frodrigo thanks for looking in depth at the SQL!  One thing to note though is that now this branch should have the same performance as before because the `postgis` image usen the same versions of Postgres and Postgis as before -- PG 9.6 and Postgis 2.4.8.  The next PR will bump just the postgis version (without Postgres), and yet another PR will migrate just the PG.

frodrigo:
This perf issue is no more linked to this PR. But note, with Postgis 3, it is ok.
```
SELECT version();
PostgreSQL 11.7 (Debian 11.7-2.pgdg90+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170
516, 64-bit

SELECT PostGIS_full_version();
POSTGIS="3.0.1 ec2a9aa" [EXTENSION] PGSQL="110" GEOS="3.7.1-CAPI-1.11.1 27a5e771" PROJ="Rel. 4.9.3, 15 August 2016" L
IBXML="2.9.4" LIBJSON="0.12.1" LIBPROTOBUF="1.2.1" WAGYU="0.4.3 (Internal)"
```

```
debug-mvt openmaptiles.yaml 5/0/0 -l water
======= Layer water: 78 bytes in MVT =======
  NOTICE: Ring Self-intersection at or near point 19910010.296251766 11475371.091019826 @ None
brunnel    class    intermittent    is_valid_mvt    mvtgeometry    is_valid_geom    geometry
---------  -------  --------------  --------------  -------------  ---------------  ---------------------
           ocean                    True            POLYGON(120)   False            MULTIPOLYGON(7163616)
0m2.299s
```

Issue #202, opened 2017-04-01
make docker-unnecessary-clean


pathmapper:
after executing: 
`make docker-unnecessary-clean`
and 
` ./quickstart.sh <<my-area>>    `

it seems like there are still 
- 3 versions of openmaptiles/import-osm
- 2 versions of openmaptiles/import-water
- 2 versions of openmaptiles/generate-vectortiles

available.

![docker_images](https://cloud.githubusercontent.com/assets/20856381/24576548/dd9f6a1a-16bd-11e7-9427-eb137c010516.png)

are these different versions really necessary?

ImreSamu:
>are these different versions really necessary?

It is depends of your use case. 

if you only need -  the  referenced  images **in the actual git version**,  then  you can examine the actual [docker-compose.yml](https://github.com/openmaptiles/openmaptiles/blob/master/docker-compose.yml)    for this images.

or you can  query yourself like:
```bash
$ docker-compose config | grep " image: " | sort -u 
    image: openmaptiles/generate-vectortiles:0.1.1
    image: openmaptiles/import-lakelines:0.9
    image: openmaptiles/import-natural-earth:1.2
    image: openmaptiles/import-osm:0.4
    image: openmaptiles/import-osmborder:0.2
    image: openmaptiles/import-sql:0.1
    image: openmaptiles/import-water:0.3.1
    image: openmaptiles/openmaptiles-tools:latest
    image: openmaptiles/postgis:2.3
    image: osm2vectortiles/mapbox-studio
```
and manually remove the unwanted images.  (   see  [docker rmi](https://docs.docker.com/engine/reference/commandline/rmi/) )

I you don't know docker-compose there are some good tutorials:  https://docs.docker.com/compose/gettingstarted/

**If you have some magic script - for removing old and tagged openmaptiles images I am open for testing**

pathmapper:
> if you only need - the referenced images in the actual git version

This is exactly I was looking for. Thank you for your explanations!
I am using docker only for OMT at the moment, so it is a manageable job to do it manually.

Sorry, no magic script here. Looking for some inspiration led me to this thread
[https://forums.docker.com/t/command-to-remove-all-unused-images/20](https://forums.docker.com/t/command-to-remove-all-unused-images/20
)
and the information, that since docker version 1.13 there are new commands available like

[docker system prune](https://docs.docker.com/engine/reference/commandline/system_prune/)
and
[...]

PR #795, opened 2020-03-20
Makefile: Use POSTGRES settings from .env

zstadler:
Avoid replication of settings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,12 @@
+# Include common settings
+include .env
+# Ensure POSTGRES settings exist for backward compatibility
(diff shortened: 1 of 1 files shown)


nyurik:
I like the idea, but lets wait until we merge v5?  #785 -- some of the variables have been renamed

nyurik:
After further discussion with @zstadler , I think we should rather go with the approach proposed in #202 . This will keep everything inside the .env, and won't expose it to Makefile.  Thus I would rather decline.  Feel free to reopen or discuss it further.

PR #806, opened 2020-04-08
Support multiple openmaptiles instances by specifying a Docker Compose project

zstadler:
This PR was suggested by @nyurik:

> zstadler  3:30 PM
Is it possible to run two instances of openmaptiles, and the postgis container in particular, on the same Linux host?
nyurik  4:17 PM
use docker-compose --project-name -- that should allow you to run everything in parallel
4:17
might need to update make file
4:17
btw, that would be a good PR for makefile -- to specify --project-name based on the current DIR name, but so that it can be overwritten by a makefile param

The implementation adds the `DC_PROJECT` parameter. It can also be set by an environment variable in the hosting shell. The environment variable can be overwritten by a make parameter, including `DC_PROJECT=` which restores the automatic project name.

#### _NOTE:_
It may be worthwhile to review the following make targets that are currently not impacted by the PR:
- `make start-tileserver`
- `make start-postserve`
- `make list-docker-images`
- `make remove-docker-images`
- `make docker-unnecessary-clean`


Makefile
```diff
@@ -57,42 +63,42 @@ init-dirs:
...
 	rm -rf build
 
 .PHONY: clean-docker
 clean-docker:
-	docker-compose down -v --remove-orphans
-	docker-compose rm -fv
-	docker volume ls -q | grep openmaptiles  | xargs -r docker volume rm || true
+	$(DOCKER_COMPOSE) down -v --remove-orphans
+	$(DOCKER_COMPOSE) rm -fv
+	docker volume ls -q -f "name=^$(DC_PROJECT)_" | xargs --no-run-if-empty docker volume rm
```
nyurik, line 82: `--no-run-if-empty` does not work on MacOS, but AFAIK MacOS version simply does not run `xargs` if input is empty.  One simple option would be to introduce `$(XARGS)` param, e.g.

```
XARGS=$(shell xargs --version|grep -q GNU && echo 'xargs --no-run-if-empty' || echo 'xargs')
```

Makefile
```diff
@@ -241,30 +247,30 @@ list-tables:
...
 .PHONY: refresh-docker-images
 refresh-docker-images:
-	docker-compose pull --ignore-pull-failures
+	$(DOCKER_COMPOSE) pull --ignore-pull-failures
 
 .PHONY: remove-docker-images
 remove-docker-images:
 	@echo "Deleting all openmaptiles related docker image(s)..."
-	@docker-compose down
+	@docker-compose --project-name $(DC_PROJECT) down
```
nyurik, line 273: why not `$(DOCKER_COMPOSE) down` ?

Makefile
zstadler: How about a blunt yet robust approach:
```
XARGS:=xargs $(shell xargs --no-run-if-empty </dev/null 2>/dev/null && echo --no-run-if-empty)
```

Makefile
zstadler: Of course. I must have missed it.

okft:
I don't if it was intended but there is now a '-' in front of "-docker rm -f maputnik_editor".

nyurik:
@okft the `-` in a makefile makes it ignore the errors

okft:
Ok never mind then!

PR #812, opened 2020-04-18
Update Makefile

zstadler:
- Add `download-osmfr` and `download-bbbike` targets
- Port `DC_OPTS` to Windows
- Use make conditions instead of shell. Also simplifies `make -n` output
- Use remote Docker machine's IP, if defined, rather than `localhost` in `http://...` messages. Also applicable for Docker Toolbox for Windows.
- Align texts in `make help` output
- Update and bug-fix in `make remove-docker-images`


Makefile
```diff
@@ -33,19 +38,21 @@ help:
...
-	@echo "  make start-tileserver                # start klokantech/tileserver-gl [ see localhost:8080 ] "
+	@echo "  make start-postserve                 # start Postserver + Maputnik Editor [ see $(OMT_HOST):8088 ] "
+	@echo "  make start-tileserver                # start klokantech/tileserver-gl     [ see $(OMT_HOST):8080 ] "
 	@echo " "
 	@echo "Hints for developers:"
 	@echo "  make                                 # build source code"
 	@echo "  make list-geofabrik                  # list actual geofabrik OSM extracts for download"
-	@echo "  make download-geofabrik area=albania # download OSM data from geofabrik, and create config file"
+	@echo "  make download-geofabrik area=albania # download OSM data from geofabrik,       and create config file"
+	@echo "  make download-osmfr area=asia/qatar  # download OSM data from opestreetmap.fr, and create config file"
```
okft, line 48: Typo: opestreetmap.fr -> openstreetmap.fr

zstadler: Thanks!
:+1:

Makefile
```diff
@@ -95,28 +102,31 @@ db-start:
...
+	@echo "ERROR: Unable to download an area if area is not given."
+	@echo "Usage:"
+	@echo "  make download-$(OSM_SERVER) area=<area-id>"
+	@echo ""
+	$(if $(filter %-geofabrik,$@),@echo "Use   make list-geofabrik   to get a list of all available areas";echo "")
+	@exit 1
+else
+	@echo "=============== download-$(OSM_SERVER) ======================="
+	@echo "Download area: $(area)"
+	@rm -fr ./data/*.pbf
```
nyurik, line 121: I'm a bit on a fence here about this cleanup. First off, the `data/` could also contain `docker-compose-config.yml` and possibly other files generated by `download-osm`. If cleaning up, they also should be cleaned up.

On the other hand, i feel the command `download` does not imply deletion of the existing stuff. I am totally OK with adding a big warning in case the directory already contains another PBF file named something different from `$(area)-latest.pbf` -- this way an existing file will get replaced (or we can delete it because aria2c might create a new copy otherwise), but if someone wants to have multiple pbf files in the data dir, its their choice...

Alternatively we could just through an error in case a pbf file already exists -- makes things more predictable...

nyurik:
Awesome PR, thanks! There is one small issue with removal of the data on download that i think we should decide on, but otherwise looks great!

Makefile
zstadler: I didn't mean to change of the logic of the current flow.

The line exists in

https://github.com/openmaptiles/openmaptiles/blob/746a2dc7b93386768f08dc9524d5fce99a3b5d37/quickstart.sh#L140-L141

Its goal is to avoid potential problems with the `*_with_first_pbf` behavior of downstream steps in the flow.

IMHO, it should be part of the Makefile. Without it, the following sequence 

```
make download-geofabrik area=albania
# ...
make download-geofabrik area=zambia
# ...
make import-osm
```

will import the **albania** data rather than **zambia**. See [import_osm.sh](https://github.com/openmaptiles/openmaptiles-tools/blob/55dd0b4622d8a18b86d74900a8fd0f5c2e7b324b/docker/import-osm/import_osm.sh#L41-L55)

nyurik: @zstadler there is a big difference between the `quickstart` doing a bunch of steps including data cleanup, and modifying `download` make target.

I agree that we may want to modify the make target, but the requirements/expectations for it are a bit different. The quickstart expects there to be just a single file, but the [future version](https://github.com/openmaptiles/openmaptiles-tools/pull/203/files#diff-0ef5c4ad508074e3fecb96fb655f1303R59) of `import-osm` will support a file parameter too.

So for now, I think it would be best not to change the current behavior, or maybe add a big warning in case another pbf file exists, but still allow it.

zstadler: @nyurik the goal is to protect the direct use of the Makefile, not the use of `quickstart.sh`.

To make use of the future version of `import-osm`, the future version of `Makefile` and would need to pass `$(area)-latest.osm.pbf` to `import-osm`. Unlike `quickstart.sh`, however, `Makefile` does not have a default area. Only the `download-*` targets require `$(area)` to be set. Bridging these requirements, while permitting multiple `data/*.pbf` files, could introduce a greater incompatibility than the one discussed here. 

On the other hand, I cannot see a user flow that would be harmed from the discussed change. At a minimum, such a flow would use multiple `data/*.pbf` files and would use `make download-*` followed by `make import-osm` according to their reverse alphabetical order.
I also have no idea, for example, what would happen if `make import-osm` is called multiple times with different `pbf` files.

I'm a very strong advocate of backward-compatibility. In this case, IMHO, preventing users from unknowingly making errors has a greater value.

nyurik: @zstadler i agree that the current method is not ideal, but it does have some usages that i would like to discuss.  Lets merge all other changes, and we can discuss the `rm -rf data` in a different PR?  Some of the changes you made will start conflicting with other changes soon, so the quicker we merge them the better.

zstadler: Sounds good.

Issue #813, opened 2020-04-19
Extend GitHub Actions to test size and generating performance for PRs.


TomPohys:
GitHub Actions should not only run through `quickstart` but also test the size and performance of new PRs.

nyurik:
We need to pick the target areas which we will test on.  Looking at the latest Geofabrik extracts, these are the smallest ones.  Which should we use?  We can [combine](https://gis.stackexchange.com/questions/242704/merging-osm-pbf-files) several into a single file.

area | size
-- | --
europe/monaco | 392,742
africa/saint-helena-ascension-and-tristan-da-cunha | 404,608
africa/sao-tome-and-principe | 737,850
europe/great-britain/england/rutland | 1,245,960
europe/andorra | 1,664,842
africa/equatorial-guinea | 2,184,657
europe/liechtenstein | 2,344,565
europe/isle-of-man | 2,393,136
africa/seychelles | 2,396,110
europe/great-britain/england/london/enfield | 2,509,057
africa/comores | 3,023,615
asia/maldives | 3,085,039
europe/faroe-islands | 3,390,201
europe/france/mayotte | 4,314,030
europe/malta | 4,630,071
africa/mauritius | 5,438,263
central-america/belize | 5,689,304
europe/great-britain/england/bedfordshire | 6,801,084
europe/great-britain/england/herefordshire | 6,813,186
europe/great-britain/england/isle-of-wight | 6,862,216
africa/djibouti | 7,112,533
asia/bhutan | 7,381,412
africa/eritrea | 7,525,215
africa/guinea-bissau | 7,526,716
africa/cape-verde | 7,762,585
africa/gabon | 8,025,899
australia-oceania/fiji | 9,541,916
central-america/jamaica | 9,812,405

klokan:
Liechtenstein used to be our smallest test area in the past, Zurich our second, Switzerland and Czech Republic as third - US New York and San Francisco.

We need ideally a data-rich region with different tags - on different parts of the planet.

nyurik:
@klokan thx, I can easily merge multiple regions into a standard "test" dataset, which I will upload to the data store you've set up.  We should also include some low-density area to evaluate near-empty tile generation speed.

nyurik:
@klokan thx, I can easily merge multiple regions into a standard "test" dataset, which I will upload to the data store you've set up.  We should also include some low-density area to evaluate near-empty tile generation speed.

TomPohys:
[...]

PR #817, opened 2020-04-22
[WIP] automated performance testing

nyurik:
Work in progress to add automated performance testing


.github/workflows/omt_ci.yml
```diff
@@ -15,29 +15,68 @@ on:
...
+        make db-start
+        make import-water
+        make import-natural-earth
+        make import-lakelines
+        make import-osm
+        make import-borders
+        make import-wikidata
+        make import-sql
+    - name: generate test data
+      run: docker-compose run --rm openmaptiles-tools test-perf openmaptiles.yaml -t eu-london -r test_results.json
```
nyurik, line 41: We would need two actions:
* one that runs on `master merge`. It will use `--record master_results.json`, and upload the results to the storage.
* another action that runs `on pull request`.  It will download `master_results.json`, run with two options `--record PR_1234_results.json --compare master_results.json` (1234 is the PR number), and upload the PR results to the storage.

I think we should use long form of the parameters to make them more readable.
Also, we will need to decide on the actual testing, as described in https://github.com/openmaptiles/openmaptiles/issues/813 .

Take a look at https://github.com/openmaptiles/openmaptiles-tools/pull/221/files -- we could define tests this way. Both of the above actions would need to run with `--test <area1> --test <area2> ...`

nyurik:
superceeded by #824

PR #821, opened 2020-04-22
Makefile target cleanup, minor improvements

nyurik:
* allow postgres image to be overwritten with an env var
* allow DIFF_MODE var to be overwritten with an env var
* add /mapping and /cache dirs into tools image
* make `build-sql` target explicit rather than relying on a filename
* `tools-dev` will open a shell in a docker to experiment and debug (instead of `import-sql-dev` and `import-osm-dev`)
* separated `start-maputnik` from `start-postserve`
* renamed `clean-docker` into `db-destroy` to make it more explicit
* cleaner `db-start`, `db-stop`, `db-destroy` targets
* better output messages

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ else
 endif
 
 .PHONY: all
-all: build/openmaptiles.tm2source/data.yml build/mapping.yaml build/tileset.sql
+all: build/openmaptiles.tm2source/data.yml build/mapping.yaml build-sql
 
 # Set OpenMapTiles host
 OMT_HOST:=http://$(firstword $(subst :, ,$(subst tcp://,,$(DOCKER_HOST))) localhost)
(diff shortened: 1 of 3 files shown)


zstadler:
Before this commit, 20 of the 29 make target names were using a verb-object template like `generate-tiles`. Now it's 20 of 34.

I find `make start-postserve` more natural than `make postserve-start`

The following changes would unify the targets' naming conventions to the verb-object template :

current name | proposed name
---|---
db-destroy | destroy-db
db-start | start-db
db-stop | stop-db
docker-unnecessary-clean | clean-unnecessary-docker
maputnik-start | start-maputnik
maputnik-stop | stop-maputnik
postserve-start | start-postserve
postserve-stop | stop-postserve
psql-analyze | analyze-psql
psql-pg-stat-reset | reset-pg-stat-psql
psql-vacuum-analyze | vacuum-analyze-psql
tileserver-start | start-tileserver

We may also  consider changing the unclear verb-less `*-dev` terget names

current name | proposed name
---|---
import-osm-dev | import-osm-shell
tools-dev | tools-shell

nyurik:
I'm ok with that too, could you submit a PR?

nyurik:
Also, bikesheding on the names a bit more --

now | proposed | counterproposal
-- | --
psql-analyze | analyze-psql | analyze-db
psql-pg-stat-reset | reset-pg-stat-psql | reset-db-stats
psql-vacuum-analyze | vacuum-analyze-psql | vacuum-db
import-osm-dev | import-osm-shell | bash (merged)
tools-dev | tools-shell | bash (merged)

PR #827, opened 2020-04-24
Use osm_id as feature id for building layer

frodrigo:
Change the feature_id of objects in the tiles from arbitrary number to osm_id.

From discussion here https://github.com/openmaptiles/openmaptiles/pull/725#issuecomment-619059931

Note the multiploygon building are load with negative relation osm_id. But the use in feature_id do an overflow (eg -7980888 -> 18446744073701570000). We can keep as is or not.

No stats done for yet. Trying to let the CI do it.


nyurik:
From the [Post a comment on Pull Request](https://github.com/openmaptiles/openmaptiles/pull/827/checks?check_run_id=616034524) step of the perf evaluation, it shows the tile sizes might grow by 5% in the high-population area.  The generation speed is currently less reliable, so ignore that.

```

  Per-test average tile sizes (shorter is better)
  ##############################################################################################
  █                    133.2 per tile +0.8% in test equatorial-guinea, 117459 tiles in 0:05:52.6
  ████████████████████  2.0K per tile +5.0% in test liechtenstein, 1511 tiles in 0:00:11.5 
```

frodrigo:
With the sign of osm_id in lower byte, is even worse than let the overflow.
```
        CASE WHEN osm_id > 0 THEN osm_id * 2
            ELSE -osm_id * 2 + 1
        END osm_id,
```
The Liechtenstein z14 size move from +5.0% to +5.1%.

frodrigo:
Rebase and remove the attempt to use sign in lower bit.

nyurik:
I don't think we should merge this before migrating to postgis3.0. the as-mvt support for IDs was added in v3. Also, there is currently a big discrepancy between testing and generation. Our testing uses as-mvt, but generation still uses antiquated Mapnik lib.

PR #833, opened 2020-04-27
Implement PR performance auto-update

nyurik:
A cron-based approach to find pull requests, possibly from forks,
that finished profiling, and post their results as comments.

See in-depth explanation of how this works at
https://github.com/nyurik/auto_pr_comments_from_forks


nyurik:
Hehe, same here... if i didn't mess up anything :)

PR #835, opened 2020-04-28
Improve PR comment updater, bug fixes

nyurik:
* use proper number parsing
* skip PRs that weren't recently updated
* Better reporting


TomPohys:
There is a slower tile generation by 10% even the PR do not change generating tiles.

PR #836, opened 2020-04-28
PR workflow - error reporting and cleanup

nyurik:
optimize a bit the CI workflow PR commenting logic
to avoid some errors and better debugging messages.


nyurik:
selfmerged to avoid error flood in the case of no recent PRs. Non-consequential for the general OMT code.

PR #837, opened 2020-04-28
display raw data on error in workflow

nyurik:
trying to figure out what strange data is returned
by github that is not returned when testing locally


nyurik:
It appears github actions have been a bit unstable [lately](https://www.githubstatus.com/)... adding a few more checks

PR #838, opened 2020-04-29
Workflow - use insecure curl when can

nyurik:
Turned out that some update jobs failed due to

```
{
  "message": "Bad credentials",
  "documentation_url": "https://developer.github.com/v3"
}
```

This is probably due to credentials expiring (long workflow startup?),
or some internal github issue.

For now, removing authenticated `curl` calls because most
of them can be done anonymously, and keeping them only when needed.


nyurik:
Hopefully last iterative workflow patch -- these for some reason do not reproduce elsewhere, so had to do them on OMT repo.  Now should have fewer github action run errors.

Issue #814, opened 2020-04-20
update-osm taking a long time to import even small planet hangesets (Postgres COMMIT slow)


skylarmt:
When using update-osm to import daily changesets, there are about five or ten minutes of "work" being done, followed by six to twelve hours where the import time increases, nothing else changes, and Postgres is pegging a single core at 100%, doing something to the disk at around 80+ megabytes per second, and appearing to just hang.

I also tried to import an hourly changeset (they're only a couple megabytes!) and killed it at the one hour mark; it was behaving the same as for the daily changesets.

This is the kind of output I've been getting:

```
skylar@mapbeast:~/openmaptiles$ docker-compose run --rm update-osm
[Apr 20 00:00:50] [INFO] [diff] Parsing changes, updating cache and removing elements took: 13.908048514s
[Apr 20 00:00:50] [INFO] [   13s] C:    8000/s (44702) N:     100/s (790) W:    1000/s (8643) R:      0/s (328)
[Apr 20 00:00:59] [INFO] [PostGIS] Updating generalized tables took: 1.550812465s
[Apr 20 00:01:50] [INFO] [  1m0s] C:       0/s (0) N:    1700/s (1709) W:     900/s (5612) R:    180/s (104)
[Apr 20 00:02:50] [INFO] [  2m0s] C:       0/s (0) N:    1700/s (1709) W:     900/s (5612) R:    180/s (104)
[Apr 20 00:03:50] [INFO] [  3m0s] C:       0/s (0) N:    1700/s (1709) W:     900/s (5612) R:    180/s (104)
[Apr 20 00:04:50] [INFO] [  4m0s] C:       0/s (0) N:    1700/s (1709) W:     900/s (5612) R:    180/s (104)
[Apr 20 00:05:50] [INFO] [  5m0s] C:       0/s (0) N:    1700/s (1709) W:     900/s (5612) R:    180/s (104)
[Apr 20 00:06:50] [INFO] [  6m0s] C:       0/s (0) N:    1700/s (1709) W:     900/s (5612) R:    180/s (104)
[Apr 20 00:07:50] [INFO] [  7m0s] C:       0/s (0) N:    1700/s (1709) W:     900/s (5612) R:    180/s (104)
[Apr 20 00:08:50] [INFO] [  8m0s] C:       0/s (0) N:    1700/s (1709) W:     900/s (5612) R:    180/s (104)
...etc, etc, etc...
```

[...]

Issue #809, opened 2020-04-16
Discussion: Snapshot vs dynamic tile gen optimizations


nyurik:
Per April community sync up, layer SQL could be optimized to quickly generate the entire planet (snapshot), or to have a fast real time tile service generating tiles on the fly, updating data every minute from OSM.  The snapshot generation speed benefits from the materialized views. The real time service need to work without mat views because they are very slow to update.

This discussion is only for the multiple ways to generate **identical** data, not for the multiple similar layers having different data.

### Assumptions and Requirements
* Materialized views require primary keys in order to be refreshed without stopping the service. Some OMT mat views might not have that, so (1) we should work on making all mat views to have a primary key, and (2) in the mean time the realtime users may need to take some servers out of rotation in order to refresh mat views.
* continue keeping the data in the low zooms z0..z8 (?) as materialized views for both snapshot and realtime - they do not update often enough, so should be OK not to refresh as often.
* layers with the higher zooms z8+ (?) should have a way to define alternative SQL with and without mat views.

### Proposed solution
Current layer definition file uses `schema` to list sql files (in the order they should be imported):

```yaml
# layers/building/building.yaml
schema:
  - ./buildings.sql
```

In addition to a string value, we could also allow an object that functions as a switch statement - depending on the current mode, it will use either the snapshot or the realtime file. Tools will assume `snapshot` by default. Each value is a file or a list of files.

```yaml
# layers/building/building.yaml
schema:
  - snapshot: ./buildings-snapshot.sql
    realtime: ./buildings-realtime.sql
  - ./buildings.sql
```

The actual implementation could use `CREATE VIEW` instead of the materialized views but with the same output -- this way no other SQL code would need to be changed.

zstadler:
In addition to the performance aspect, there is a functionality aspect for this discussion. During their refresh, materialized views are locked for reading, such as serving dynamic tiles or `generate-tiles-diff`. https://github.com/openmaptiles/openmaptiles/issues/787 and https://github.com/openmaptiles/openmaptiles/issues/819 seem to hit this problem. 

The `REFRESH CONCURRENTLY` option can be used to allow read access, but the refresh performance is worse and the option cannot be deployed on all views

> _This option is only allowed if there is at least one UNIQUE index on the materialized view which uses only column names and includes all rows_

[...]

PR #853, opened 2020-05-05
Replace materialized view of water by tables with diff update

frodrigo:
Replacing materialized view by a tables with update from trigger on change only.

Start with the most simple cases.

Just replicate the change on:
* `osm_water_polygon` to `osm_water_lakeline`,
* `osm_water_polygon` to `osm_water_point`.

Use a view to factorize the `osm_water_lakeline` and `osm_water_point_view` definition and reuse it in the trigger.

The update of `osm_important_waterway_linestring` is more complex, as it is a merge of `osm_waterway_linestring`. It not done in the same way. At the end of the transaction we remove impacted and recompute them.

The goal is to update more quickly the content of derivated table by just updating the changing content. It replaces the update of materialized view because their need a full recompute (with lock issue).

Note, an advanced version of differential update over materialized view as already implemented in the building cluster PR #725.

It addresses #814 and a part of #809.

### DIFF
--- a/layers/water_name/update_water_lakeline.sql
+++ b/layers/water_name/update_water_lakeline.sql
@@ -1,11 +1,8 @@
-DROP TRIGGER IF EXISTS trigger_flag_line ON osm_water_polygon;
-DROP TRIGGER IF EXISTS trigger_refresh ON water_lakeline.updates;
+DROP TRIGGER IF EXISTS trigger_delete_line ON osm_water_polygon;
+DROP TRIGGER IF EXISTS trigger_update_line ON osm_water_polygon;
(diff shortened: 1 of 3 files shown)


nyurik:
@frodrigo could you allow editing when submitting? Would make it a bit easier to upgrade to master in order to re-run with the new base.  Use this checkbox when creating PRs. Thanks!

image

frodrigo:
It is done.

nyurik:
> It is done.

@frodrigo are you sure the checkbox on the right side of this conversation is checked? I can't update your branch (which is what we have to do before merging)

frodrigo:
I rabsed It.

I don't know why, but the box does not keep the check.

Issue #820, opened 2020-04-22
Dicussion: Makefile support for multiple data/*.pbf files


zstadler:
This discussion is about the way `Makefile` is expected to handle a potential situation of multiple `data/*.pbf` files.

The discussion is a spin-off of [this discussion](https://github.com/openmaptiles/openmaptiles/pull/812#pullrequestreview-395977085).
_Note: this is irrelevant for `quickstart.sh` that prevents this condition._

#### Problem Description
`make download-*` allows multiple `data/*.pbf` files to be created.
`make import-osm` [imports the first `data/*.pbf` file found](https://github.com/openmaptiles/openmaptiles-tools/blob/55dd0b4622d8a18b86d74900a8fd0f5c2e7b324b/docker/import-osm/import_osm.sh#L41-L55). 

As a result, the following sequence will import the **albania** data rather than **zambia**, unlike the expectations of an ordinary user. 

```
make download-geofabrik area=albania
# ...
make download-geofabrik area=zambia
# ...
make import-osm
```

#### Discussion framework
I suggest the discussion will be done in two phases:
1. Discuss the user-flows that use `make download-*` and `make import-osm` with multiple `data/*.pbf` files.
2. Explore the possible ways to consolidate
   - The expectations of users of the above flows 
   - The expectations of ordinary users of the Makefile
   - The desire to maintain backward-compatibility

@nyurik, can you start the discussion on the user-flows, as [you proposed here](https://github.com/openmaptiles/openmaptiles/pull/812#discussion_r412316627)?

nyurik:
My perspective on this is that we should limit downloads if possible, so ideally we should move towards the situation where this sequence would only download two files:

```
./quickstart albania

nyurik:
I added the implementation of this as part of the #855 -- allows `make import-data PBF_FILE=...`. This way one could want to download multiple files using `make download-geofabrik ...`, and use different `import-data` without redownloading

TomPohys:
Thank @nyurik for PR and @zstadler for discussion.

PR #856, opened 2020-05-06
parametrized pbf file

nyurik:
* `make import-data` now accepts `PBF_FILE=...` parameter (see #820)
* quickstart uses specific data file when importing

### DIFF
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -27,7 +27,8 @@ if [ $# -eq 0 ]; then
 else
     osm_area=$1
 fi
-testdata="${osm_area}-latest.osm.pbf"
+
+pbf_file="./data/${osm_area}-latest.osm.pbf"
 
 ##  Min versions ...
 MIN_COMPOSE_VER=1.7.1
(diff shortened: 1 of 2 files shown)


quickstart.sh
```diff
@@ -139,7 +140,7 @@ echo "--------------------------------------------------------------------------
 echo "====> : Removing old MBTILES if exists ( ./data/*.mbtiles ) "
 rm -f ./data/*.mbtiles
 
-if [[ ! -f "./data/${testdata}" || ! -f "./data/docker-compose-config.yml" ]]; then
+if [[ ! -f "${pbf_file}" || ! -f "./data/docker-compose-config.yml" ]]; then
     echo " "
     echo "-------------------------------------------------------------------------------------"
     echo "====> : Downloading ${osm_area} from Geofabrik..."
```
zstadler, line 146: Would you also consider removing https://github.com/openmaptiles/openmaptiles/blob/dd3cb077a267733bea8fc18ab746413065b725e8/quickstart.sh#L147 below?

nyurik: @zstadler that's a good point, thx, need to clean it up a.

PR #858, opened 2020-05-07
Use CleanNumeric in buildings layer

nyurik:
`nullif(as_numeric(height),-1)` is doing a double conversion -- null into -1 followed by -1 back into null. Remove that, and delete unused `as_numeric`


nyurik:
For some reason this change seems to often show slower perf tests. Not sure what causes it.

nyurik:
@Duiesel thx, but I think I will wait to merge until the next version of tools is out because it has a few performance optimizations related to this.

nyurik:
Now that v5.1 tools has been released, this can be merged as part of v5.1 upgrade.

Duiesel:
> > as_numeric is not parallel safe. CleanNumeric works with #839
> 
> #839 is unrelated -- it simply generates multiple SQL files instead of one. Parallel safety breaks in a different place -- if a function marked as `parallel safe` calls `as_numeric`, PostgreSQL will throw an error.

Oh, I got that error. So is it fixed in v5.1 of tools?

TomPohys:
Hi @Duiesel, yes, it is.

PR #861, opened 2020-05-08
quickstart support for osmfr and bbike areas

zstadler:
- Geofabrik as the default server
- osmfr is used for hierarchical area names such as `europe/austria`
- bbbike is used for Capitalized area names such as `Adelaide`
- Use the `--empty` flag to start with an empty database


nyurik:
Thanks for a good suggestion. Lets make the param ordering the same as in `download-osm` -- `geofabrik <area>` rather than `<area> geofabrik`. Could just test if the first word is one of the known ones, and if not, prepend `geofabrik` to it, and pass the whole string (except the `--empty`) to `download-osm` as is without modifications.

zstadler:
There are many reasons to prefer the proposed parameter order to `quickstart.sh`, including:
- The vast majority of `quickstart.sh` users would not use the server parameter. It is needed only if the user wants to download a continent from osmfr. 
- It is unusual for scripts to have a sub-option before the option itself, as may be the case if the usage would be:
```
./quickstart.sh [--empty] [[geofabrik|osmfr|bbbike] area]
```
- Parameters are passed to `make` by name, not by order.

nyurik:
Thanks, makes sense. If osm.fr supports catalog, i think we should do similar code there as I did for geofabrik - so that even a substring of the full (hierarchical) name would work.

One other issue, possibly addressable by a separate pr, is that we keep stumbling on the fact that the downloaded filename is "semi-"unknown by the quickstart. Instead, I think we may want to update download-osm to accept the "output" param -- this way it will preconfigure aria2c to output the file where it is needed, and quickstart will just pass in the name of the file it wants.

nyurik:
P.S. WRT urls -- we should do a check if the aria name begins with `http`, and if so, use `url` param. Eventually download-osm should be smart enough to deal with it automatically.

zstadler:
As far as I see, the downloaded filename is known by `quickstart.sh` https://github.com/openmaptiles/openmaptiles/blob/b7b92542338ff068e5f450b5aebb873a69813002/quickstart.sh#L70

nyurik:
> As far as I see, the downloaded filename is known by `quickstart.sh`
> 
> https://github.com/openmaptiles/openmaptiles/blob/b7b92542338ff068e5f450b5aebb873a69813002/quickstart.sh#L70

Not exactly -- it is not "known", it is "assumed". When downloading, aria2c will check if the file already exists, and if it does, it will download it as `*_1` (or something like that).  I think we should make this a parameter to `download-osm`, so that it checks if the file already exists, and fail (unless there is a `--force`), and also pass a parameter to aria2c to use that filename + filepath as the location.

zstadler:
> 
> 
> > As far as I see, the downloaded filename is known by `quickstart.sh`
> > https://github.com/openmaptiles/openmaptiles/blob/b7b92542338ff068e5f450b5aebb873a69813002/quickstart.sh#L70
> 
> Not exactly -- it is not "known", it is "assumed". When downloading, aria2c will check if the file already exists, and if it does, it will download it as `*_1` (or something like that). I think we should make this a parameter to `download-osm`, so that it checks if the file already exists, and fail (unless there is a `--force`), and also pass a parameter to aria2c to use that filename + filepath as the location.

`make download-*` overwrites the pbf file. This is also what a user would expect.

nyurik:
we have 3 levels of "abstraction":
* `download-osm` -- has no preconceived notion of where it will output the file, or the file's name.
* `Makefile` -- does not know the name of the output file, but specifies that the file should be placed into the `/import` dir.
* `quickstart` -- assumes that if it deletes the `/import/*.pbf` files, and download `area`, the resulting file will be `/import/area-latest.osm.pbf`.

I think the above assumptions are too magical and inconsistent.  Instead, I think we need to make one simple change to the `download-osm` -- make it aware of what it will download by adding the `--output` parameter.  This way quickstart will know for sure the name of the PBF file, and will be able to test if that file exists properly.

A separate consideration, maybe for the future, is that I think Makefile is more trouble than its worth. We use it as a simple wrapper around `docker run`, and as such, we might as well use some simple scripts instead.  But that's a much bigger discussion perhaps for the future.

zstadler:
I suggest that the "assumptions" will be coded as the defaults and overrides will be added:
- `download-osm` will accept an optional `--output` flag
- If the `PBF_FILE` environment is set, `make download-*` will also use it to pass a `--output` flag to `download-osm`.
- Since `quickstart.sh` used the default, it need not pass to to `make`.

nyurik:
I agree on the first two, and disagree on the last. I don't think quickstart should assume anything -- assumptions often lead to bad code and unexpected magic. I think quickstart should explicitly set PBF_FILE, and thus expect makefile to pass it on to download-osm as --output param, and thus all levels can be explicitly aware of what gets downloaded and later imported. Thus no more "first pbf file found in the dir" magic (we will leave this behavior as part of `import-osm` for now, but won't use it in quickstart)

zstadler:
Agreed.
Actually, the `PBF_FILE` environment variable will also be passed to `download-osm` and no need for `make` to use an `--export` flag.

quickstart.sh
```diff
@@ -7,28 +7,67 @@ set -o nounset
...
     echo "No parameter - set area=$osm_area "
 else
     osm_area=$1
 fi
 
-pbf_file="./data/${osm_area}-latest.osm.pbf"
+if [ $# -eq 2 ]; then
+    osm_server=$2
+else
+  if [ ${osm_area} != $(basename ${osm_area}) ]; then
```
nyurik, line 58: this is incorrect -- https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/download-osm#L443

zstadler: What do you suggest?

nyurik: I think the best way is to remove as much logic as possible from quickstart, and move it to `download-osm`.  For example, we could make `download-osm <anything>` automatically determine the source of the download. `<anything>` could be
* `planet`
* `http://some.url`
* `area name` -- where that name will be searched first in geofabrik catalogue, followed by `osmfr`, followed by bbbike (assuming they also have a catalogue).

Quickstart and makefile should be as dumb as possible, just passing all params to `download-osm`.  That's by the way is the reason why I am not a big fan of the makefile -- instead i think we should have a generic `./run` bash script that executes `openmaptiles-tools` docker image with whatever parameters we give it.  Makefile was designed to build software, not to be a substitute for bash scripts.

zstadler: I'll phrase my question differently...

___What should be the user experience in providing the data source to OpenMapTiles?___

I have no preference as to which component implements the desired user experience.

Here is my take on the question:

- `quickstart.sh` is the primary entry point for most users and it should provide the desired user experience.
- There is significant value in providing a simple user interface for the common use cases.
- Support for uncommon or complex use cases may require higher complexity, and in rare cases would not be provided by `quickstart.sh`.
- Whenever possible, the user should not need to specify the server
  1. The server list for `planet` is not required
  2. The server of a URL is given explicitly
  1. Only and all `bbbike` area names are capitalized: Any capitalized area name will be directed to `bbike`
  2. All `geofabrik` areas have a short (non-hierarchical) name: All lower-case short area names will be directed to `geofabrik`.
  3. All `osmfr` areas except continents require an hierarchical name: All hierarchical area names will be directed to `osmfr`.
  4. User-provided a server name takes precedence: This allows passing hierarchical names to `geofabrik`, passing continent names to `osmfr`, or specifying a particular server for `planet`.
- This user experience should be revisited if and when `osmfr` has a catalog.
- Implementing a [catalog for `bbbike`](https://github.com/openmaptiles/openmaptiles-tools/issues/214) would not make a difference in the user experience because of the capitalized area names.

nyurik: > * Support for uncommon or complex use cases may require higher complexity, and in rare cases would not be provided by `quickstart.sh`.
> * Whenever possible, the user should not need to specify the server
agree to both

>   1. The server list for `planet` is not required
>   2. The server of a URL is given explicitly
>   3. Only and all `bbbike` area names are capitalized: Any capitalized area name will be directed to `bbike`

I disagree that capitalization should be the magical determining factor. Often people will retype either the entire area name or a part of it in the command line, sometimes forgetting the capitalization.  I believe if the user types `quickstart salzburg`, it should "just work" (Salzburg is listed in bbbike).

>   4. All `geofabrik` areas have a short (non-hierarchical) name: All lower-case short area names will be directed to `geofabrik`.

Geofabrik is also hierarchical -- you can use `osm-download geofabrik europe/monaco` and it will work.  Also, some IDs include the slash -- like `us/georgia`.  Which makes it a bit confusing with `georgia` (Europe).

>   5. All `osmfr` areas except continents require an hierarchical name: All hierarchical area names will be directed to `osmfr`.

same thing -- i don't want to put some assumptions into the code like this.  If we do magic for usability sake, lets iterate through all available sources -- until we find what we want, without internal assumptions.  First search through Geofabrik (as the most well known), followed by bbbike (has a catalog), and osmfr. 

>   6. User-provided a server name takes precedence: This allows passing hierarchical names to `geofabrik`, passing continent names to `osmfr`, or specifying a particular server for `planet`.

We don't support limiting planet to a single mirror (you can do it by just passing a url though). And yes, passing `service` name (not `server`) would restrict to just that one service.

> * This user experience should be revisited if and when `osmfr` has a catalog.
ok

> * Implementing a [catalog for `bbbike`](https://github.com/openmaptiles/openmaptiles-tools/issues/214) would not make a difference in the user experience because of the capitalized area names.

see above - i don't think we should be basing `search` on capitalization.  On the other hand, if the search string is an `exact` ID match, than yes, it would take precedence.  So if bbbike and geofabrik have the same name that only differs by capitalization, than we would take the one that matches. But if the user provides something that doesn't have an exact match, we could search them all, and print a warning if there is more than one match.

zstadler: The goal of this PR is to provide easier access to the recently-added download services by modifying the current assumption that all downloads are directed to `geofabrik`.

It adds two simple simple rules that will address the needs most users of all supported services without affecting the vast majority of current user flows with geofabrik.
# 
> > 3. Only and all `bbbike` area names are capitalized: Any capitalized area name will be directed to `bbike`
> 
> I disagree that capitalization should be the magical determining factor. Often people will retype either the entire area name or a part of it in the command line, sometimes forgetting the capitalization. I believe if the user types `quickstart salzburg`, it should "just work" (Salzburg is listed in bbbike).

This PR does not prevent a future, catalog-based bbbike to recognize 
`./quickstart.sh salzburg` as equivalent to `./quickstart.sh Salzburg bbbike`. 
It just recognizes 
`./quickstart.sh Salzburg` as equivalent to `./quickstart.sh Salzburg bbbike`
which will surely be recognized by that future enhancement too.

# 

> > 1. All `geofabrik` areas have a short (non-hierarchical) name: All lower-case short area names will be directed to `geofabrik`.
> 
> Geofabrik is also hierarchical -- you can use `osm-download geofabrik europe/monaco` and it will work. Also, some IDs include the slash -- like `us/georgia`. Which makes it a bit confusing with `georgia` (Europe).

If a user prefers to download monaco from geofabrik rather than osmfr, than both `./quickstart.sh monaco` and `./quickstart.sh europe/monaco geofabrik` and would work with this PR.  I believe the vast majority of users already use the short version, so no changes are required from them.

Without this PR, `./quickstart.sh europe/monaco` and `./quickstart.sh monaco` should abort because they may refer to either `geofabrik` or `osmfr`. Again, preferring geofabrik is an existing assumption that this PR preserves for the short names.

With this PR, a user that wants to download georgia from geofabrik will need to use either `./quickstart.sh europe/georgia geofabrik` or `./quickstart.sh north-america/us/georgia geofabrik`. BTW, `download-osm` already does a good job in detecting the ambiguity:
```
More than one ID 'georgia' was found in Geofabrik, use longer ID:
  * europe/georgia
  * north-america/us/georgia
```

# 

> > 1. All `osmfr` areas except continents require an hierarchical name: All hierarchical area names will be directed to `osmfr`.
> 
> same thing -- i don't want to put some assumptions into the code like this. If we do magic for usability sake, lets iterate through all available sources -- until we find what we want, without internal assumptions. First search through Geofabrik (as the most well known), followed by bbbike (has a catalog), and osmfr. 

You are proposing a different set of assumptions that translate to "both short and hierarchical names will go to geofabrik" - because both will be found in the catalog. I do not think it will provide a better user experience. Hardly anyone uses hierarchical names to access geofabrik. The proposed assumptions will only complicates things for accessing osmfr without a significant added value for accessing geofabrik.

# 

> > * Implementing a [catalog for `bbbike`](https://github.com/openmaptiles/openmaptiles-tools/issues/214) would not make a difference in the user experience because of the capitalized area names.
> 
> see above - i don't think we should be basing `search` on capitalization. On the other hand, if the search string is an `exact` ID match, than yes, it would take precedence. So if bbbike and geofabrik have the same name that only differs by capitalization, than we would take the one that matches. But if the user provides something that doesn't have an exact match, we could search them all, and print a warning if there is more than one match.

It seems that we are in agreement about the end-result for the user. With this PR and after the bbbike catalog is implemented, capitalized names will be going to bbbike.
Implementing a bbbike catalog will allow the flow to flag non-existent capitalized area names as errors before accessing any download service and potentially provide case-insensitive lookup for short names.

# 

This discussion is about the user experience. Once we have a clear understanding of "what" to do, the "how" to do it can include passing-down user-given arguments as-is to `download-osm` and the entire service selection logic, including catalog lookups, can done there.

PR #864, opened 2020-05-13
Update route member tagging support of Trans-Canada Highway

frodrigo:
Fix typo and review the current tagging of the Trans-Canada Highway.


TomPohys:
Thanks @frodrigo, great PR!

PR #866, opened 2020-05-14
Test PR to review v5.2 with Postgres 12

nyurik:
This PR uses tools v5.2 from #903 + Postgres 12


nyurik:
After experimenting with this PR using the [newly documented](https://github.com/openmaptiles/openmaptiles-tools/pull/267) testing process, it seems the `preloaded` image (water polygons) are broken everywhere on zoom 5.  When importing the same data using `make import-data`, the issues goes away. The import can be done even on top of the existing preloaded image and it still fixes that.

PR #870, opened 2020-05-15
Simplify update_gbr_route_members()

frodrigo:
Code cleaning.

The function `update_gbr_route_members()` use two sub queries, but only one is relay needed.

### DIFF
--- a/layers/transportation_name/update_route_member.sql
+++ b/layers/transportation_name/update_route_member.sql
@@ -5,22 +5,18 @@ DROP TRIGGER IF EXISTS trigger_flag_transportation_name ON osm_route_member;
 CREATE OR REPLACE FUNCTION update_gbr_route_members() RETURNS VOID AS $$
 DECLARE gbr_geom geometry;
 BEGIN
-  select st_buffer(geometry, 10000) into gbr_geom from ne_10m_admin_0_countries where iso_a2 = 'GB';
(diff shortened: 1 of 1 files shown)


layers/transportation_name/update_route_member.sql
```diff
@@ -5,22 +5,18 @@ DROP TRIGGER IF EXISTS trigger_flag_transportation_name ON osm_route_member;
...
-    ) UNION (
-      SELECT 0, hw.osm_id, substring(hw.ref from E'^[AM][0-9AM()]+'), 'omt-gb-trunk'
-      from osm_highway_linestring hw
-      where length(hw.ref)>0 and ST_Intersects(hw.geometry, gbr_geom)
-        and hw.highway IN ('trunk')
-    )
-  ;
+    INSERT INTO osm_route_member (osm_id, member, ref, network)
+    SELECT 0, osm_id, substring(ref FROM E'^[AM][0-9AM()]+'),
+        CASE WHEN highway = '' THEN 'omt-gb-motorway' ELSE 'omt-gb-trunk' END
```
nyurik, line 13: I think you meant `WHEN highway = 'motorway'` ?

frodrigo: Yes, good catch.
Fixed.

PR #887, opened 2020-05-22
Allow custom ports for postserve/tileserver

nyurik:
* `make start-postserve PPORT=8000`
* `make start-tileserver TPORT=9000`

Note that both ports need to be the same on host and inside the docker container because the services include a URL to the tiles as part of their manifest.


frodrigo:
Note, in my setup I also need to add the `--serve` in the `docker-compose.yaml` file, because I have a nginx reverse proxy on top of it. (Yes I know, postserve it is not for production)

I have not make a PR, because I am not sure of the status of this.

PR #893, opened 2020-05-25
Use new mbtiles-tools to set mbtiles metadata

### LINKED ISSUE #401: In TileJSON (and mbtiles metadata) add content/values to description field
In TileJSON (and mbtiles metadata) the content/values of _"description"_ field in object vector_tiles is currently an empty string. See e.g. https://free.tilehosting.com/data/v3.json?key={your_api_key} and look at landuse:

```
  vector_layers:
    1: 
    ...
    3:
      maxzoom:	14
      fields:	
        class:	"String"
      minzoom:	0
      id:	"landuse"
      description:	""
```

But "description" should explain the layer, e.g. "Landuse polygons for parks, university, sports and other areas. Geometry types: polygon, line, point, label_placement." 

See e.g. this dialog from the QGIS Vector Tiles reader plugin (https://github.com/geometalab/Vector-Tiles-Reader-QGIS-Plugin/blob/master/README.md ):
grafik

nyurik:
mbtiles-tools uses openmaptiles.yaml file and Postgres connection
to correctly configure generated .mbtiles metadata.

The tool will also reset min/max zoom of the file depending
on the generated zooms, and copy layer descriptions.

Closes #401

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ generate-tiles: all start-db
 	echo "Generating tiles ..."; \
 	$(DOCKER_COMPOSE) $(DC_CONFIG_TILES) run $(DC_OPTS) generate-vectortiles
 	@echo "Updating generated tile metadata ..."
-	$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools generate-metadata ./data/tiles.mbtiles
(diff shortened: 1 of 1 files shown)


nyurik:
Merged into #903

PR #896, opened 2020-05-26
Make update_osm_peak_point use incremental update #814

frodrigo:
Replacing update on the whole table with an update only on changed rows.

The goal is to update more quickly by just updating the changing content.
The update now focus on `osm_id` of changed rows, it use index. Add a where clause `tags != update_tags(tags, geometry)` en ensure only update when changed.

It now requires only one trigger in place of two.

The `UPDATE` is keep in a function to be reutilisable for initial setup and trigger update, the switch is done with `(new_osm_id IS NULL OR osm_id = new_osm_id)`.

It is a prof of concept. If it is ok, I will do batch of these on other similar updates.

It addresses #814.

Note: `update_tags` is already IMMUTABLE but can be switched to pure SQL and parallel safe https://github.com/openmaptiles/openmaptiles-tools/blob/master/sql/zzz_language.sql#L154

### DIFF
--- a/layers/mountain_peak/update_peak_point.sql
+++ b/layers/mountain_peak/update_peak_point.sql
@@ -1,57 +1,33 @@
-DROP TRIGGER IF EXISTS trigger_flag ON osm_peak_point;
-DROP TRIGGER IF EXISTS trigger_refresh ON mountain_peak_point.updates;
+DROP TRIGGER IF EXISTS trigger_update_point ON osm_peak_point;
 
 -- etldoc:  osm_peak_point ->  osm_peak_point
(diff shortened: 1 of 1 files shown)


nyurik:
I propose we solve #899 first -- this way we can test if these types of changes are ok to merge automatically.  Anyone wants to tackle that?

nyurik:
@frodrigo hi, i merged the conflict with your patch, but please double check the logic

frodrigo:
Thank, I will make the similar others when I will have more free time. Probably in July.

TomPohys:
Awesome, thanks.

PR #907, opened 2020-05-31
Minor makefile/quickstart cleanup

nyurik:
* Make a few spacing adjustments for ease-of-reading and consistency
* fix bbbike naming and other source urls
* remove unneeded `override`
* added `list-bbbike` target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,41 +1,45 @@
+#
+# First section - common variable initialization
+#
+
 # Ensure that errors don't hide inside pipes
 SHELL         = /bin/bash
 .SHELLFLAGS   = -o pipefail -c
 
 # Options to run with docker and docker-compose - ensure the container is destroyed on exit
 # Containers run as the current user rather than root (so that created files are not root-owned)
-DC_OPTS?=--rm -u $(shell id -u):$(shell id -g)
+DC_OPTS ?= --rm -u $(shell id -u):$(shell id -g)
 
 # If set to a non-empty value, will use postgis-preloaded instead of postgis docker image
-USE_PRELOADED_IMAGE?=
+USE_PRELOADED_IMAGE ?=
 
 # If set, this file will be imported in the import-osm target
 PBF_FILE?=
 
 # Local port to use with postserve
-PPORT?=8090
+PPORT ?= 8090
 export PPORT
 # Local port to use with tileserver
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -119,7 +128,8 @@ clean:
 	rm -rf build
 
 .PHONY: destroy-db
-destroy-db: override DC_PROJECT:=$(shell echo $(DC_PROJECT) | tr A-Z a-z)
+# TODO:  Use https://stackoverflow.com/a/27852388/177275
```
zstadler, line 131: IMHO, including the [GNU Make Standard Library](https://sourceforge.net/projects/gmsl/) in the openmaptiles distribution just for the conversion of the `$(CD_PROJECT)` string to lowercase is not necessarily an improvement.

BTW, the library uses the following code to implement `tr` using Make's `foreach` loop with `eval` and a couple of `subst``:
```make
# ----------------------------------------------------------------------------
# Function:  tr
# Arguments: 1: The list of characters to translate from 
#            2: The list of characters to translate to
#            3: The text to translate
# Returns:   Returns the text after translating characters
# ----------------------------------------------------------------------------
tr = $(strip $(__gmsl_tr3)$(call assert_no_dollar,$0,$1$2$3)              \
     $(eval __gmsl_t := $3)                                               \
     $(foreach c,                                                         \
         $(join $(addsuffix :,$1),$2),                                    \
         $(eval __gmsl_t :=                                               \
             $(subst $(word 1,$(subst :, ,$c)),$(word 2,$(subst :, ,$c)), \
                 $(__gmsl_t))))$(__gmsl_t))

# Common character classes for use with the tr function.  Each of
# these is actually a variable declaration and must be wrapped with
# $() or ${} to be used.

[A-Z] := A B C D E F G H I J K L M N O P Q R S T U V W X Y Z #
[a-z] := a b c d e f g h i j k l m n o p q r s t u v w x y z #
[0-9] := 0 1 2 3 4 5 6 7 8 9 #
[A-F] := A B C D E F #
```

nyurik: i think you are right. This all depends on what we would like as the "minimum requirements" for the server. In an ideal world, nothing except the docker service should be needed (working equally well for Linux/Mac/Win), but in reality we may have to jump through a few hoops before we get there.

PR #910, opened 2020-06-02
Removed name_de from mapping

### LINKED ISSUE #769: Is 'name_de' not a waste of space ?
I was wondering if this tag is relevant. I like the idea of having name_en available as English is the de facto OSM and international language, but, besides the fact that the original creators were german-speaking (kinda ;-), it seems like a waste of space as name:de is already available and coalescing can be made more easily at the SDK level (with mapbox sdk at least).

Danke für ihr Verständnis ;-)

kukulich:
Fixes #769


nyurik:
I re-expanded on this idea in a new PR #928 - it also removes `name_en`, and does a few related changes (updates code that handles `update_osm_city_point()`.  I think we should close this PR in favor of the other one?

kukulich:
Yes :) I made the PR just because the commit already exists in my fork :)

Issue #776, opened 2020-01-29
Add area-base filtering for park names in lower zoom levels


zstadler:
Park names are available as point features for all parks in a given zoom level. In lower zoom levels, some parks are completely hidden by their own labels and park labels can extend far beyond their boundaries.

This problem has already been handled in the `water_name` layer, using zoom-dependent filtering:

https://github.com/openmaptiles/openmaptiles/blob/9f4d48cf25fec4dfa80e3abb67801edd6607745f/layers/water_name/layer.sql#L38

I'll be glad to prepare a PR that use the `area > 70000*2^(20-zoom_level)` expression for filtering park labels in zoom levels other than 14.

zstadler:
Please let me know if you like me to create a PR based on https://github.com/IsraelHikingMap/openmaptiles/commit/1530dc551a0bd38a11e51ed59354917e294423f6.

It would be best to apply it after https://github.com/openmaptiles/openmaptiles/pull/773 is merged.

TomPohys:
Hello @zstadler, thank you for this update. 

Your PR is very welcome and can be part of v3.12 release.

zstadler:
Hello @TomPohys,

I'll be happy to prepare it after https://github.com/openmaptiles/openmaptiles/pull/773 is closed becuase of its dependency.

klokan:
Hi @zstadler - do you think you could manage to push this change till feature freeze of 3.12 (Friday?)

It could be a great contribution for the next release.

zstadler:
@klokan per your request, I've prepared a PR.

TomPohys:
@zstadler, thanks a lot. It going to be part of 3.12 release.

klokan:
Thank you very much @zstadler - cool!

PR #911, opened 2020-06-03
Add area-base filtering for park names in lower zoom levels

zstadler:
Resolve https://github.com/openmaptiles/openmaptiles/issues/776


zstadler:
It was shamelessly taken from
https://github.com/openmaptiles/openmaptiles/blob/9f4d48cf25fec4dfa80e3abb67801edd6607745f/layers/water_name/layer.sql#L38

nyurik:
> It was shamelessly taken from
> https://github.com/openmaptiles/openmaptiles/blob/9f4d48cf25fec4dfa80e3abb67801edd6607745f/layers/water_name/layer.sql#L38

yes, i saw it, i was referring to that line as well - we shouldn't be using magic computations without any explanation and without anyone actually knowing what it does :). Would have been great if we either used some known functions or at least document it well. But i'm not blocking the merge, just raising a TODO for later.

PR #913, opened 2020-06-03
Building aggregation at zoom 13

TomPohys:
Creating building blocks by aggregation of small buildings at zoom 13. 

The main advantage is:
- much, much nicer zoom level 13 at urban areas

The possible drawbacks are:
- rise size (for the Czech Republic it was 1%)
- a significant rise of `import-sql` time


TomPohys:
In advance of #936

PR #917, opened 2020-06-06
NOOP: Format all layer's SQL code

nyurik:
I would like to reformat all of our SQL to have a consise coding style.
This makes it far easier to understand the code for a casual contributor,
and lets us spot errors more easily.

Most importantly, it makes it much easier to grep (search) the code because it is more likely to be in the same syntax

Some key changes:
* SQL keywords are always UPPERCASE, e.g. `SELECT WHEN AS END ...`
* types, variables, aliases, and field names (identifiers) are always lower case
* `LANGUAGE 'plpgsql'` is now `LANGUAGE plpgsql` (no quotes)
* a few minor spacing/semicolon cleanups


TomPohys:
Hi @nyurik, thanks for this big clean up. 

Based on [PostgreSQL Docs](https://www.postgresql.org/docs/9.6/sql-createfunction.html) should be [RETURNS TABLE](https://github.com/openmaptiles/openmaptiles/pull/917/files#diff-bec695f5c5f648d663f03bda324558ceR6) in upper case. 

and `LANGUAGE plpgsql;` vs. `LANGUAGE SQL;`

Based on [PostgreSQL Docs](https://www.postgresql.org/docs/9.6/sql-createtrigger.html) should be [NEW and OLD](https://github.com/openmaptiles/openmaptiles/pull/917/files#diff-c0284b5d204184e44b839bf40d3bdf74R98) in upper case.

Could you please add united format to these?

Thanks

nyurik:
@TomPohys i could adjust those, but please note that PostgreSQL docs are not consistent here:  `table` is the TYPE, so if documentation states `CREATE FUNCTION foo(int, int default 42)` (types like `int` are in the lower case), `table` in theory should also be lower case. Anyway, I will adjust.

frodrigo:
Which tool did you use to format the code?

I find the align of `AS` counterproductive when diff code, all fields may be in diff just to change the align of the AS.

I also dislike some other part of the formatting, but It is more personal.

nyurik:
I used inteli j idea, but I'm ok with any tool as long as we use it consistently. The old code was a bit of a mess stylistically and was hard to read.

frodrigo:
Is it a one shoot reformatting? If so, can I make a PR to remove the extra space before the AS ?

frodrigo:
I finally got it, It is in the non free intellij idea version.

Could be nice to share yours setting.

PR #918, opened 2020-06-06
Fix & optimize incorrect function declarations

nyurik:
* All functions that access database must be declared as `STABLE`, not `IMMUTABLE` -- because database can change at any moment, e.g. during an update
* tagged a few functions as `STRICT` -- passing `NULL` as a parameter will always result in a `NULL
* tagged one function as parallel safe

NOTE: somehow `ST_AsMVT()` method of tile generation is showing 70-90% slowdown with this patch. I am not sure of why this is happening. If the reason is the `IMMUTABLE` -> `STABLE` change, we may have to dig deeper into PG optimization

### DIFF
--- a/.env
+++ b/.env
@@ -38,3 +38,6 @@ IMPOSM_CONFIG_FILE=/usr/src/app/config/repl_config.json
 BORDERS_CLEANUP_FILE=data/borders/cleanup.pbf
 BORDERS_PBF_FILE=data/borders/filtered.pbf
 BORDERS_CSV_FILE=data/borders/lines.csv
+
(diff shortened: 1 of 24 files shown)


TomPohys:
Hi @nyurik, thanks for optimizations. How these changes affect tile generation through mapnik?

nyurik:
I do not know -- most of these are not optimizations, they are bug fixes -- it was a mistake to use `immutable` -- if you run update, the DB could produce incorrect results.

TomPohys:
Hi @nyurik, do you still testing the performance? Should we mark this PR as Draft?

PR #921, opened 2020-06-06
Fix `clean-unnecessary-docker` make target

zstadler:
Add missing `-q` flag


zstadler:
Please note the margins of error in the above evaluation.
AFAIK, the changed line was not activated by any of the tests.

PR #928, opened 2020-06-09
[BREAKING] Remove name_en and name_de

### LINKED ISSUE #796: Strategy for dependent layer's code
We currently have some hidden cross-layer dependencies that cannot be easily tracked.

For example, `waterway_brunnel` function is used in [waterway](https://github.com/openmaptiles/openmaptiles/blob/master/layers/waterway/waterway.sql#L1) and [water](https://github.com/openmaptiles/openmaptiles/blob/master/layers/water/water.sql#L56) layers. At the moment the declaration is identical, but this only works if functions are imported sequentially, and we don't accidentally modify just one of them and forgetting the other.

Another example is `highway_class` function, declared in `transportation` layer, but used in both `transportation` and `transportation_name` layers.

One possible solution is to have dependency tracking: each layer can define what other layers it depends on. So `transportation_name` would list `transportation` as a dependency.

```yaml
layer:
  id: "transportation_name"
  requires: "transportation"   # this can be a string or a list of strings
```

nyurik:
Closes #796 and includes some of the same code as in #910 (rebased) by @kukulich

This PR removes both `name_en` and `name_de` fields from all layers.

This PR requires some changes to the styling as well:
current styles use `name_en` and `name_de` instead of `name:en` and `name:de`.
All other languages are already done in the new system.

TODO: decide how to handle `name_int` and `name_latin`.


layers/transportation_name/update_transportation_name.sql
```diff
@@ -57,12 +55,8 @@ SELECT (ST_Dump(geometry)).geom                AS geometry,
...
          SELECT ST_LineMerge(ST_Collect(geometry)) AS geometry,
                 name,
-                name_en,
-                name_de,
-                hstore(string_agg(nullif(slice_language_tags(tags ||
-                                                             hstore(ARRAY ['name', name, 'name:en', name_en, 'name:de', name_de]))::text,
-                                         ''), ','))
-                                                   AS "tags",
+                hstore(string_agg(nullif(slice_language_tags(tags || hstore(ARRAY ['name', name]))::text, ''),
+                                  ',')) AS "tags",
```
zstadler, line 59:
```suggestion
                hstore(string_agg(nullif(slice_language_tags(tags)::text, ''), ',')) AS "tags",
```

I can see why `name_en` and `name_de` were originally added to `tags` - their value **was** computed earlier using `name`, `name:en` and `name:de`:
https://github.com/openmaptiles/openmaptiles/blob/fb7c1ef42b4fdb21f98466274b8e4636047739cb/layers/transportation_name/layer.sql#L27-L28

I fail to see the purpose of adding `name` to `tags` - `name` is included in the result as a separate column and it is not used by the `{name_languages}` expression:
https://github.com/openmaptiles/openmaptiles/blob/9bcc51919d3f6167e92cd10f696e1cecfec81c0b/layers/transportation_name/transportation_name.yaml#L87

ZeLonewolf:
@TomPohys are we able to proceed with breaking changes like this targeting 4.0 or do we still expect 3.x development to continue on mainline?

PR #929, opened 2020-06-10
Change GitHub Actions to dedicated server

TomPohys:
Change Github Actions to the dedicated server name "ralph"

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
 
   integrity_test:
     name: Run integrity test
-    runs-on: ubuntu-latest
+    runs-on: ralph
     steps:
 
       - name: Checkout the changes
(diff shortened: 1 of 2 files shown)


nyurik:
Also, note that the test did not complete -- so the custom server must still be down

daliborjanak:
It is, we are working on it. It will be solved today

.github/workflows/tests.yml
```diff
@@ -23,6 +23,9 @@ jobs:
           MIN_ZOOM: 0
           MAX_ZOOM: 14
         run: |
+          # For now, change the quickstart values directly in the .env file
```
nyurik, line 26: why is this being added?

TomPohys: MAX_ZOOM: 14 was not taken. Please look into old integrity test on monaco-dc_config.yaml

PR #932, opened 2020-06-12
Remove alignment of AS in SQL and few others

frodrigo:
Remove alignment of `AS` in `SELECT` and type in declaration in order to keep it more maintainable manually.

- Keep the AS close, it is easier to read, far `AS` on the right is relay not obvious.
- Make the diff smaller on edge case as no need to realign all the code block.


frodrigo:
Rebase/redo commit.

But the CI fails inconsistently.

frodrigo:
Thank

PR #944, opened 2020-06-26
Make more simple incremental update

frodrigo:
Replacing update on the whole table with an update only on changed rows.

The goal is to update more quickly by just updating the changing content.
The update now focus on osm_id of changed rows, it use index. Add a where clause tags != update_tags(tags, geometry) en ensure only update when changed.

It requires one more trigger and a table to store changed osm_id.

The UPDATE is keep in a function to be reusable for initial setup and trigger update.

I try many code layout before done it in this way with the goal to keep the code for initial pass and for update. It should have low impact on initial data load. Better performance for row update can be achieve with BEFORE UPDATE, but require to duplicate the logic.

It is not based on the already merged https://github.com/openmaptiles/openmaptiles/pull/896 because calling and update within a function for each updated row was not efficient for larger table (like housenumber).

It addresses #814.

### DIFF
--- a/layers/aerodrome_label/update_aerodrome_label_point.sql
+++ b/layers/aerodrome_label/update_aerodrome_label_point.sql
@@ -1,25 +1,44 @@
 DROP TRIGGER IF EXISTS trigger_flag ON osm_aerodrome_label_point;
+DROP TRIGGER IF EXISTS trigger_store ON osm_aerodrome_label_point;
 DROP TRIGGER IF EXISTS trigger_refresh ON aerodrome_label.updates;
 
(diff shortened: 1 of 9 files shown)


frodrigo:
Benchmark on Wales extract. Using same extract date and same update range (20 daily updates).
Size: 77 MB

Master
* Import time 932 s
* AVG update time : 191.7 s (std dev 14.8)
* Median update time : 188.4 s

This PR
* Import time 913 s (-1.9%), no special reason to be faster
* AVG update time : 175.6 s (std dev 13.9)
* Median update time : 172.8 s

Update are 8.3% faster.

PR #950, opened 2020-07-22
Fix More fail safe incremental update on water_lakeline and water_point (#949)

frodrigo:
Sorry my PR from yesterday was buggy. It seems obvious and not fully tested.
One of the change need to be fixed.


TomPohys:
No problem. The test was passed and I was not expecting any problem. Thanks for fix.

PR #972, opened 2020-08-30
Expose COPY_CONCURRENCY to users

bravetheheat:
Allow users to override how the number of threads used when generating map vector tiles.


TomPohys (approved):

TomPohys:
Thanks, @bravetheheat.

Issue #964, opened 2020-08-18
Generating disputed_name for NE data source of borders


eva-j:
For now `NULL AS disputed_name, NULL AS claimed_by` is defined for all the NE inputs of boundary layer.
https://github.com/openmaptiles/openmaptiles/blob/master/layers/boundary/boundary.sql#L221

I suggest to use `ogc_fid` to make `disputed_name` filled (something like: `"ne110m_" + ogc_fid`, `"ne50m_" + ogc_fid`, `"ne10m_" + ogc_fid`), then it is possible to filter in style. There is no better info to distinguish the lines.

![Screenshot from 2020-08-18 16-54-40](https://user-images.githubusercontent.com/8367628/90529468-0c7a0280-e174-11ea-9967-1e666b76109a.png)

PR #976, opened 2020-09-02
Adding disputed_name to NE sources

eva-j:
`disputed_name` is filled for disputed lines of NaturalEarth sources - it contains `ogc_fid` as suggested in https://github.com/openmaptiles/openmaptiles/issues/964. This way it is possible to filter disputed lines in style.


TomPohys (approved):

TomPohys:
Thanks @eva-j.

PR #983, opened 2020-09-12
#766 remove housenumbers inside polygon and same number

### LINKED ISSUE #766: Housenumber duplicates
E.g. in Kaunas, Lithuania (18/54.89617/23.88801)
house_no_labels_dublicates

Housenumbers are imported as points and polygons:
Screenshot from 2020-01-13 18-06-44

Then polygons are transformed into points, so in the end we see e.g. 3 times number 8.

We should probably first remove points laying on polygon and then make Centroid/PointOnSurface of the polygon (only the second part is done in https://github.com/openmaptiles/openmaptiles/blob/master/layers/housenumber/housenumber_centroid.sql#L8-L13).

smellman:
PR for fix #766 .
This PR may resolve https://github.com/openmaptiles/openmaptiles/issues/766#issue-549067569 comment.
But this PR doesn't resolve https://github.com/openmaptiles/openmaptiles/issues/766#issuecomment-642719388 comment because it seems contains data bugs.
Klokantech_Basic_-_TileServer_GL_と_Housenumber_duplicates_·_Issue__766_·_openmaptiles_openmaptiles


eva-j (approved):

eva-j:
Thanks a lot, @smellman! Data bugs need to be corrected in OSM, it is not OMT problem.

frodrigo:
This delete query does not support the diff update, it scan the full table.
It should be support the ://github.com/openmaptiles/openmaptiles/pull/983/files#diff-e9764212e27693c39a6704d190df1b97L22

layers/housenumber/housenumber_centroid.sql
```diff
@@ -12,6 +12,19 @@ CREATE TABLE IF NOT EXISTS housenumber.osm_ids
 -- etldoc: osm_housenumber_point -> osm_housenumber_point
 CREATE OR REPLACE FUNCTION convert_housenumber_point(full_update boolean) RETURNS void AS
 $$
+    -- Delete housenumber duplicates
+    DELETE FROM osm_housenumber_point
+    WHERE osm_id IN (
+      SELECT pt.osm_id
+      FROM osm_housenumber_point pt
+      INNER JOIN osm_housenumber_point poly
+      ON (ST_GeometryType(poly.geometry) = 'ST_Polygon'
```
frodrigo, line 21: Duplicate `addr:housenumber` can also be point.

eva-j:
Thanks a lot, @frodrigo, I've just made following PR.

layers/housenumber/housenumber_centroid.sql
eva-j: Yes, this can happen. However point house numbers are mostly imported from point POI and I expect they are not that often on the same place. I checked it on Lithuania and there was no duplicate point after data import.

PR #992, opened 2020-09-20
openmaptiles-tools: consider MIN_ZOOM/MAX_ZOOM env

zstadler:
Allow MIN_ZOOM, and MAX_ZOOM to be overwritten from shell for `quickstart.sh`, `make generate-dc-config`, and other `docker-compose openmaptiles-tools` uses.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,8 +34,10 @@ services:
       # Must match the version of this file (first line)
       # download-osm will use it when generating a composer file
       MAKE_DC_VERSION: "2.3"
-      # Allow DIFF_MODE to be overwritten from shell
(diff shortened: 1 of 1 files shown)


zstadler:
For example,
```
export MAX_ZOOM=14; ./quickstart.sh monaco
```
ignores the `MAX_ZOOM=14` envionment variable and used the `MAX_ZOOM=7` value from `.env` file

zstadler:
This PR will also make this code unecessary
https://github.com/openmaptiles/openmaptiles/blob/5c640daf4f5be0559d4c7b3824ebc09abaf42ef9/.github/workflows/tests.yml#L26-L28

TomPohys (commented):
Approved

PR #987, opened 2020-09-15
Replace materialized view osm_transportation_name_network by tables with diff update

frodrigo:
Replacing materialized view by a tables with update from trigger on change only.

The osm_id of object changing on table osm_route_member and osm_highway_linestring are logged.
At the end of the transaction the dependent row of osm_transportation_name_network are updated.

The goal is to update more quickly the content of derivated tables by just updating the changing content. It replaces the update of materialized view because their need a full recompute (with lock issue).

Note, it is only a part of the materialized view of the transportation layer, I am also working on the other parts, other parts are in #892 (omaterialized view on sm_transportation_name_linestring and osm_transportation_name_linestring_genX).

It addresses #814 and a part of #809.

PR #996, opened 2020-09-21
Diff update osm transportation name network linestring

frodrigo:
Replacing materialized view by a tables with update from trigger on change only.

Differential update of `osm_transportation_name_linestring`, `osm_transportation_name_linestring` and `osm_transportation_name_linestring_genX`).

At the end of the transaction the dependent rows are updated.

The goal is to update more quickly the content of derivated tables by just updating the changing content. It replaces the update of materialized view because their need a full recompute (with lock issue).

It is the last part of the replacement of materialized view for the transportation layer.

It addresses #814 and a part of #809.


layers/transportation_name/network_type.sql
```diff
@@ -3,10 +3,9 @@ DROP TRIGGER IF EXISTS trigger_store_transportation_highway_linestring ON osm_hi
 DROP TRIGGER IF EXISTS trigger_flag_transportation_name ON transportation_name.network_changes;
 DROP TRIGGER IF EXISTS trigger_refresh_network ON transportation_name.updates_network;
 
-DROP MATERIALIZED VIEW IF EXISTS osm_transportation_name_linestring_gen1 CASCADE;
-DROP MATERIALIZED VIEW IF EXISTS osm_transportation_name_linestring_gen2 CASCADE;
-DROP MATERIALIZED VIEW IF EXISTS osm_transportation_name_linestring_gen3 CASCADE;
-DROP MATERIALIZED VIEW IF EXISTS osm_transportation_name_linestring_gen4 CASCADE;
```
TomPohys, line 9: These DROPs, aren't they supposed to be in `update_transportation_name.sql` where the materialized views are created? In repeated `make import-sql` steps are rising error. 

It is the same as for #987 [`DROP MATERIALIZED VIEW IF EXISTS osm_transportation_name_linestring CASCADE;`](https://github.com/openmaptiles/openmaptiles/pull/987/files#diff-1ed2951173082c9f9c3b1ed896dcd768L2) which supposed to be in `update_transportation_name.sql` and not in `network_type.sql` and what is now rising error for repeated `make import-sql`.

frodrigo: Here I remove the `DROP MATERIALIZED`, but replace it by `DROP TRIGGER`.
I agree this is odd. I was like this before I start working on it.
It is due because interdependencies between `network_type.sql` and `update_route_member.sql`. The two may probably need to be merged. But please, if you want to merge it, let do it after this PR.

TomPohys: Thanks, I see it now.

TomPohys (approved):

TomPohys:
Thanks a lot. It looks great.

zstadler:
@frodrigo,
I may be missing something, but I do not see any references to changes the `tags` field. 

Does this code handle updates that only change the `tags` fiels, such as an update to a `name:<language>` tag for a language other than `en` or `de`?

PR #997, opened 2020-09-22
Better update of osm_important_waterway_linestring, use frist and last version of osm object

frodrigo:
Current implementation of osm_important_waterway_linestring have two bugs:
* The distinct on is_old keep the oldest version of the old object, and the oldest version of the new object, but need the last version of the new object.
* Delete the old version of the object and rebuild the using the new version of the object. But we need to remove matchings the old and the new version of the object, then rebuild the two.

Then only delete and update using first and last version of the object, intermediate versions are ignored.

Similar implementation of what is done in #996.


TomPohys (approved):

TomPohys:
Thanks, looks great.

PR #1000, opened 2020-09-24
Add landuse=salt_pond as water

frodrigo:
Add salt pond as water.

Most of salt pond are already tagged with natural=water even if it's not required. So the load added by this tag is very light.


TomPohys:
Thanks @frodrigo.

PR #1002, opened 2020-09-25
Remove large unused index osm_highway_linestring_highway_idx

frodrigo:
The index `osm_highway_linestring_highway_idx` is not used. There is also a used partial index `osm_highway_linestring_highway_partial_idx`.

I cross check the usage of the index from postgres query stats and all SQL query code where the table `osm_highway_linestring` is refered.

This index is relatively big.


TomPohys (approved):

TomPohys:
Hi @frodrigo, if the index `osm_highway_linestring_highway_idx`on attribute `highway` is not used at all, then it is really useless to have it. Could you please update the branch?

frodrigo:
Updated

Issue #970, opened 2020-08-27
Implement Contrib layer


frodrigo:
I am thinking about how to implement contrib layer in OpenMapTiles. Do you have ideas about this ? Contrib layers should be in OMT github repo or not ? How to activate the use of a contrib layer ?

I already open a PR as first step to make tileset definition configurable. https://github.com/openmaptiles/openmaptiles/pull/968
It helps me run an alternative tileset definition from outside the OMT project with some more hack.

nyurik:
I'm all for storing alternative layers in this repo, and yes, #968 is the good first step.

Things to decide/confirm.
* directory structure -- I think we should keep those layers all together, and only differentiate based on what is listed in the tileset definition file.
* tileset definition files should probably be stored together as well, with the `openmaptiles.yml` being the one we advertise as the default.
* testing -- we would need to run quickstart for all tileset definitions in parallel so that tests don't take too long. There should not be any tileset files that are not being tested automatically.

jsanz:
I'd prefer to have a system were new layers can be added, independently of being a contribution to this repo or not.

Something like allowing to use `git submodules` to add new layers to a `layers/contrib` folder in this repo and then when building, attach to the processing pipeline any new layers, if that makes sense.

frodrigo:
I also have a preference for out standing layers, more agile and manged by other people, to not report review and merge PR to maintainers of OpenMapTiles repo. In the other hand, some simple and frequent asked extra layers could be there.

Extra layers on OMT will be easier to test.

On OMT project I think to sibling directory of `layers`, like `layers_contrib` or `layers_extra`. Providing `openmaptiles.yml` for contrib layers may be undecidable, as the contrib layer my be combined or not. We may add a `openmaptiles-contribs.yml` to test all layers at once.

About contrib layers from external project. A project may come with one or more layers.

frodrigo:
Start a PR for request for comments #1003

PR #1003, opened 2020-09-25
Support for addons layers

frodrigo:
The main goal of this PR is to discuss on the right to achieve contrib/extra/addons layers #970.

First, I try many vocabulary and in the end I (personally) prefer "addons" layers.

* [x] Add `layers_addons` directory with a first layer. For this I revival an old rejected PR I made : a layer for trees #554
* [ ] Fix openstreetmap-tools to support layers in multiple directories https://github.com/openmaptiles/openmaptiles-tools/issues/320
* [ ] Doc
* [ ] Test, for enbded addons layers.

In the mean time I work on external addons layers with this toy projects : https://github.com/makina-maps/openmaptiles-contrib-bicycle_hiking_route / https://github.com/makina-maps/osm-bright-bicycle-gl-style

The buildings cluster PR #725 might be provided as addons as replacement or complementary layer.


ZeLonewolf:
One issue that I've encountered when exploring this topic, is that if you have a layer that extends another layer (for example, `transportation_name` extends `transportation`, there is no obvious way to render only the extended layer in the mbtiles, but not the base layer.  So if I wanted to create `transportation_optimized` that used all of `transportation`'s tables but created a smaller layer with just the data I care about, there's no way in the current scripts to generate a tiles.mbtiles that has only `transportation_optimized` but not `transportation`.  That seems like a problem we would want to solve in order to enable customizations of the base layers.

PR #1006, opened 2020-09-28
Add timers to trigger refresh function

frodrigo:
Add timer to all trigger refresh functions to help point where the time is spend on update.


TomPohys (approved):

TomPohys:
Thanks @frodrigo.

PR #1007, opened 2020-09-28
Add protocol to postserve --serve paramter

frodrigo:
The current option of postserve in the docker-compose produce an invalid tilejson.
```json
"tiles": ["localhost:8090/tiles/{z}/{x}/{y}.pbf"]
```
`tiles` should be URL. The protocol is missing.

Not sure about the best way to fix this. Just adding `http://` to the docker-compose.yaml does not allow https usage. But setting protocol in `OMT_HOST` seems weird.

Could add an extra OMT_PROTOCOL with `http` as default.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -90,7 +90,7 @@ services:
 
   postserve:
     image: "openmaptiles/openmaptiles-tools:${TOOLS_VERSION}"
-    command: "postserve ${TILESET_FILE} --verbose --serve=${OMT_HOST:-localhost}:${PPORT:-8090}"
+    command: "postserve ${TILESET_FILE} --verbose --serve=${OMT_HOST:-http://localhost}:${PPORT:-8090}"
     env_file: .env
     environment:
       TILESET_FILE: ${TILESET_FILE}


nyurik:
`PPROTOCOL` to match `PPORT` ?  Not sure about the best course of action either.

TomPohys:
Hi @frodrigo, after I start postserve I see a protocol in TileJSON - `"tiles": ["http://localhost:8090/tiles/{z}/{x}/{y}.pbf"]`

image

frodrigo:
I look more closely.

When running from makefile it have the `http://` part in the `OMT_HOST` from https://github.com/openmaptiles/openmaptiles/blob/master/Makefile#L58

I misrepresented by the name of the var when running without the makefile.

TomPohys:
with your change the `make start-postserve` creates `"tiles": ["http://http://localhost:8090/tiles/{z}/{x}/{y}.pbf"]`

frodrigo:
Fixed to have `http://` part in the default value.

The variable is still loosely named, but it will be consistent at least.

TomPohys (approved):

TomPohys:
Hi @frodrigo, thanks for PR, could you please update the branch?

frodrigo:
> Hi @frodrigo, thanks for PR, could you please update the branch?

Done

Issue #1008, opened 2020-09-29
Update makefile for hierarchical area names


zstadler:
Update `make` to detect existing `osm.pbf` files of hierarchical area names like `europe/monaco`.

For example:
```
$ make download area=europe/monaco
```
Followed by
```
$ ls data/europe/monaco.osm.pbf
data/europe/monaco.osm.pbf
$ make import-osm
Starting postgres docker compose target using default image (no recreate if exists)
Creating openmaptiles_postgres_1 ... done

ERROR: The 'area' parameter or environment variable have not been set, and there is no data/*.osm.pbf file

  make import-osm area=<area-id>

To download an area, use   make download <area-id>
To list downloadable areas, use   make list-geofabrik   and/or   make list-bbbike
Makefile:343: recipe for target 'import-osm' failed
make: *** [import-osm] Error 1
```

PR #1009, opened 2020-09-29
Update makefile for hierarchical area names

zstadler:
- Use `find` instead of `wildcard` to detect existing `.osm.pbf` files
- Use `patsubst` to extract an `area` from a filename
- Clarify messssage, especially for errors

Resolve https://github.com/openmaptiles/openmaptiles/issues/1008

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,38 +67,64 @@ endef
 
 #
 # Determine area to work on
-# If $(area) parameter is not set and data/*.osm.pbf finds only one file, use it as $(area).
-# Otherwise all make targets requiring area param will show an error.
-# Note: If there are no data files, and user calls  make download area=... once,
-#       they will not need to use area= parameter after that because there will be just a single file.
+# If $(area) parameter is not set, and only one *.osm.pbf file is found in ./data, use it as $(area).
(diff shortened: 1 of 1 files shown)


zstadler:
@nyurik,
Could you review and approve this PR?
It is important for using osmfr

nyurik (approved):
i'm fine with this, but one issue others may have is that this makes the script linux/macOS only, whereas before it was "kinda" able to run on Windows too (at least that's what I heard).

zstadler:
I've tested this PR on both Windows and Linux. 

AFAIK, the OMT project requires a `bash` shell, such as Docker Toolbox, in order to run on Windows.
The ability to run Linux/MacOS commands is already assumed by the `Makefile` in several places, including
```
DC_OPTS ?= --rm --user=$(shell id -u):$(shell id -g)
```
A `bash` shell is also needed in order to run `quickstart.sh`.

TomPohys (approved):

TomPohys:
Thanks @zstadler.

zstadler:
Thanks @nyurik and @TomPohys!

Issue #943, opened 2020-06-26
recipe for target 'update-osm' failed


RS974:
I just ran a new installation test (newly-installed server).

```
git clone https://github.com/openmaptiles/openmaptiles.git
cd openmaptiles
make
./quickstart.sh
```

Installation is successfully completed.
I encountered an error with update-osm.

```
make update-osm
```

Starting postgres docker compose target using default image (no recreate if exists)
Starting openmaptiles_postgres_1 ... done
Detected area=albania based on the found data/ pbf file. Use 'area' parameter (or env var) to override.
docker-compose run --rm -u 0:0 openmaptiles-tools sh -c 'pgwait && import-update'
2020/06/26 11:56:25 config.go:290: open data/albania.repl.json: no such file or directory
Makefile:336: recipe for target 'update-osm' failed
make: *** [update-osm] Error 1

I'm confirming that the data/last.state.txt file is missing #942

TomPohys:
Hi @RS974, when you set `DIFF_MODE=true` in `.env` file and run `make download area=albania`, then it will download also `last.state.txt` file for this area (after #1010 is merged)

frodrigo:
Note: the `DIFF_MODE=true` must be set before the initial import.

RS974:
Hi

I formatted my server, I have no more data.
But I left DIFF_MODE=false during the initial import.

I was hoping for the release  v3.12 to resume my tests.

Thank you for your feedback, I will do a test as soon as possible.

TomPohys:
Hi @RS974, work on [release 3.12](https://github.com/openmaptiles/openmaptiles/milestone/16) is in progress (https://github.com/openmaptiles/openmaptiles/issues/900). Thanks

Issue #942, opened 2020-06-25
Why state file are not download allong PBF in diff mode?


frodrigo:
For diff mode in Makefile we have
```makefile
	@$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools download-osm $(OSM_SERVER) "$(DOWNLOAD_AREA)" \
				--imposm-cfg "$(IMPOSM_CONFIG_FILE)" \
				--output "$(PBF_FILE)"
```

But the state file is missing. On update it requirer the `/import/last.state.txt`, but missing.

```diff
--- a/Makefile
+++ b/Makefile
@@ -285,6 +285,7 @@ ifeq (,$(wildcard $(PBF_FILE)))
        @echo "Downloading $(DOWNLOAD_AREA) with replication support into $(PBF_FILE) and $(IMPOSM_CONFIG_FILE) from $
        @$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools download-osm $(OSM_SERVER) "$(DOWNLOAD_AREA)" \
                                --imposm-cfg "$(IMPOSM_CONFIG_FILE)" \
+                               --state /import/last.state.txt \
                                --output "$(PBF_FILE)"
  else
        @echo "Downloading $(DOWNLOAD_AREA) into $(PBF_FILE) from $(if $(OSM_SERVER),$(OSM_SERVER),any source)"
```

frodrigo:
State file is actually download be default when `DIFF_MODE=true` i initial import.

PR #1010, opened 2020-09-29
download `last.state.txt` if DIFF_MODE=true

TomPohys:
Download `last.state.txt` file with `make download area=?` if `DIFF_MODE=true`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,7 +289,7 @@ ifeq (,$(wildcard $(PBF_FILE)))
 	@echo "Downloading $(DOWNLOAD_AREA) with replication support into $(PBF_FILE) and $(IMPOSM_CONFIG_FILE) from $(if $(OSM_SERVER),$(OSM_SERVER),any source)"
 	@$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools download-osm $(OSM_SERVER) "$(DOWNLOAD_AREA)" \
 				--imposm-cfg "$(IMPOSM_CONFIG_FILE)" \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -289,6 +289,7 @@ ifeq (,$(wildcard $(PBF_FILE)))
 	@echo "Downloading $(DOWNLOAD_AREA) with replication support into $(PBF_FILE) and $(IMPOSM_CONFIG_FILE) from $(if $(OSM_SERVER),$(OSM_SERVER),any source)"
 	@$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools download-osm $(OSM_SERVER) "$(DOWNLOAD_AREA)" \
 				--imposm-cfg "$(IMPOSM_CONFIG_FILE)" \
+				--state /import/last.state.txt \
```
frodrigo, line 292: The file should be named according to the area?

TomPohys: so `--state /import/${DOWNLOAD_AREA}.state.txt` is better?

frodrigo: Yes. But unsure for now from where and how the state file is used.

frodrigo:
I made tests with geofabrik and osm-fr and it look like it is no necessary to download the state file. Imposm create it.

TomPohys:
Could you please write a comment on how to create a state file by Imposm to #942 and #943? After that, I will close this PR. Thanks

frodrigo:
I think we can close this one.

TomPohys:
Thanks

PR #1013, opened 2020-10-01
Allow `BBOX` to be set in `.env` file

zstadler:
Currently, the `BBOX` setting in `.env` is ignored for all areas except `planet`.
On the other hand, the `planet` area is an overkill for any `BBOX` setting other than the default - `-180.0,-85.0511,180.0,85.0511`.

With this PR, `quickstart.sh` would not override a modified `BBOX` value in `.env`. 

Also, this provides a way to avoid the pessimistic `BBOX` computation for `osmfr` extracts, as described in https://github.com/openmaptiles/openmaptiles-tools/pull/297#issuecomment-700792702


zstadler:
@nyurik,

I expect this PR to work well also with the planned replacement of `make generate-dc-config`.
Please review and approve.

nyurik (commented):
I think your PR may conflict with https://github.com/openmaptiles/openmaptiles/pull/1012/files

zstadler:
I believe there is no _functional_ conflict between https://github.com/openmaptiles/openmaptiles/pull/1012 and this PR.

When combined, if `BBOX` was modified in `.env`, `quickstart.sh` would avoid calling `make generate-bbox-file` rather than avoid calling `generate-dc-config`. As a result, there will be no `data/${area}.bbox` file and the `BBOX` value will be taken from the `.env` file.

TomPohys:
Hi @zstadler, thanks for PR. 

I think that the `quickstart.sh` is mainly for beginners. If they e.g. accidentally change the `bbox` in `.env` and not know that changing `bbox` in `.env` will not regenerate `*.dc-config.yml` file for a different area, then they will be maybe lost. 

I think that change a `bbox` in `.env` file should not place generating by `quickstart.sh` at risk.

zstadler:
Hi @TomPohys 

First, a confession - I'm a proud user of `quickstart.sh` in the production flow for https://israelhiking.osm.org.il/ :-)

Unlike _all_ other settings in the `.env` file, modifying `MIN_ZOOM`, `MAX_ZOOM`, and `BBOX` require no knowledge of OMT internals. Their functionality is so basic, it can be understood from their names.
I can see a use-case where novice users takes an country-wide extract and create a map for a specific city or region by setting `BBOX`. The default value also provides a nice hint about the ordering of the bounds.
However, the current functionality implies that the `BBOX` value in `.env` is applicable only to users of "planet" extracts.

`MIN_ZOOM`, `MAX_ZOOM`, and `BBOX` are also the only settings in `.env` that get overwritten by the machine-generated `dc-config.yaml` file.
Luckily, @nyurik's recent work in https://github.com/openmaptiles/openmaptiles-tools/pull/297 and https://github.com/openmaptiles/openmaptiles/pull/1012 will remove the `dc-config.yaml` file and eliminate the overwriting of `MIN_ZOOM` and `MAX_ZOOM`. This PR is independently doing the same for `BBOX`.

Hopefully these PRs will be the last step in a long process of cleaning-up legacy complications, a process that began with the removal of `QUICKSTART_MIN_ZOOM` and `QUICKSTART_MAX_ZOOM` in https://github.com/openmaptiles/openmaptiles/pull/915.

TomPohys:
Hi @zstadler, sorry for the delayed answer. I still thinking about this. It can be very confusing, that change in `BBOX` in .env file will not generate *.bbox file (OMT-T 5.3) for `quickstart.sh` and then block every `make generate-tiles` to that bbox.

manual steps are now like - if there is *.bbox file, take bbox from *.bbox, if there is not *.bbox, take bbox from .env

for quickstart will be like - if not an `area=planet` then generate *.bbox, but if .env change to any value besides `-180.0,-85.0511,180.0,85.0511`, take that value even there is *.bbox file.

I understand, that now is not possible to change bbox for quickstart, but it works every time with the right bbox for the downloaded area.

zstadler:
Hi @TomPohys,

I'm not sure if your comment is about the proposed user experience or the implementation. I would like to discuss the user experience first. If we agree on the desired "what", we'll proceed to the "how".

#### Currently
- If the user does not do anything, `quickstart.sh` and `make generate-tiles` create an `mbtiles` file for the full extent of the data source. This applies to `planet` and other data sources.
- If a user sets the `BBOX` value in `.env` and a `planet` data source is used, `quickstart.sh` and `make generate-tiles` create an `mbtiles` file for the extent set in the `.env` file.
- If a user sets the `BBOX` value in `.env` and a non-`planet` data source is used, it is ignored - `quickstart.sh` and `make generate-tiles` create an `mbtiles` file for the full extent of the data source.

#### Problem statement
While users of a `planet` data source have a simple way to override the default extend of the tile generation, users of other data sources have no simple way of doing that. In fact, for such users the `BBOX` setting in the `.env` file is ignored and therefore misleading.

#### Proposal
- If the user does not do anything, `quickstart.sh` and `make generate-tiles` create an `mbtiles` file for the full extent of the data source. This applies to `planet` and other data sources.
- If a user sets the `BBOX` value in `.env` **<sup>(1)</sup>**, `quickstart.sh` and `make generate-tiles` create an `mbtiles` file for the extent set in the `.env` file. This applies to `planet` and other data sources.

_Note:_
(1) The contents of the `.env` file can be updated to explain and clarify the role of `BBOX`. For example:

```diff
- # BBOX may get overwritten by the computed bbox of the specific area:
- #   make generate-dc-config
- BBOX=-180.0,-85.0511,180.0,85.0511
+ # By default, tile generation is done for the full extent of the given area.
+ # Un-comment and set BBOX below to overwrite it.
+ # BBOX=-180.0,-85.0511,180.0,85.0511
```

TomPohys:
Hi @zstadler, thank you again. Let's use it. Change bbox for `quickstart.sh` is a welcome feature. Maybe update a `quickstart.md` and create a notice, that changing of BBOX in .env will lead to ignoring `*dc-config.yaml` (`*.bbox`) e.g. simmilar as `Set which zooms to generate`.

Let's keep the comment about bbox in `.env` as you already suggested. 

Thanks

TomPohys:
Thanks, look good. I just change bbox site that using the OpenMapTiles background map.

TomPohys (approved):

Issue #308, opened 2017-07-14
Improve split boundaries of the water polygons 


klokan:
Check what can be done to eliminate or improve the water polygons data in the vector tiles.

If styles are not made with workarounds the boundaries gets visible now:

![28102060-1911b954-66cd-11e7-8754-1bd7871169da](https://user-images.githubusercontent.com/59284/28211787-bb58c540-689e-11e7-81d3-bcc76ece4539.png)

If nothing else - then merging in the upper zoom levels where the shapes are generalised can be done. We have elevated the more detailed shapes of the coast to upper zoom levels - so these borders may get sooner visible than in the previous versions of OpenMapTiles now - if the styles are not adjusted.

Ideally, we should eliminate this completely.

Mapzen is doing that in the vector tile post-processing step https://github.com/osm2vectortiles/osm2vectortiles/issues/484

There may be a way how to integrate an merge of all polygons inside of a tile into the SQL tile query during the rendering step with mapnik - if it does bring too high CPU cost...

Related tickets: https://github.com/openmaptiles/positron-gl-style/issues/6

jirik:
I guess that 1. merging polygons and 2. tile buffer would enable styling water's coastline without these visual bugs.

But
- it will be slower
- mbtiles will be probably bigger
- and I think we have more urgent priorities for v3.7

The [positron issue](https://github.com/openmaptiles/positron-gl-style/issues/6) is fixed.

zflamig:
Can I +1 this issue as a priority? Using the water boundaries as an overlay and these extra lines are very distracting. 
<img width="472" alt="screen shot 2017-08-11 at 9 46 49 pm" src="https://user-images.githubusercontent.com/20603302/29237190-a1b100ca-7ede-11e7-82af-28fd0c170384.png">

WBMol:
This issue also appears for me from zoom 8 onwards. Quite distracting, so have to remove the coast outline at zoom 9+ unfortunately. 
![image](https://user-images.githubusercontent.com/13695103/29403892-efdf150c-8339-11e7-893a-6c9e5e2f0a9c.png)

jliebrand:
This was raised more than a year ago - can I assume this will thus never get fixed?

klokan:
As this is not a problem for our default map styles (https://www.maptiler.com/maps/) it is not as urgent for us internally...

[...]

PR #1021, opened 2020-10-09
union ocean regular squares into complex polygon

TomPohys:
The ocean layer (table osm_ocean_polygon insert by `make import-data`) is made from squares that have a 10m buffer. 

This creates extra geometry in vector tiles as:
water_z8a

Union of full ocean squares should decrease the size of the water layer.

From this:
ocean_squares

to this:
ocean_union

The red squares contain the holes from the islands and coastline.


TomPohys:
Partial fix #308

nyurik:
I might be mistaken here, but I think it is done on purpose - breaking water into many polygons greatly speeds up the time needed to generate each tile because it allows much less complex polygons to be processed. Cc @Komzpa

TomPohys:
Hi @nyurik, there are 8042 polygons (squares, 5 points, more then 100km<sup>2</sup>), which was reduced to 22 polygons - 1184kB vs 40kB respectively. 

In vector tiles should be the size reduction even more significant, due to the creation of more polygons based on a 10m buffer of each square.

daliborjanak (approved):

indus:
I may be mistaken but I think this PR is the cause for some spikes in my water layer at zoom 6 (white vertical lines in the second image )
grafik
grafik

Issue #999, opened 2020-09-23
"brunnel" in lower zoom level + adding to transportation_name


eva-j:
1. The property 'brunnel' == 'tunnel' only shows on features from zoom level 12 and on. At z8-11 the property is missing and thus it's not possible to hide the tunnels.

2. For road labels, the property is never there - it's not possible to hide the labels for tunnels.

eva-j:
Solved in #1027 and #1038.

PR #1027, opened 2020-10-13
Brunnel changes in transportation and transportation_name layers

eva-j:
https://github.com/openmaptiles/openmaptiles/issues/999
- Point 1 solved only partially - there is `brunnel` value added into zooms 9, 10, 11. For zooms 8 and smaller it would need change in osm_transportation_merge_ tables - I am going to try it but maybe its SQL costs will be to high.
- Point 2 solved.


TomPohys (approved):

TomPohys:
Thanks @eva-j.

Issue #1022, opened 2020-10-09
make import-sql fails in New-Zealand with lwgeom_union: GEOS Error: TopologyException:


quentar:
Trying to generate `australia-oceania` on fresh machine (no previous db), it fails currently on import-sql step, 
drilling down it seems it is is `australia-oceania/new-zealand` sub region (ie `quickstart  australia-oceania/new-zealand` fails in the same way)
Last release `3.11 `does work with same `osm.pbf` input files - completing all steps and generate tiles.
It seems reproducible for me every time running this region now.

Console :
```
CREATE FUNCTION
Time: 0.991 ms
psql:/sql/parallel/building.sql:82: ERROR:  lwgeom_union: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection at or near point 19566767.608174294 -4873356.7422359968 at 19566767.608174294 -4873356.7422359968
CONTEXT:  PL/pgSQL function osm_building_block_gen1() line 6 at FOR over SELECT rows
Time: 150915.098 ms (02:30.915)
```

`docker-compose logs`  : 
```
postgres_1                      | ERROR:  lwgeom_union: GEOS Error: TopologyException: Input geom 1 is invalid: Self-intersection at or near point 19566767.608174294 -4873356.7422359968 at 19566767.608174294 -4873356.7422359968
postgres_1                      | CONTEXT:  PL/pgSQL function osm_building_block_gen1() line 6 at FOR over SELECT rows
postgres_1                      | STATEMENT:  CREATE MATERIALIZED VIEW osm_building_block_gen1 AS
postgres_1                      | 	SELECT *
postgres_1                      | 	FROM osm_building_block_gen1();
postgres_1                      | ERROR:  canceling autovacuum task
```

(Complete docker-compose logs do not seem helpful but for completeness)
<details>
Attaching to openmaptiles_postgres_1
postgres_1                      |
postgres_1                      | PostgreSQL Database directory appears to contain a database; Skipping initialization
postgres_1                      |
postgres_1                      | LOG:  database system was shut down at 2020-06-05 17:19:18 UTC
postgres_1                      | LOG:  MultiXact member wraparound protections are now enabled
postgres_1                      | LOG:  autovacuum launcher started
postgres_1                      | LOG:  database system is ready to accept connections
[...]

PR #1028, opened 2020-10-14
buildings: prevent slivers of self-intersecting geometry

quentar:
To address issue #1022 , where current New Zealand region with invalid data which aborts the import-sql loop.

It seems adding ST_IsValid condition to would pass and does not pass through New Zealand, however
flipping negative and positive radius for ST_Buffer helps to remove self-intersection.
It was suggested online at https://gis.stackexchange.com/questions/281935/topology-exception-on-geometry-intersection

Hopefully that should not have performance impact as it is not adding to the query.
Visually checking, regions buildings looks the same in output tiles, with benefit of New Zealand passing.

💡 At first I've tried adding ST_IsValid conditions at   https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/update_building.sql#L27 
OR https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/update_building.sql#L31
though it still aborted.

Issue log: (when doing `quickstart.sh new-zealand` , `make import-sql` step) 
<details>
CREATE FUNCTION
Time: 0.991 ms
psql:/sql/parallel/building.sql:82: ERROR:  lwgeom_union: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection at or near point 19566767.608174294 -4873356.7422359968 at 19566767.608174294 -4873356.7422359968
CONTEXT:  PL/pgSQL function osm_building_block_gen1() line 6 at FOR over SELECT rows
Time: 150915.098 ms (02:30.915)
</details>

### DIFF
--- a/layers/building/update_building.sql
+++ b/layers/building/update_building.sql
@@ -27,7 +27,7 @@ BEGIN
             FROM osm_building_polygon o
         )
         SELECT (array_agg(dta.osm_id))[1]                                                                   osm_id,
-               ST_Buffer(ST_MemUnion(ST_Buffer(dta.geometry, zres14, 'join=mitre')), -zres14, 'join=mitre') geometry
+               ST_Buffer(ST_MemUnion(ST_Buffer(dta.geometry, -zres14, 'join=mitre')), zres14, 'join=mitre') geometry
         FROM dta
         GROUP BY cid
 


TomPohys:
Hi @quentar, thanks for PR. If the visual check is OK, let's use it.

TomPohys:
Hi @quentar, where did you test the buildings? 

The green buildings are original, red is your suggestion. Ideally, there should be just red buildings. A lot of buildings in zoom level 13 is now missing (pictures from Albania).

Screenshot from 2020-10-14 17-00-10

and detail with all building (zoom 14)
Screenshot from 2020-10-14 17-04-43

quentar:
@TomPohys Thank you for testing that! I did not use this tool. I don't have many ideas what to try next though.

fricna:
Was this issue resolved? I still encounter it when trying to process New Zealand and also Spain.

TomPohys:
Let's close this PR in advance of https://github.com/openmaptiles/openmaptiles/pull/1044. 

New Zealand was fixed by ST_SnapToGrid function in [update_building.sql](https://github.com/openmaptiles/openmaptiles/pull/1044/files#diff-22d79c2879e2711c40bfdf951f23192854728687daa287c5dd8ec926372e57fcR44)

PR #1042, opened 2020-11-19
Fix landcover generalization.

lazaa32:
This PR brings few modification which make it possible to generalize landcover on larger areas. Tested on Europe.

- Use ST_SnapToGrid to avoid ERROR:  GEOSUnaryUnion: TopologyException: found non-noded intersection between LINESTRING.
- Cluster only landcover subclasses: wood and forest.
- Use less generalized tables as a starting point for next step generalization.


TomPohys (approved):

TomPohys:
Thanks @lazaa32.

PR #1048, opened 2020-12-04
Enable pgquery

smellman:
I try to use pgquery, it needs some configuration in .env file.
I think if we use pgquery instead of mapnik, the performace will be good.


TomPohys:
Hi @smellman, thanks for the PR. 
For backward compatibility, I would like to keep Mapnik still as the primary rendering tool. There are still not cleaned dependencies and packages for Mapnik. Could we make another makefile target like `make generate-tiles-pg` which will be used as a generator using PGquery?

smellman:
@TomPohys Thank you your comment and fixes.
I tested new Makefile with japan region and zoom 0-14, and no problem.

TomPohys (approved):

TomPohys:
Thanks @smellman!

Issue #1064, opened 2021-01-16
Add city subdivision place=quarter


jdhoek:
The OSM-based website [knooppuntnet.nl](https://knooppuntnet.nl/en/map/hiking) uses OSM Liberty through OpenMapTiles. The rendered tiles currently lack `place=quarter` tags, even though the larger `place=suburb` and smaller `place=neighbourhood` are included. I would like to see `place=quarter` rendered as well (https://github.com/vmarc/knooppuntnet/issues/136).

Because `place=quarter` sits between `place=suburb` and `place=neighbourhood` in the OSM tagging hierarchy of city subdivisions, adding it to OpenMapTiles looks straightforward. I have done so here: https://github.com/jdhoek/openmaptiles/commit/bc8b21c883e34ef58289861bc0aaa7a54e2245c9

Is this sufficient to be able to use it in a style such as OSM Liberty and submit a patch there?

TomPohys:
Hi @jdhoek, thank you for your notice. It looks useful to have a `quarter` in OpenMapTiles. Do you have time to create a PR?

jdhoek:
I had the commit ready, but this project's guidelines forbade offering a PR without an issue and the `help wanted` tag edit by one of the maintainers.

jdhoek:
#1065 has that commit.

PR #1065, opened 2021-01-18
Add place=quarter to city subdivisions

jdhoek:
This adds `place=quarter` which was introduced in 2011 as a subdivision
of villages/towns/cities as a value that sits between `place=suburb` and
`place=neighbourhood`. It has by now been used over 40,000 times and is
rendered in the OSM-Carto.

Might solve #1064.


jdhoek:
This is just a very naïve commit that puts `place=quarter` in between anywhere `place=suburb` and `place=neighbourhood` are used. I don't know if this is sufficient, or if this brakes things.

jdhoek:
In terms of ordering this is correct of course, all three are urban subdivisions, starting from the largest: `suburb`, `quarter`, `neighbourhood`. So where the other two are used `quarter` sits between them.

TomPohys:
Thank you for PR. Could you please update a devdoc by `make generate-devdoc` to update diagrams? Thanks

jdhoek:
Done. Two files changed, looks good:

mapping_diagram

mapping_diagram

TomPohys (approved):

PR #1069, opened 2021-01-26
`make generate-tiles-pg` without editing `.env`

### LINKED ISSUE #1070: `make generate-tiles-pg` without editing `.env`
Currently, `make generate-tiles-pg` always produces an error

```
[Error: SQLITE_CANTOPEN: unable to open database file] {
  errno: 14,
  code: 'SQLITE_CANTOPEN'
}
make: *** [generate-tiles-pg] Error 1
```

In [this Slack thread](https://osmus.slack.com/archives/CH6GFAFRC/p1609608482047400), @TomPohys wrote:
> Hi, did you uncomment line 55 in `.env` ? I got this error when I forgot to uncomment `EXPORT_DIR`.

In addition to not being obvious, such editing of the `.env` file breaks the ability to run `quicktime.sh`.

zstadler:
Currently, `make generate-tiles-pg` always produces an error

```
[Error: SQLITE_CANTOPEN: unable to open database file] {
  errno: 14,
  code: 'SQLITE_CANTOPEN'
}
make: *** [generate-tiles-pg] Error 1
```

In [this Slack thread](https://osmus.slack.com/archives/CH6GFAFRC/p1609608482047400), @TomPohys wrote:
> Hi, did you uncomment line 55 in `.env` ? I got this error when I forgot to uncomment `EXPORT_DIR`.

In addition to not being obvious, such editing of the `.env` file breaks the ability to run `quicktime.sh`.

This PR allows using `make generate-tiles-pg` without editing `.env` while keeping `quickstart.sh` available for most users.

Resolves #1070

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -412,7 +412,7 @@ generate-tiles-pg: all start-db
 	@$(assert_area_is_given)
 	@echo "Generating tiles into $(MBTILES_LOCAL_FILE) (will delete if already exists)..."
 	@rm -rf "$(MBTILES_LOCAL_FILE)"
-	$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools generate-tiles
+	$(DOCKER_COMPOSE) run $(DC_OPTS) -e EXPORT_DIR=/import openmaptiles-tools generate-tiles
 	@echo "Updating generated tile metadata ..."
 	$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools \
 			mbtiles-tools meta-generate "$(MBTILES_LOCAL_FILE)" $(TILESET_FILE) --auto-minmax --show-ranges


lazaa32:
Hi @zstadler,

I don't like the current approach with editing .env either. However your solution brings the approach of exporting environment variable for container as part of make target which we do not use anywhere else.
I would prefer simply mount `./data:/export` in docker-compose (see https://github.com/openmaptiles/openmaptiles/blob/master/docker-compose.yml#L55).
That way generated mbtiles will be placed in `/data` folder once the generating is done, we don't have to modify .env file and it behaves the same way as generating tiles using mapnik (see https://github.com/openmaptiles/openmaptiles/blob/master/docker-compose.yml#L78)

zstadler:
Hi @lazaa32,

Will you create a PR that resolves #1070 using the approach you described?

zstadler:
Superceeded by #1071

PR #1079, opened 2021-02-23
Remove useless indices on osm_transportation_merge_linestring_genX

frodrigo:
On creation of materialized views `osm_transportation_merge_linestring_gen_zX`, an index is created each time the new view and not used else where than the next generalization step.
It does not worth it to to a scan on the table to create index that will be used only once, while the next generalisation step can also be done in one full scan.
So, removing these indices.


TomPohys (approved):

TomPohys:
Thanks a lot, @frodrigo.

Issue #1084, opened 2021-03-09
embassy and project report on taginfo


Marc-marc-marc:
in 2018, the community approved  the depreciation of the amenity=embassy tag and approved office=diplomatic diplomatic=embassy
osm2VectorsTiles is listed as using the old tag and I don't find OpenMapTiles [in the project using the new tag](https://taginfo.openstreetmap.org/tags/office=diplomatic#projects)
is OpenMapTiles using the tag but does not report it to taginfo? or is it still the old tag that is used and if so, could you take the new one into account?

TomPohys:
Hi @Marc-marc-marc, you are right, the OMT is using the old `amenity=embassy`. It should be updated to `office=diplomatic`.

frodrigo:
`office=diplomatic` should replace `amenity=embassy`? or as an addition?
From a quick survey, only the Bright style use `embassy`.

Marc-marc-marc:
office=diplomatic replace amenity=embassy.
there are still ~2000 objects having only the old tag (mostly errors as the correct objects have been massively converted) , a manual review and mass editing procedure is underway

PR #1085, opened 2021-03-11
Switch from amenity=embassy to office=diplomatic #1084

frodrigo:
Switch from amenity=embassy to office=diplomatic as was about here #1084.
Still unsure about switch or addition.


TomPohys:
Every month there are more and more `office=diplomacy` at the expense of `amenity=embassy`. So, let's switch these two and using only `office=diplomacy`.

https://taginfo.openstreetmap.org/tags/amenity=embassy#chronology
https://taginfo.openstreetmap.org/tags/office=diplomatic#chronology

Do you agree?

frodrigo:
Rebased.
Change still ok for me.

TomPohys (approved):

TomPohys:
Thanks, looks great!

PR #1089, opened 2021-03-16
Avoid grep errors in `clean-unnecessary-docker`

zstadler:
When `make clean-unnecessary-docker` is run twice, or when there are no docker images tagges as "\<none\>", make fails because grep fails:

```
Deleting unnecessary container(s)...
Deleting unnecessary image(s)...
make: *** [Makefile:562: clean-unnecessary-docker] Error 1
```

Using `awk` to _also_ search for "\<none\>", instead of `grep`, solves this issue.


TomPohys (approved):

TomPohys:
Thanks a lot, @zstadler.

PR #1093, opened 2021-03-25
Update naming convention for networks

remyleone:
Update naming convention for networks with the goal to make it easier to use with kompose

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,7 @@ volumes:
   pgdata:
 
 networks:
-  postgres-conn:
+  postgres:
     driver: bridge
 
 services:
@@ -16,7 +16,7 @@ services:
     volumes:
       - pgdata:/var/lib/postgresql/data
     networks:
-      - postgres-conn
+      - postgres
     ports:
       - "5432"
     env_file: .env-postgres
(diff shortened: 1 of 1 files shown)


nyurik:
Would it make sense to get rid of the dash/underscore completely, e.g. use `postgres` or even `pg` network name?  Simpler is usually better :)

remyleone:
Let's go for `postgres` then :)

TomPohys (approved):

TomPohys:
Thanks @remyleone

frodrigo:
Should also be fixed here?
- https://github.com/openmaptiles/openmaptiles-tools/blob/master/tests/sql/docker-compose.yml
- https://github.com/openmaptiles/openmaptiles-tools/blob/master/README.md

nyurik:
> Should also be fixed here?
> 
> * https://github.com/openmaptiles/openmaptiles-tools/blob/master/tests/sql/docker-compose.yml
> * https://github.com/openmaptiles/openmaptiles-tools/blob/master/README.md

good catch @frodrigo , want to submit a PR?  I will try to get to it tomorrow

remyleone:
I can do it :)

PR #1096, opened 2021-03-30
Improve logging - show BBOX if exists

nyurik:
The generate-bbox target will now show the value of BBOX if it is already generated.

(only affects the logs)


TomPohys (approved):

TomPohys:
Thanks!

PR #1101, opened 2021-04-16
Fix own OSM instructions

jvolker:
Before it was throwing `ERROR:  make generate-bbox-file area=<area-id>` and `Area source has not been specified. Auto-detecting...
ID 'my' was not found in Geofabrik.` `make: *** [download] Error 1`

# Pull request guidelines

Before working on a pull request, create an issue explaining what you want to contribute. This ensures that your pull request won't go unnoticed, and that you are not contributing something that is not suitable for the project. Once a core developer has set the `help wanted` label on the issue, you can submit a pull request. The pull request description should reference the original issue.

Your pull request must:

 * Address a single issue or add a single item of functionality.

 * Contain a clean history of small, incremental, logically separate commits,
   with no merge commits.

 * Use clear commit messages.

 * Be possible to merge automatically.

When you are making a PR that adds new spatial features to OpenMapTiles schema, please make also PR for at least one of our GL styles to show it on the map. Visual check is crucial.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -368,8 +368,8 @@ Mbtiles can be generated from an arbitrary osm.pbf (e.g. for a region that is no
 ```
 mkdir -p data
 mv my-latest.osm.pbf data/
-make generate-bbox-file
-./quickstart.sh my
+make generate-bbox-file area=my-latest
+./quickstart.sh my-latest
 ```
 
 ### Check postserve
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -367,9 +367,9 @@ Mbtiles can be generated from an arbitrary osm.pbf (e.g. for a region that is no
 
 ```
 mkdir -p data
-mv my-latest.osm.pbf data/
-make generate-bbox-file area=my-latest
-./quickstart.sh my-latest
+mv Tanzania-latest.osm.pbf data/
+make generate-bbox-file area=Tanzania-latest
+./quickstart.sh Tanzania-latest
 ```
 
 ### Check postserve
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -363,13 +363,13 @@ This is generating `.mbtiles` for your area :  [ MIN_ZOOM: "0"  - MAX_ZOOM: "7" 
 ./quickstart.sh yukon                        # Yukon, Canada
 ```
 ### Using your own OSM data
-Mbtiles can be generated from an arbitrary osm.pbf (e.g. for a region that is not covered by an existing extract) by making the `data/` directory and placing an *-latest.osm.pbf inside. Inside of folder have to be `docker-compose-config.yml` file too, otherwize whole folder `data/` will be deleted and `download-osm` will try to download osm.pbf file from `geofabric`, `osmfr` or `bbbike`.
+Mbtiles can be generated from an arbitrary osm.pbf (e.g. for a region that is not covered by an existing extract) by making the `data/` directory and placing an *.osm.pbf (e.g. `mydata.osm.pbf`) inside.
 
 ```
 mkdir -p data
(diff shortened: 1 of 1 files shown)


nyurik (approved):

nyurik:
I think `my-latest` is a bit confusing name for an area. Something like `area-name` might be a bit clearer, but I'm ok with the current as well if you want to keep it

jvolker:
> I think my-latest is a bit confusing name for an area.

I agree and had the same feeling.

Actually, is the `-latest` still are a requirement?

nyurik:
> > I think my-latest is a bit confusing name for an area.
> 
> Actually, is the `-latest` still are a requirement?

I don't think latest is needed anymore -- makefile has a section to auto-rename any `*-latest.osm.pbf` to drop the `-latest`.

nyurik:
I think i'm having second thoughts about the names -- using a specific area name implies it was downloaded in some way with the standard tools.... so maybe we should keep it as `area=mydata` ?

TomPohys:
As this is a readme document, I would like to avoid any specific area name e.g. Tanzania. I can be misleading if it custom area or downloaded from e.g. Geofabrik (agree with nyurik). 

> Actually, is the `-latest` still are a requirement?

No, it is not a requirement anymore. The renaming from `name-latest.osm.pbf` to `name.osm.pbf` is done only if in `data` folder exists only one `*.osm.pbf`. The `.osm.pbf` is the only requirement. If in `data` folder are e.g. `mydata-latest.osm.pbf` and `my_area-latest.osm.pbf` then the renaming will not be apply (`make generate-bbox-file area=mydata-latest`).

the `make generate-bbox-file` works, if in `data` folder is only one `*.osm.pbf` file. If there are many `*.osm.pbf`, then `make generate-bbox-file area=name` (for `name.osm.pbf` file in `data` folder)

for run `./quickstart.sh` is enough to specify name (e.g. `./quickstart.sh my_area` for `my_area.osm.pbf`, `./quickstart.sh my_area2` for `my_area2.osm.pbf`)

My suggestion is then `my_area` or `mydata` instead of `my`

```diff

  mkdir -p data
-  mv my-latest.osm.pbf data/
+  mv my_area.osm.pbf data/
  make generate-bbox-file
-  ./quickstart.sh my
+  ./quickstart.sh my_area
```

The text before code
```
### Using your own OSM data
Mbtiles can be generated from an arbitrary osm.pbf (e.g. for a region that is not covered by an existing extract) by making the `data/` directory and placing an *-latest.osm.pbf inside. Inside of folder have to be `docker-compose-config.yml` file too, otherwize whole folder `data/` will be deleted and `download-osm` will try to download osm.pbf file from `geofabric`, `osmfr` or `bbbike`.
```
can be shortened to 
```
### Using your own OSM data
Mbtiles can be generated from an arbitrary `osm.pbf` (e.g. for a region that is not covered by an existing extract) by making the `data/` directory and placing a `*.osm.pbf` inside.
```

PS:
If in `data` folder is only `my_area-latest.osm.pbf` and in terminal you will write `make gen` and then use `Tab` for autocomplete, the `*.pbf` is rename even before starting the command. Maybe I have some strange setup of Ubuntu.

TomPohys (changes requested):
Instead of Tanzania (or originally `my`), please change to `mydata`, or `my-area`, or `my-area-name`, ...

Thanks

TomPohys (approved):

TomPohys:
It is OK now?

nyurik (approved):

jvolker:
Sorry for the late reply. I didn't find the time to get back to this earlier. 

Yes, looks all good now. Thanks for the changes. 👍 

I had another issue with this though. I started a new thread for it here: https://github.com/openmaptiles/openmaptiles/issues/1105

Issue #1116, opened 2021-05-30
Update documentation with requirement to re-run imposm for new layers


ZeLonewolf:
The README.md currently says:

Each time you modify layer SQL code run `make` and `make import-sql`

The README should also note that when adding/changing imposm mappings (for example, when creating a new vector layer) that `make import-osm` needs to be run in order to create the new tables.

TomPohys:
Thanks for point this out. This should be added to readme.

PR #1118, opened 2021-05-31
Add notice into README about import-osm

TomPohys:
Adding information when run `make import-osm`.

closes https://github.com/openmaptiles/openmaptiles/issues/1116


README.md
```diff
@@ -130,6 +130,14 @@ make import-wikidata
 ```
 
 ### Work on Layers
+Each time you modify layer's `mapping.yaml` file or adding new OSM tags run `make` and `make import-osm` for recreate (with additional data) tables in PostgreSQL. With new data there can be a new Wikidata records.
```
ZeLonewolf, line 133: Suggested copy-edit:

Each time you modify a layer's `mapping.yaml` file or add new OSM tags, run `make` and `make import-osm` to recreate tables (potentially with additional data) in PostgreSQL. With the new data, there can be new Wikidata records also.

TomPohys: Thanks

eva-j (approved):

PR #1119, opened 2021-05-31
Add new layer to serve highway=motorway_junction nodes

### LINKED ISSUE #1114: Add support for highway=motorway_junction
Highway exits are an important feature of street centric maps.  Adding `highway=motorway_junction` and `ref` would allow rendering of highway exits as icons and/or labels.

ZeLonewolf:
Closes #1114 
Documentation update in openmaptiles/www.openmaptiles.org#68
Unblocks ZeLonewolf/openstreetmap-americana#12

This PR adds a layer for `highway=motorway_junction` features.

Please let me know if the PR is incomplete, and I would be happy to address any deficiencies.

A visual example from openstreetmap-americana in Rhode Island, USA is below:
image


TomPohys:
Thanks for awesome PR. It looks very good. 

The extra layer for highway junction is not a necessity. We think that this point layer can be implemented into `transportation_name` layer next to `highway_polygon` and `highway_linestring`. There are all attributes in `transportation_name` that you are using (basically `name` and `ref`) and some extra that will not be used. 

Would it be possible, in suggested `mapping.yaml`, to rename the`motorway_junction:` to `highway_point:` and add it into [layers/transportation/mapping.yaml](https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/mapping.yaml#L334-L368) similar as e.g. `highway_polygon`?

Then add the suggested functionality into `transportation_name.sql` or move some preprocessing into `update_transportation_name.sql`?

Or do you have any other idea how to implement it without an extra layer?

ZeLonewolf:
Thanks for the review; I agree it should be feasible to amend the `transportation_name` layer to add motorway junctions.  I do not have the permissions to mark this PR as draft, however, I will be working on changes.

ZeLonewolf:
One issue is that `highway=motorway_junction` can be present on `highway=motorway` but also potentially other road classes, such as `highway=trunk` which might have exits.  It seems reasonable that a style might wish to know which class of road the junction is a part of (for example, to control zoom-specific behavior).

I'm not sure the best way to encode this information in the existing structure.  My initial thought was something like:

class: motorway/trunk/primary (etc)
subclass: junction

My only concern with this structure is whether it might be an issue that actual roadways encoded class: motorway/trunk etc might be confused with the junction objects.

TomPohys:
I think, that junctions will be filtered in style base on geometry type as `point` where all other features in classes motorway/trunk/primary are `LineString`.

ZeLonewolf:
I can't seem to figure out why "subclass" is not appearing in the `transportation_name` layer no matter what I do.  It shows up in tiles.json, and is present in the yaml layer definition and in the returned SQL table.  Is there something I'm fundamentally doing wrong here?

ZeLonewolf:
This mockup demonstrates differentiating between highway exits on motorways versus exits on lesser classes of road (in this case, the motorway has big diamonds and labels, while the trunk road exits have small diamonds and no labels:
image

ZeLonewolf:
This implementation of highway exits in the transportation_name layer add to the existing layer table structure, and renames the internal column name from "construction" (which was already overloaded with non-construction usages) to "subclass", which will be less confusing to future developers.  The string 'junction' is used as the universal sub-class for highway exits.

A new documentation PR has been opened at openmaptiles/www.openmaptiles.org#69 to reflect these changes in the documentation.

I'm still navigating my way through the nuts and bolts so feedback welcome :)

TomPohys (approved):

TomPohys:
It looks great! Thank you very much!

PR #1121, opened 2021-06-05
Improve "make help" output

ZeLonewolf:
This PR makes three improvements to the current `make help` output:

1. Adds additional help entries for useful `make` targets that were not currently in `make help`
2. Pipes the help output through `less` so that the user can scroll up or down through the various `make` targets if it overflows the length of the user's screen
3. Re-groups the "developer" targets into separate entries for database and docker-related commands

In addition, an unneeded whitespace was removed from line 7 of the Makefile.


TomPohys:
Thanks for additional help targets and re-groups. Just one question about `less`, it is necessary to create the extra file in to project root folder? It can be used like `make help | less`. Or is it possible to tune-up makefile to use less without `| less`?

ZeLonewolf:
I updated the PR to put the help content inline in the Makefile via an environment variable, which still preserves the single echo command and then pipes to less.  If we don't want to include a call to less in the Makefile, I think the options are:

1. Remove it and tolerate the long output (user can still call `make help | less` as suggested)
2. Create sub-help targets, where `make help` would list all the sub-targets, and you could call something like `make help-db` or `make help-docker` and it would give you just the help from that section.

TomPohys:
Thank you, nice work! Let's keep it with `less`.

TomPohys (approved):

PR #1129, opened 2021-06-28
Update osm_poi_point only on change

frodrigo:
After each diff import the table `osm_poi_point` is fully rewritten due to a full update of the field tags. It is now good to do an a update, event if the content does not change, postgres delete and reinsert internally the record. Resulting in more write and internal table size raising.

Note: not directly linked, but there is a problem in this case, the autocaccum is not sufficient to keep this table size moderate, but grow indefinitely.

### DIFF
--- a/layers/poi/update_poi_point.sql
+++ b/layers/poi/update_poi_point.sql
@@ -17,7 +17,8 @@ BEGIN
 
     UPDATE osm_poi_point
     SET tags = update_tags(tags, geometry)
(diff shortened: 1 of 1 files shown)


TomPohys:
Thanks for the improvement! Could you please rebase it? Thanks a lot!

frodrigo:
Rebased.

TomPohys (approved):

TomPohys:
Thanks!

PR #1132, opened 2021-06-30
Remove empty subclass tags from tiles

ZeLonewolf:
PR #1119 (adding support for `highway=motorway_junction`) missed a few cases where NULLIF functions were needed in order to suppress empty-string `subclass` tags from appearing in vector tiles, resulting in unnecessary  `subclass` entries on `transportation_name` objects where it was not needed, for example:
image

This PR adds the missing NULLIF function so that those empty tags are suppressed:
image


TomPohys (approved):

TomPohys:
Thanks!

PR #1135, opened 2021-07-01
Compute separate concurrency_index on osm_route_member

### LINKED ISSUE #1134: Store multiple-route concurrency indexes
Unblocks #1128
Unblocks #1131

This issue proposes to compute and store the rank-order number of [concurrencies](https://en.wikipedia.org/wiki/Concurrency_(road)) in the `osm_route_member` table, in order from most important multiple-route concurrency to least-important.  

Thus, a hypothetical concurrency between I-90, I-80, and US-6 in the United States would look like this:
I-80: concurrency-index=1
I-90: concurrency-index=2
US-6: concurrency-index=3

This change has four purposes:

1. It is a required change for #1128 to support exposing multiple concurrencies in the `transportation_name` layer.
2. It appears to have better performance than the current approach of using a `ROW_NUMBER()` / `rank` scheme on a table join.
3. It allows us to enforce a specific ranked ordering of concurrency importance (which routes "go first" in an ordered list).
4. It allows for easier troubleshooting of route-related issues such as #1131.

The following changes are proposed:

1. Modify the `osm_route_member` table to add and populate a `concurrency_index` column, which would store rank-ordered route concurrencies in the following sort order:
 - network type (for example, `us-interstate`)
 - network name
 - ref length
 - ref

2. Change the construction of the `osm_transportation_name_network` table to replace the `ROW_NUMBER()` / `rank` table join with a simple join on `osm_route_member` using `concurrency_index=1`.

ZeLonewolf:
Closes #1134 

Adds a `concurrency_index` column to `osm_route_member` as described in #1134.  For example, listing concurrencies on a [segment of the Washington Bridge](https://www.openstreetmap.org/way/43080535) in Providence, Rhode Island, USA:

```
openmaptiles=# select osm_id, network, ref, network_type, name, concurrency_index from osm_route_member where member=43080535 order by concurrency_index asc;
  osm_id  | network | ref | network_type  |     name      | concurrency_index
----------+---------+-----+---------------+---------------+-------------------
 -1694950 | US:I    | 195 | us-interstate | I 195 (RI/MA) |                 1
 -2308410 | US:US   | 6   | us-highway    | US 6 (RI)     |                 2
 -1347856 | US:US   | 1A  | us-highway    | US 1A (RI)    |                 3
 -2309143 | US:US   | 44  | us-highway    | US 44 (RI)    |                 4
(4 rows)
```

### DIFF
--- a/layers/transportation_name/update_route_member.sql
+++ b/layers/transportation_name/update_route_member.sql
@@ -91,3 +91,12 @@ CREATE INDEX IF NOT EXISTS osm_route_member_name_idx ON osm_route_member ("name"
 CREATE INDEX IF NOT EXISTS osm_route_member_ref_idx ON osm_route_member ("ref");
 
 CREATE INDEX IF NOT EXISTS osm_route_member_network_type_idx ON osm_route_member ("network_type");
+
+ALTER TABLE osm_route_member ADD COLUMN IF NOT EXISTS concurrency_index int;
+
+INSERT INTO osm_route_member (id, concurrency_index)
+  SELECT
+    id,
+    ROW_NUMBER() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index
+  FROM osm_route_member
+  ON CONFLICT (id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index;
--- a/layers/transportation_name/update_transportation_name.sql
+++ b/layers/transportation_name/update_transportation_name.sql
@@ -42,16 +42,12 @@ FROM (
         CASE WHEN highway IN ('footway', 'steps') THEN layer END AS layer,
         CASE WHEN highway IN ('footway', 'steps') THEN level END AS level,
         CASE WHEN highway IN ('footway', 'steps') THEN indoor END AS indoor,
-        ROW_NUMBER() OVER (PARTITION BY hl.osm_id
-            ORDER BY rm.network_type) AS "rank",
         hl.z_order
     FROM osm_highway_linestring hl
-            LEFT JOIN osm_route_member rm ON
-        rm.member = hl.osm_id
+            LEFT OUTER JOIN osm_route_member rm ON rm.member = hl.osm_id AND rm.concurrency_index=1
     WHERE (hl.name <> '' OR hl.ref <> '')
       AND NULLIF(hl.highway, '') IS NOT NULL
-) AS t
-WHERE ("rank" = 1 OR "rank" IS NULL);
+) AS t;
 CREATE INDEX IF NOT EXISTS osm_transportation_name_network_osm_id_idx ON osm_transportation_name_network (osm_id);
 CREATE INDEX IF NOT EXISTS osm_transportation_name_network_name_ref_idx ON osm_transportation_name_network (coalesce(name, ''), coalesce(ref, ''));
 CREATE INDEX IF NOT EXISTS osm_transportation_name_network_geometry_idx ON osm_transportation_name_network USING gist (geometry);
@@ -308,18 +304,14 @@ BEGIN
             CASE WHEN highway IN ('footway', 'steps') THEN layer END AS layer,
             CASE WHEN highway IN ('footway', 'steps') THEN level END AS level,
             CASE WHEN highway IN ('footway', 'steps') THEN indoor END AS indoor,
-            ROW_NUMBER() OVER (PARTITION BY hl.osm_id
-                ORDER BY rm.network_type) AS "rank",
             hl.z_order
         FROM osm_highway_linestring hl
                 JOIN transportation_name.network_changes AS c ON
             hl.osm_id = c.osm_id
-                LEFT JOIN osm_route_member rm ON
-            rm.member = hl.osm_id
+                LEFT OUTER JOIN osm_route_member rm ON rm.member = hl.osm_id AND rm.concurrency_index=1
         WHERE (hl.name <> '' OR hl.ref <> '')
           AND NULLIF(hl.highway, '') IS NOT NULL
-    ) AS t
-    WHERE ("rank" = 1 OR "rank" IS NULL);
+    ) AS t;
 
     -- noinspection SqlWithoutWhere
     DELETE FROM transportation_name.network_changes;


TomPohys:
Thanks, looks good!

TomPohys (approved):

farfromrefug:
@TomPohys  i cant build anymore after that PR.  I get the error `ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification`

ZeLonewolf:
Can you provide steps to reproduce?  I've built many times with this change in place.

farfromrefug:
I building for france and get it on every run. Though i know think it might be because i use a newer version of imposm. I will check

farfromrefug:
I use this fork of imposm3 https://github.com/farfromrefug/imposm3 but it should not make a difference. Could you test with "Lyon bbbike" to see if you see the error too? It is small and broke here. 
I reverted this commit and now it works

ZeLonewolf:
I'm currently building a France mbtiles (using just a stock `./quickstart.sh france`) and it seems to be working so far without issue.

The section of SQL that the error message is referring to is this:

```
ALTER TABLE osm_route_member ADD COLUMN IF NOT EXISTS concurrency_index int;

INSERT INTO osm_route_member (id, concurrency_index)
  SELECT
    id,
    ROW_NUMBER() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index
  FROM osm_route_member
  ON CONFLICT (id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index;
```

What that code does is to add a column to `osm_route_member` (if it doesn't exist) and then populates that column with the concurrency rank of routes associated with a way (useful for when a way is a member of more than one route).  The `ON CONFLICT ... DO UPDATE` syntax is equivalent to `REPLACE INTO` or `ON DUPLICATE KEY UPDATE` commands in MySQL, for example.

This syntax is supposed to be valid as of postgres 9.5.

Since it works on the dockerized openmaptiles postgres but fails on yours, I suspect the different may be with the version of postgres in use?  The postgres in the dockerized container in the distro is reported as:
`PostgreSQL 9.6.20 on x86_64-pc-linux-gnu (Debian 9.6.20-1.pgdg90+1), compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit`

Can you check on which version of postgres you're running?

ZeLonewolf:
Here's the [quickstart.log](https://github.com/openmaptiles/openmaptiles/files/6835582/quickstart.log) from a load of France using the current OpenMapTiles master.

farfromrefug:
Thanks a lot for looking deeply at this. I ll report tonight whrn in front of my computer

farfromrefug:
@ZeLonewolf sorry for the late answer. Looking at my openmaptiles_postgres_1 container the version is 9.6.21-1.pgdg90+1>

ZeLonewolf:
@farfromrefug there's no obvious reason why that shouldn't work.  What happens when you log into the container with psql and run the SQL fragment in https://github.com/openmaptiles/openmaptiles/pull/1135#issuecomment-881958349 ?

farfromrefug:
@ZeLonewolf i ll test. My build for france is still running. It takes likes 4 days on my machine :P May i ask what computer you ran it on to be so fast?

ZeLonewolf:
Just my day-to-day laptop with SSD hard drive:
image

Cactusbone:
> 
> 
> I'm currently building a France mbtiles (using just a stock `./quickstart.sh france`) and it seems to be working so far without issue.
> 
> The section of SQL that the error message is referring to is this:
> 
> ```
> ALTER TABLE osm_route_member ADD COLUMN IF NOT EXISTS concurrency_index int;
> 
> INSERT INTO osm_route_member (id, concurrency_index)
>   SELECT
>     id,
>     ROW_NUMBER() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index
>   FROM osm_route_member
>   ON CONFLICT (id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index;
> ```
> 
> What that code does is to add a column to `osm_route_member` (if it doesn't exist) and then populates that column with the concurrency rank of routes associated with a way (useful for when a way is a member of more than one route). The `ON CONFLICT ... DO UPDATE` syntax is equivalent to `REPLACE INTO` or `ON DUPLICATE KEY UPDATE` commands in MySQL, for example.
> 
> This syntax is supposed to be valid as of postgres 9.5.
> 
> Since it works on the dockerized openmaptiles postgres but fails on yours, I suspect the different may be with the version of postgres in use? The postgres in the dockerized container in the distro is reported as:
> `PostgreSQL 9.6.20 on x86_64-pc-linux-gnu (Debian 9.6.20-1.pgdg90+1), compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit`
> 
> Can you check on which version of postgres you're running?

Hi ! I'm having the same problem loading the whole world.
I've checked, and the primary key is a compound of (osm_id, id), and there's no unique index on the `id` column alone, which could explain the raised error. Can you check if you have an index on the `id` column ? related [docs](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT) for quick access :)

It worked on `PostgreSQL 9.6.21 on x86_64-pc-linux-gnu (Debian 9.6.21-1.pgdg90+1), compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit` (openmaptiles/postgis), the "official" docker image.  At first it did not work but I had forgotten the `import-borders` step, and after that, it worked well (yes, this seems completely unrelated...)
but raised the error on a custom built postgis `PostgreSQL 13.3 (Debian 13.3-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit`, however I could not find any change in postgresql related to on conflict which could explain  this behavior change.

proposed fragment do raise the error, however if i add `osm_id` in it, there is no error raised.
```sql
INSERT INTO osm_route_member (id, osm_id, concurrency_index)
  SELECT
    id, osm_id,
    ROW_NUMBER() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index
  FROM osm_route_member
  ON CONFLICT (id, osm_id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index;
```

ZeLonewolf:
@Cactusbone great catch - that index mismatch is definitely a problem.  I'll work on a fix if you don't beat me to it!

ZeLonewolf:
So, investigating this further, in my database, the primary key is the `id` column alone:

```
openmaptiles=# \d osm_route_member
                                       Table "public.osm_route_member"
      Column       |        Type        | Collation | Nullable |                   Default
-------------------+--------------------+-----------+----------+----------------------------------------------
 id                | integer            |           | not null | nextval('osm_route_member_id_seq'::regclass)
 osm_id            | bigint             |           |          |
 member            | bigint             |           |          |
 role              | character varying  |           |          |
 type              | smallint           |           |          |
 ref               | character varying  |           |          |
 network           | character varying  |           |          |
 name              | character varying  |           |          |
 network_type      | route_network_type |           |          |
 concurrency_index | integer            |           |          |
Indexes:
    "osm_route_member_pkey" PRIMARY KEY, btree (id)
    "osm_route_member_member_idx" btree (member)
    "osm_route_member_name_idx" btree (name)
    "osm_route_member_network_idx" btree (network)
    "osm_route_member_network_type_idx" btree (network_type)
    "osm_route_member_osm_id_idx" btree (osm_id)
    "osm_route_member_ref_idx" btree (ref)
Triggers:
    trigger_store_transportation_route_member AFTER INSERT OR DELETE OR UPDATE ON osm_route_member FOR EACH ROW EXECUTE PROCEDURE transportation_name.route_member_store()
```

The proposed fix above gives me the error:
`ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification`

I am not sure why your table has a primary key on both fields, while my table has a primary key on id alone, but it appears to be the route of the problem.  The `osm_route_member` table is generated by imposm.  Is it possible that there's some kind of imposm difference causing the issue?

farfromrefug:
@ZeLonewolf indeed i use my own fork of imposm https://github.com/farfromrefug/imposm3/commits/master
But seeing the commit log i think we use the same version 1.11.1. I just added a commit so support multilinestring

Cactusbone:
```
\d osm_route_member
                                       Table "public.osm_route_member"
      Column       |        Type        | Collation | Nullable |                   Default
-------------------+--------------------+-----------+----------+----------------------------------------------
 id                | integer            |           | not null | nextval('osm_route_member_id_seq'::regclass)
 osm_id            | bigint             |           | not null |
 member            | bigint             |           |          |
 role              | character varying  |           |          |
 type              | smallint           |           |          |
 ref               | character varying  |           |          |
 network           | character varying  |           |          |
 name              | character varying  |           |          |
 network_type      | route_network_type |           |          |
 concurrency_index | integer            |           |          |
Indexes:
    "osm_route_member_pkey" PRIMARY KEY, btree (osm_id, id)
    "osm_route_member_member_idx" btree (member)
    "osm_route_member_name_idx" btree (name)
    "osm_route_member_network_idx" btree (network)
    "osm_route_member_network_type_idx" btree (network_type)
    "osm_route_member_ref_idx" btree (ref)
Triggers:
    trigger_store_transportation_route_member AFTER INSERT OR DELETE OR UPDATE ON osm_route_member FOR EACH ROW EXECUTE FUNCTION transportation_name.route_member_store()
```
so yeah, something's wrong here, I'll try to narrow it down

Cactusbone:
This [commit](https://github.com/omniscale/imposm3/commit/5b0125cb85ad82ff6d2f5786a3e032944a10692e) looks like a good culprit, however it's been around since 0.8.0

Issue #1131, opened 2021-06-29
Unnamed bridges disrupt transportation_name continuity


ZeLonewolf:
Blocks #1128

I discovered this while working on #1128 and felt that it deserves its own discussion and a separate PR focusing just on this bugfix.

Highway shield rendering depends on highway continuity over long stretches.  Adjacent segments of highway with the same network, ref, etc, get coalesced together using an `ST_LineMerge()` function in postgres.  That way, at lower zooms, you still have one continuous segment of highway that the style can use to place shield markers.

However, road continuity is disrupted by bridges or tunnels (`brunnel`, in the schema).  This means that any highway with a non-trivial number of bridges, including tiny unnamed bridges, will cause OMT to chop up the road at each transition between bridge and regular roadway.  Thus at lower zooms, these road segments between bridges quickly become small and suppress `transportation_name` layer labelling.

Example section of I-80 with an unnamed bridge:
![image](https://user-images.githubusercontent.com/3254090/123790979-ebfaf880-d8ac-11eb-8825-917540a710c7.png)

Adjacent section without a bridge:
![image](https://user-images.githubusercontent.com/3254090/123791057-01702280-d8ad-11eb-8bb7-1f52f35ce10b.png)

I propose the following behavior:
- The `brunnel` key will only be set when it has different attributes (such as `name`) from the adjacent stretches of highway.  Unnamed bridges/tunnels should be coalesced into adjacent highway stretches.

This will ensure that distinctly named bridge/tunnel objects will still expose a `brunnel` key.  For example, we would still like to separately label the [Ted Williams Tunnel](https://www.openstreetmap.org/way/290277568) as "Ted Williams Tunnel" rather than "Massachusetts Turnpike"

This behavior can be accomplished by creating a custom aggregation function for bridges/tunnels during the construction of the `transportation_name_linestring` table

TomPohys:
Great catch! The PR is very welcome. Thanks!

zstadler:
Doing spatial analysis, such as "it has different attributes (such as name) from the adjacent stretches of highway" is a complex task.

A different implementation approach is to include two types of objects in the `transportation_name` layer:
1. Merged sections of highway with identical name, ref, network, etc, ignoring their `brunnel` field
2. Merged sections of highway with identical name, ref, network, etc, and identical non-empty `brunnel` value

The style file would show both. Moreover, style authors create separate style layers for the two and give higher priority to names of non-empty `brunnel` highway sections.

[...]

Issue #1128, opened 2021-06-14
Expose concurrent highways in transportation_name layer


ZeLonewolf:
In the US, occasionally highway routes will overlap for short portions (aka, concurrencies).

Here is an extreme example:
![image](https://user-images.githubusercontent.com/3254090/121891170-22e7e080-cce9-11eb-982b-4f06a5bb5285.png)

In US-centric maps, these are often rendered as shield clusters or shields arrange linearly along the route.

OpenMapTiles currently does not expose highway concurrencies in the vector tiles.  Instead, one of the concurrent routes is chosen for the tile.  Instead, it would be useful if each concurrent route had its own separate transportation_name object.  This might look something like this:

![image](https://user-images.githubusercontent.com/3254090/121891565-9ab60b00-cce9-11eb-9b82-305bb43a2968.png)

I propose to modify the SQL in the transportation_name layer which currently suppresses these concurrencies.

Here is an example of rendering exposed concurrencies:
![image](https://user-images.githubusercontent.com/3254090/121891791-ea94d200-cce9-11eb-81e1-2ce791e37c85.png)

Since route concurrencies are relatively uncommon, this should have only a small size increase in tiles which have a concurrency, and will make the SQL that generates transportation_name objects slightly simpler.

ZeLonewolf:
Here is an in-progress demonstration of a concurrency-per-column implementation, which provides values that can be passed to custom highway-shield generators:
![image](https://user-images.githubusercontent.com/3254090/123194818-1bf16880-d475-11eb-93b7-8b83addff529.png)

TomPohys:
Hi, it looks very good. Based on Slack conversation I assumed, that separate attributes for the road concurrencies are the right approach. The NULL columns in the database are OK and the NULL attribute will not be in the vector tiles. 

Did you figure out the styling? 

The PR for this is very welcome. Thanks!

ZeLonewolf:
We have managed to come up with a styling for this, so far this is the most attractive version: 
![image](https://user-images.githubusercontent.com/3254090/123722623-b4f9f800-d856-11eb-9da3-9b08b91acb65.png)

Once maplibre/maplibre-gl-js#188 is resolved, we'll be able to render the shields upright, rather than rotated along the road.

I've been testing the code for this functionality, and it's exposed a few edge cases due to the complexity of the various route networks in the United States, so I'm working through those issues.  I hope to have a PR ready soon.  Thanks for the support!

[...]

PR #1139, opened 2021-07-03
include route relation ref in existence check

### LINKED ISSUE #1138: osm_transportation_name excludes route relation members that lack a ref
The following code in `osm_transportation_name`:

https://github.com/openmaptiles/openmaptiles/blob/1b0b1fd12131167904b61628c1ea403359f4a6e7/layers/transportation_name/update_transportation_name.sql#L48

excludes cases where a way is a member of a route relation, but the way member itself doesn't have the `ref` tag set.  For completeness, these should be included in the merge.

ZeLonewolf:
Fixes #1138
Unblocks #1131
Unblocks #1128 

Below is an example of a way that is part of a route relation (in this case, RI-3), but does not have a ref set on the way itself:
image

This screenshot demonstrates that the `transportation_name` object now contains the ref from the route relation:
image

### DIFF
--- a/layers/transportation_name/update_transportation_name.sql
+++ b/layers/transportation_name/update_transportation_name.sql
@@ -45,7 +45,7 @@ FROM (
         hl.z_order
     FROM osm_highway_linestring hl
             LEFT OUTER JOIN osm_route_member rm ON rm.member = hl.osm_id AND rm.concurrency_index=1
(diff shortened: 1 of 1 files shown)


TomPohys (approved):

TomPohys:
Thanks a lot! It makes sense.

PR #1143, opened 2021-07-06
Simplify Trans-Canada Highway logic

### LINKED ISSUE #1142: Simplify Trans-Canada Highway network code
The following is the detection code for setting the `ca-transcanada` network type:
https://github.com/openmaptiles/openmaptiles/blob/1b0b1fd12131167904b61628c1ea403359f4a6e7/layers/transportation_name/update_route_member.sql#L36-L47

This is the current state of Trans-Canada Highway route relations:
http://overpass-turbo.eu/s/198m

Is is complete across the country, except for a section in Quebec which runs through Val d'Or (route 117).  However, the missing Quebec section is also not present in the OpenMapTiles code above.

I propose that this code be simplified to "any route network that begins with `CA:transcanada`", which would include `CA:transcanada:namedRoute`, which is in use in Ontario.

ZeLonewolf:
Fixes #1142

Screenshot from Alberta showing TCH still shown in `transportation_name` layer:
image

Screenshot from Ontario showing TCH still shown in `transportation_name` layer:
image

### DIFF
--- a/layers/transportation_name/update_route_member.sql
+++ b/layers/transportation_name/update_route_member.sql
@@ -29,22 +29,7 @@ SELECT CASE
            WHEN network = 'US:US' THEN 'us-highway'::route_network_type
            WHEN network LIKE 'US:__' THEN 'us-state'::route_network_type
            -- https://en.wikipedia.org/wiki/Trans-Canada_Highway
-           -- TODO: improve hierarchical queries using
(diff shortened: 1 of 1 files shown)


TomPohys (approved):

TomPohys:
Thanks! Now it is much more simplier!

PR #1144, opened 2021-07-06
fix: prevent crash when re running import-sql

farfromrefug:
Without that change any subsequent call to `make import-sql` would fail because the index already exists.

### DIFF
--- a/layers/building/update_building.sql
+++ b/layers/building/update_building.sql
@@ -95,7 +95,7 @@ CREATE MATERIALIZED VIEW osm_building_block_gen1_dup AS
 SELECT *
 FROM osm_building_block_gen1();
 
-CREATE INDEX ON osm_building_block_gen1_dup USING gist (geometry);
+CREATE INDEX IF NOT EXISTS ON osm_building_block_gen1_dup USING gist (geometry);
 
 -- etldoc: osm_building_polygon -> osm_building_block_gen_z13
 DROP MATERIALIZED VIEW IF EXISTS osm_building_block_gen_z13;


TomPohys:
There is `DROP MATATERIALIZED VIEW` before creating new one, so index should be dropped too (tested many times on my local machine, with all layers, or with just `building layer`).  And as you write on Slack, if there was an error which raise `pg_type_typname_nsp_index`, there is maybe some flaw with parallel procesing of sql. 

On the other hand, you do not want to end up with wrong (old) geometry index.

farfromrefug:
@TomPohys Well i am not an expert in SQL. i understand what you say but in the meantime we do it that way in every other place like here for example https://github.com/openmaptiles/openmaptiles/blob/master/layers/boundary/boundary.sql#L23.

But you are right it might be some parallel issue. But then i dont know how to fix it

TomPohys:
Could you please provide some more information about when, which area this `pg_type_typname_nsp_index` error raise? Because with `IF NOT EXISTS` you fix the consequence, not the cause (why the `osm_building_block_gen1_dup` was not drop with its index?)

But I understand that this `IF NOT EXISTS` are in many places, where should not be because there is a `DROP` before.

TomPohys:
I am sorry that you have still trouble with this. Maybe there could be a problem with the parallel processing of buildings (just a guess).

nyurik:
could be easily tested by setting `MAX_PARALLEL_PSQL = 1` in .env

farfromrefug:
@nyurik good idea. Honestly did not tried this. would have to test it. No big deal tbh when you about it. I re run quickstart everytime.

PR #1147, opened 2021-07-07
Remove unused osm_id column

### LINKED ISSUE #1146: Remove unused osm_id column in transportation_name layer
The definition for the `transportation_name_linestring` table includes an osm_id column which is not used (set to null in all cases).

https://github.com/openmaptiles/openmaptiles/blob/4eb240466e1e555dd39b9d627ee357b7ad7ba4bf/layers/transportation_name/update_transportation_name.sql#L57-L59

This null column is perpetuated all the way up to the layer definition:

https://github.com/openmaptiles/openmaptiles/blob/4eb240466e1e555dd39b9d627ee357b7ad7ba4bf/layers/transportation_name/transportation_name.sql#L4-L9

However, the layer does not select from this (null) column:

https://github.com/openmaptiles/openmaptiles/blob/4eb240466e1e555dd39b9d627ee357b7ad7ba4bf/layers/transportation_name/transportation_name.yaml#L96-L99

I propose that this unused column can be removed, which will save space in the database and also make it easier for future developers to understand the code.  Since `transportation_name` objects used views with merged and generalized geometries, there is no possibility to populated osm ids.

ZeLonewolf:
Fixes #1146

This PR removes the always-null osm_id column from the `transportation_name_linestring` table and the series of generalized tables that derive from it.

Demonstration of `transportation_name` objects behaving normally after the column has been removed:

image

### DIFF
--- a/layers/transportation_name/transportation_name.sql
+++ b/layers/transportation_name/transportation_name.sql
@@ -4,7 +4,6 @@
 CREATE OR REPLACE FUNCTION layer_transportation_name(bbox geometry, zoom_level integer)
     RETURNS TABLE
             (
-                osm_id     bigint,
                 geometry   geometry,
                 name       text,
                 name_en    text,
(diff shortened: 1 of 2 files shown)


TomPohys (approved):

TomPohys:
Thanks, the empty `osm_id` is really unnecessary.

PR #1149, opened 2021-07-12
Transportation ref layer

### LINKED ISSUE #1140: Create a transportation_ref layer
This proposes to create a new layer, `transportation_ref`, which contains data used for highway route and shield labeling.

This issue is based on discussion in #1131 which identified that intermingling route and name information was becoming unwieldy to use and document in the `transportation_name` layer.  In addition, since highway route information doesn't have a name component, it didn't really fit in `transportation_name`.

A similar issue, #1128 proposes to expose highway concurrencies in the `transportation_name` layer.  This can still be done, but is probably not needed if a separate `transportation_ref` layer exists.

This proposed layer would contain the following information:
* `geometry` - a `LineString` containing the zoom-generalized geometry of the route being labeled
* `highway` - highway classification
* `network_1` - network name of the most important route concurrency (e.g. "US:I")
* `ref_1` - ref of the most important route concurrency (e.g. "95")
* `network_N` - network name of the Nth most important route concurrency
* `ref_N` - ref of the Nth most important route concurrency

Based on testing in the US and Slack discussions, a reasonable maximum N is 6 simultaneous concurrencies.  Concurrency importance is based on the algorithm described in #1134 and implemented in #1135.

ZeLonewolf:
Closes #1140

This is an alternative to the approach described in #1128, which would modify the `transportation_name` layer rather than create a new layer.  The approach of creating a new layer was discussed in #1131.

This PR implements a new layer, `transportation_name`, which renders contiguous features with a common route and route concurrencies, with support for up to 6 concurrent routes on a way.  This layer groups features with a common highway class, without regard for road names or bridges when merging geometries.

This layer implements the following fields:
`class` and `subclass`, which match the `transportation` / `transportation_name` usage
`network`, which matches the `transportation_name` usage (for compatibility and common filtering across layers)
`network_1` through `network_6` which is the route network of the 1st through 6th route concurrency
`ref_1` through `ref_6`, which is the route number of the 1st through 6th route concurrency

This draft is offered for initial feedback.  Additional work is still needed to compose samples/screen shots and documentation updates to the openmaptiles.com web site.


ZeLonewolf:
Closed in favor of adding this information to the `transportation_name` layer.

PR #1154, opened 2021-07-19
Fix SQL failures in transportation_name update code

ZeLonewolf:
This PR fixed bugs introduced in #1147 and #1119 which broke the update triggers in `transportation_name_update`.  I noticed this issue while I was working a different PR and tried to update a table in the `transportation` layer.

**Test**

The following code currently fails because of cascading update failures, but will complete successfully after this PR is merged.
```
UPDATE osm_highway_linestring hl
    SET network = rm.network_type
    FROM osm_route_member rm
    WHERE hl.osm_id=rm.member;
```


TomPohys (approved):

TomPohys:
Thanks!

PR #1157, opened 2021-07-21
Remove useless null osm_id from transportation layer

ZeLonewolf:
This issue is similar to the fix in #1147, which removed null osm_id columns from the `transportation_name` layer.  This PR does the same cleanup in the `transportation` which similarly had unused null osm_id columns that were not used in the layer.

Below is a rendering of Rhode Island, USA that I generated after removing the null columns:
image
image

While I was in this file, I also removed a few unneeded spaces at the ends of lines.

### DIFF
--- a/layers/transportation/transportation.sql
+++ b/layers/transportation/transportation.sql
@@ -11,7 +11,6 @@ $$ LANGUAGE SQL IMMUTABLE
 CREATE OR REPLACE FUNCTION layer_transportation(bbox geometry, zoom_level int)
     RETURNS TABLE
             (
-                osm_id    bigint,
                 geometry  geometry,
                 class     text,
                 subclass  text,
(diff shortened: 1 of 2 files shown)


TomPohys:
Hi, in `transportation_name` the removal `osm_id` makes more sense, because it is using layer `osm_transportation_name_linestring` and `osm_highway_point` which, for linestring, was `osm_id` always `NULL`, and for `motorway_junction` it is OK, if there is not any case of usage.

But in `transportation` layer, from z4 to z11 is used `osm_transportation_merge_linestring_gen_zX` which does not have `osm_id`, that is true. But from z12 `osm_highway_linestring` uses `osm_id`. Other layers like `osm_railway_linestring`, `osm_aerialway_linestring`, `osm_shipway_linestring` and `osm_highway_polygon` are using `osm_id` too.

Nevertheless, the `layer_transportation` function returns `osm_id`, and `query` function does not (https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/transportation.yaml#L177).

I would prefer to keep the possibility to add the `osm_id`  to `query` function, so it could be used in vector tiles.

PR #1158, opened 2021-07-21
Add network to transport layer

### LINKED ISSUE #1153: Expose network_type in the transportation layer
Currently, the field `network` is available in the  [`transportation_name`](https://openmaptiles.org/schema/#transportation_name) layer.  The field provides the type of network associated with a road (for example, us-interstate, ca-transcanada, etc).  However, this field is not available in the `transportation` network.  That means that the network type can be used for labelling, however, it cannot be used for styling of the roadway itself, or for determining at which zoom level a road should appear.

In our under-development [americana style](https://github.com/ZeLonewolf/openstreetmap-americana) ([live map](https://zelonewolf.github.io/openstreetmap-americana/#6/40.65/-75.681)), we would like to distinguish Interstate highways from other types of motorways (such as state or local highways which are also tagged `highway=motorway`.  The reason for this, is that Interstate highways form a meaningful, contiguous road network, while "all motorways" includes many "motorway islands" and fragments, when rendered separately from other classes of road.  An example of this can be seen here:

image

Since `highway=motorway` is a physical classification, and `highway=trunk` is a functional classification, they need to be rendered at the same time.  However, these two road classification together form a denser network of roads than we would like to show at the lowest highway-visible zoom level.

In order to solve this problem, we would like to be able to use the following zoom logic:
1. At the lowest zoom in which roads are shown, show only interstate highways (`network=us-interstate`) or equivalent "top level" highways in countries where we have this defined in OpenMapTiles.
2. At some higher zoom, show all `highway=motorway` + `highway=trunk`.

This scheme would allow us to show a reasonable connected highway transportation network at lower zoom level, while introducing the remainder of the motorway features at the same time that all trunk roads are shown.

In order to accomplish this, it is proposed to move layer/transportation_name/update_route_member.sql into the `transportation` layer, and add an additional column to expose `network` on `transportation` features, so they can be used for zoom-level determinations.

ZeLonewolf:
Closes #1153 

This PR populates the existing `network_type` field in the route table with highway route network information, using the most important value in cases of concurrencies.  In order to expose this value, two files `network_type.sql` and `update_route_member.sql` were moved from the `transportation_name` layer to the `transportation` layer.

Below is a representative zoom level 6 with motorways only shown for `us-interstate` network types.  This sample was generated from this PR using a [custom branch](https://github.com/ZeLonewolf/openstreetmap-americana/tree/selective-highway-zoom) of openstreetmap-americana.  This provides a simplified rendering of the highway network which excludes many minor motorway roads and sections.
image

Below is zoom level 6 on the current [openstreetmap-americana](https://zelonewolf.github.io/openstreetmap-americana/#6/42.148/-73.712).  Notice the many additional stubs, motorway islands, and minor routes which are present.  Instead of rendering `highway=trunk` to make this network look nice, we can instead suppress non-interstate roads and withhold motorway+trunk rendering to a closer zoom:
image


TomPohys:
Thanks a lot. It makes sense to filter not just `name` but also data for lower zooms.

PR #1160, opened 2021-07-26
Implement park rendering at z4-5

ZeLonewolf:
Unblocks ZeLonewolf/openstreetmap-americana#51

Currently, OpenMapTiles renders protected areas at zoom 6 and higher.  However, the western United States has protected areas that are so vast that it makes sense to render these areas as low as zoom 4, as noted in ZeLonewolf/openstreetmap-americana#51

This PR extends the generalization of park features from its current termination at zoom 6 to zoom 4.

Here is an example from the Americana style of rendering protected areas at zoom 6, which is currently the lowest supported zoom:
https://zelonewolf.github.io/openstreetmap-americana/#6/45.313/-111.721

We would like to provide rendering at zoom 4 (no labels) and zoom 5 (with labels) in the Americana style.  Below are samples generated from this PR for Idaho, USA:

Zoom 4:
image

Zoom 5:
image


TomPohys:
Hi @ZeLonewolf, thanks for the PR. I did a quick test on the Czech Republic and it added parks too (not just for the US), which are now from zoom 6. I have to do more tests in Europe (for a bigger context), but I need a little more time.

ZeLonewolf:
Thanks for the update @TomPohys, and as always, I'm happy to offer revisions if needed.  Note that I fixed the tolerance typo at zoom 6, which was set one level too high, and is the reason for the -0.3% reduction in tile size at that zoom, so you might want to look closely at that zoom to make sure you're happy with how it looks.

ZeLonewolf:
Please re-run CI.  It looks like there was an error downloading an extract during the run.

ZeLonewolf:
> Hi @ZeLonewolf, sorry for the long time waiting. Would it be a blocker if the `parks` will be extended just into zoom 5? or it is vital to have them in zoom 4?

@TomPohys I think, strictly speaking, its probably not a blocker to not have them in zoom 4, but it is a strong nice-to-have for the largest areas -- particularly areas in the size of > 5,000 km².  Here I am thinking of places like Yellowstone, Denali, Arctic National Wildlife Refuge, Tongass National Forest, or Wood Buffalo in Canada.

Further to this point, I think that we can increase the generalization filter for zoom 5, as it is really not useful at that zoom to render tiny parks that are only a few pixels in size.

Would it be acceptable to offer a revision to this PR that reduces the impact on tile size by decreasing the level of detail in the park layer at z5, while still including the largest parks in z4 at a low level of detail?  I think we can leverage @farfromrefug's work on #1164 as part of the solution on how to accomplish this.

TomPohys:
Thanks a lot, it looks OK. Just one question. Is the `geometry_point` necessary for zoom 4? Could be this functionally be removed? https://github.com/openmaptiles/openmaptiles/pull/1160/files#diff-39456d58366bd6e9be471d48a09732710112d0c5a9e801a949268cbd061bc159R21-R22

ZeLonewolf:
Thanks for the review.  The point features are not necessary, so I've removed everything related to z4 centroid points from this PR.

TomPohys (approved):

TomPohys:
Thank you very much! It looks great!

PR #1162, opened 2021-07-29
Render ramps up to zoom 7

### LINKED ISSUE #1161: Ramp rendering at zoom 7-8
Fixes ZeLonewolf/openstreetmap-americana#53

Below zoom 9, `highway=motorway_link`, `highway=trunk_link`, and `highway=primary_link` are no longer rendered.  This causes gaps in the road network for those road classes at zooms as low as 7-8.  An example of this at zoom 8.95 in the Americana style is shown here:

image

At this zoom, most of these links will be too small to be visible, however, longer links currently leave gaps as shown.

The code that strips out these links is here:

https://github.com/openmaptiles/openmaptiles/blob/9e4be3e3b03c26fcc1e63ee4976f44a5d1f8dba1/layers/transportation/update_transportation_merge.sql#L114-L123

I propose one of the following two solutions to this:
1. Extend rendering of link roads for these three highway classes up to zoom 7
2. Convert links to their parent class types and merge the geometry

ZeLonewolf:
Fixes #1161

This PR extends ramp rendering up to zoom 7 (from the current zoom 9) and makes associated optimizations.

Screenshot of ramp rendering at zoom 8:
image

Current ramp rendering at zoom 8:
image

The approach is as follows:
* Link roads are added to the where clause in `osm_transportation_merge_linestring`
* `WHERE` clauses in the z7 and z8 merge tables are removed as unnecessary because motorway/trunk/primary filtering is already done in the underlying `osm_transportation_merge_linestring` table
* The existing `WHERE` clause in the z6 merge table will drop primary and all ramp types
* An `ST_Length()` filter is applied to the z8 merge table in order to filter out smaller link road segments that won't be visible at this zoom.

### DIFF
--- a/layers/transportation/update_transportation_merge.sql
+++ b/layers/transportation/update_transportation_merge.sql
@@ -119,8 +119,8 @@ FROM (
                 is_ford,
                 min(z_order) AS z_order
          FROM osm_highway_linestring
-         WHERE (highway IN ('motorway', 'trunk', 'primary') OR
-                highway = 'construction' AND construction IN ('motorway', 'trunk', 'primary'))
+         WHERE (highway IN ('motorway', 'trunk', 'primary', 'motorway_link', 'trunk_link', 'primary_link') OR
+                highway = 'construction' AND construction IN ('motorway', 'trunk', 'primary', 'motorway_link', 'trunk_link', 'primary_link'))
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
I no longer believe this is necessary.

PR #1163, opened 2021-07-29
Generalize ramps at zoom 7-8

ZeLonewolf:
This is an alternate approach to #1162 for consideration.

In this version of the PR, ramps are generalized into their parent roadways at zoom 8 and above.  I'm offering this as a comparison in order to compare performance and tile size.

Below is a render of the same test area which shows no unusual geometry artifacts at zoom 8:
image


ZeLonewolf:
Didn't actually work.

PR #1165, opened 2021-07-30
fix: move cache to a docker volume

farfromrefug:
As explained in slack, on macos the current build system is extremely slow. The reason is because the cache folder is a "shared/local" folder. This PR moves it to a classic docker volume


nyurik:
this PR goes much further than just changing the cache folder. It reverts back to running docker as root (not good), and makes other unrelated env var changes. I might be misunderstanding the issue, but if you share a `data` dir as a volume with docker, the cache is a subdir, isn't it?

farfromrefug:
Ok so running as root is necessary if using docker volumes and not "shared" volumes
The core of the issue is that "shared" folder is extremly slow on macos (known issue).
Plus there isnt really a reason (that i know of) to have the cache volume as a "shared" volume.
I did the same thing for the "export" folder (where the mbtiles is generated) because of the same reason. However the issue was far less obvious with the mbtiles generation.

Now about changing env variables. What are you seeing ? I have changed only one variable (i think) to remove local user (which wont work in true docker volume).

The idea of this PRs was first to have some people try it on windows/linux as i cant test it. 

On macos the gain is more than obvious. The full processs is more that 12 times faster

ZeLonewolf:
I am on Windows (with WSL), can you describe what test you would like to see run?

farfromrefug:
@ZeLonewolf A quickstart running correctly from start to finish is enough. If we want to push this to the max, comparing build time on master vs this would also be nice.
Thanks

TomPohys:
Hi @farfromrefug, what are the next steps with this PR? Thank you

farfromrefug:
@TomPohys It would need to be fixed on Windows. Or maybe makes this be used only on macOS. Though not sure if possible.
Right now dont have much time to do this. If you prefer to close this and reopen if necessary dont hesitate.

ZeLonewolf:
I think it would be a good idea to post an issue describing the the problem that you're having with performance on MacOS, so that the issue is not lost.

TomPohys:
Hi @farfromrefug, let's close this PR and if there is still a problem with macOS please open an issue (as ZeLenewolf also suggested) and you can link this PR.

But thanks anyway for this PR!

PR #1166, opened 2021-08-01
Update the description of the water layer's class

zstadler:
Clarify that OSM water bodies that do not have a `waterway` tag are classified as `lake`.

This refers to the following tags:
https://github.com/openmaptiles/openmaptiles/blob/9e4be3e3b03c26fcc1e63ee4976f44a5d1f8dba1/layers/water/mapping.yaml#L86-L95


TomPohys (approved):

TomPohys:
Thanks @zstadler for the doc update!

PR #1167, opened 2021-08-01
Add spring water bodies

zstadler:
The [OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dspring) says `natural=spring` may be used on areas.

image


TomPohys:
Hi @zstadler, in the schema is used only [`amenity=drinking_water`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddrinking_water), where some `natural=springs` are also included, but not all of them. 

I found a place where are all of them (spring as point and as way) https://overpass-turbo.eu/s/1abU 

Isn't this more into Outdoor?

zstadler:
Hi @TomPohys,

I don't think I fully understand your comment. In particular, I don't see any reference to `amenity=drinking_water` in the `layers/water` code or in the layer's [schema documentation](https://openmaptiles.org/schema/#water)

This PR is about the water bodies/areas/polygons layer. It does not change anything regarding nodes/points.

TomPohys:
The 'drinking_water' is part of the 'poi' layer. Some 'natural=springs' have the tag 'amenity=drinking_water', so some point springs are in the schema as poi points. 

Nevertheless, should be the springs (and only as polygons, which is like 1% of all springs) in the base map?  But I am keen to discuss. 

There are also some spring areas that have tag 'building' like https://overpass-turbo.eu/s/1ahS (https://www.maptiler.com/maps/#topo//vector/18.25/8.044398/47.838762)

zstadler:
Thanks @TomPohys for the clarification.

I don't see a reason why a `natural=spring` water body should not be in the schema when a similar water body with an in-flowing stream is included. 

I also don't see a reason why a `leisure=swimming_pool`, with or without a `building` tag, should not be in the schema while a similar water body with a natural water source would not be included. 

P.S., 
Please see the difference in the way the maptiler vector and raster maps render [a `leisure=swimming_pool` area with a `building=yes` tag](https://www.openstreetmap.org/way/301526861) at [this location](https://www.maptiler.com/maps/#topo//vector/19/8.14972/47.79485)

zstadler:
@TomPohys,

Would it be possible to merge this PR? Up till now I've merged `master` into this branch 6 times.

BTW, Regarding the "schema-change" label:
This PR does not change the schema fields. It only adds some OSM elements under the existing schema and the existing `class=lake`.

TomPohys:
I am sorry, I will merge it now. You are right, it is not schema change.

PR #1168, opened 2021-08-04
Remove useless WHERE clauses from the transportation layer

ZeLonewolf:
This PR removes several redundant/unnecessary WHERE clauses from the transportation layer.  Specifically:

The table `osm_transportation_merge_linestring` is a view of `osm_highway_linestring` which exposes only motorway/trunk/primary roads:
https://github.com/openmaptiles/openmaptiles/blob/9e4be3e3b03c26fcc1e63ee4976f44a5d1f8dba1/layers/transportation/update_transportation_merge.sql#L103-L123

However, the create statement for the table `osm_transportation_merge_linestring_gen_z8`, which is a view of `osm_transportation_merge_linestring`, also contains  a `WHERE` clause which selects down to motorway/trunk/primary roads.  This `WHERE` is unnecessary:
https://github.com/openmaptiles/openmaptiles/blob/9e4be3e3b03c26fcc1e63ee4976f44a5d1f8dba1/layers/transportation/update_transportation_merge.sql#L144-L145

This unneeded `WHERE` clause is similarly present in the create statement for `osm_transportation_merge_linestring_gen_z7`, which is a view of `osm_transportation_merge_linestring_gen_z8`:
https://github.com/openmaptiles/openmaptiles/blob/9e4be3e3b03c26fcc1e63ee4976f44a5d1f8dba1/layers/transportation/update_transportation_merge.sql#L163-L164

Likewise, there is a similar redundant `WHERE` clause in the `osm_transportation_merge_linestring_gen_z5` and `osm_transportation_merge_linestring_gen_z6` tables, both of which select down to `motorway` and `trunk`.  This `WHERE` clause is only needed on the z6 table, and is redundant on the z5 table.

I am not sure what the performance penalty is for these redundant `WHERE` clauses, but there does not appear to be any reason to keep them, and they may incur a performance cost.

### DIFF
--- a/layers/transportation/update_transportation_merge.sql
+++ b/layers/transportation/update_transportation_merge.sql
@@ -141,8 +141,6 @@ SELECT ST_Simplify(geometry, ZRes(10)) AS geometry,
        is_ford,
        z_order
 FROM osm_transportation_merge_linestring
-WHERE highway IN ('motorway', 'trunk', 'primary')
-   OR highway = 'construction' AND construction IN ('motorway', 'trunk', 'primary')
     ) /* DELAY_MATERIALIZED_VIEW_CREATION */;
 CREATE INDEX IF NOT EXISTS osm_transportation_merge_linestring_gen_z8_geometry_idx
     ON osm_transportation_merge_linestring_gen_z8 USING gist (geometry);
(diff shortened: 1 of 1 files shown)


nyurik:
i'm all for optimizations like this. One thing we should definitely do is record intentions/assumptions as comments - so perhaps it would be good to add these where clauses as a comment, either verbatim, as a WHERE clause with a comment that these are already implemented somewhere else, or as text specifying what each SQL statement assumes

ZeLonewolf:
Good idea.  I just added comments to document the implicit WHERE clauses.

TomPohys (approved):

TomPohys:
Thanks! It looks much cleaner!

PR #1169, opened 2021-08-08
Add long-distance hiking networks to transportation and transportation_name

ZeLonewolf:
This PR add hiking network information to the `transportation` and `transportation_name` layer.  Hiking network information is important because it allows styles to emphasize the most important hiking trails, either by selective rendering or differentiated styles.  This information is important to the Americana style, in order to be able to render important long-distance trails such as the [Appalachian Trail](https://en.wikipedia.org/wiki/Appalachian_Trail) or the [Pacific Crest Trail] (https://en.wikipedia.org/wiki/Pacific_Crest_Trail) at appropriate zoom levels.

Additionally, for the most important hiking networks, it is useful to be able to render these trails at lower zooms.  The current behavior is for all paths to be rendered starting at zoom 14.  This PR implements the following:
* international and national hiking networks are rendered at zoom 12+
* regional hiking networks are rendered at zoom 13+
* all other paths maintain the current behavior of rendering at zoom 14+

This PR parses the [`route=hiking`](https://wiki.openstreetmap.org/wiki/Tag:route%3Dhiking) tag, and the `iwn`, `nwn`, `rwn`, and `lwn` hiking network tags.  Four new corresponding network_type enums are created:
* hiking-international
* hiking-national
* hiking-regional
* hiking-local

Below is an example of differentiated styling from the Americana style.
image


farfromrefug:
I think those hiking routes should be in a different layer. Most of us use a fork of openmaptiles and not all of us need this. Plus some (like me) already have this in a different layer. 
Having this in transportation will make it very hard to keep forks which dont want it in think with this repo.

zstadler:
I also use a separate `trails` layer for `hiking`, `bicycle`, and `mtb` routes. This layer also includes fields like `osmc_symbol`, `colour`, and `way_colour` that enable the coloring of the route's lines in addition for `name*` and `ref` labels that would be enabled by this PR.

farfromrefug:
@zstadler can we see your layer? Would love to se how you did it

ZeLonewolf:
I'm not sure why there would be a problem populating the existing fields _which already exist in these layers_ with hiking route information, as it seems this is still of interest in a general-purpose style.  It doesn't make sense to require a fork for this.  After all, I'm not proposing to add new fields to the schema.

It seems that there are several approaches that have been done for hiking data (including the original one presented at SOTM a few years back), and it would be a shame to have to create another fork to yet again include off-road data in a style.

frodrigo:
An other approach I try is too add only new reusable layer rather than fork the whole.

Look at
https://github.com/openmaptiles/openmaptiles/issues/970
https://github.com/makina-maps/openmaptiles-contrib-bicycle_hiking_route
https://github.com/makina-maps/osm-bright-bicycle-gl-style

Contrib welcome.

farfromrefug:
@ZeLonewolf if i were to use your changes it would make my tiles unnecessarily bigger. I really think hiking/biking routes etc... should not be in this layer

farfromrefug:
@frodrigo iam not sure i understand how your approach is different from defining a new layer ?

frodrigo:
> @frodrigo iam not sure i understand how your approach is different from defining a new layer ?

Sure, it is new layers, but outside of openmaptile project.

farfromrefug:
@frodrigo interesting idea! I really like it

ZeLonewolf:
> @ZeLonewolf if i were to use your changes it would make my tiles unnecessarily bigger.

That's not a great rationale.  Are you going to object to any and all additions to OpenMapTiles simply because your particular use case won't use it?  There is certainly data in the tiles today that my style doesn't and won't use (for example, lower zoom buildings) that I would remove if/when optimizing the tiles for size for my specific use case.

> Having this in transportation will make it very hard to keep forks

You shouldn't be maintaining a fork of the `transportation` layer -- you should define a new layer that selects just the data you need from the existing `transportation` layer.  That way you can filter by zoom, choose only the fields you need, or even add additional fields.  Otherwise you're going to constantly have to deal with merge conflicts on every change to OpenMapTiles in that layer.

> I think those hiking routes should be in a different layer.

We added comprehensive support for highway routes in openmaptiles/openmaptiles#1152.  Since we have everything from `path` through `motorway` in the `transportation` and `transportation_name` layer, I don't see why populating routes that apply to paths would be out of scope for those layers.  If we're supporting routes, then let's support routes.  This is basic information that a general-purpose (i.e. not hiking-specific) style could use.  I did attempt to put route information into a separate `transportation_ref` as suggested by @zstadler in https://github.com/openmaptiles/openmaptiles/issues/1131#issuecomment-873475296 but there did not appear to be support for that approach, and thus we put it in `transportation_name`.

I am wondering if there a general objection to including hiking information in OpenMapTiles and it's not clear to me what's "in scope" or "out of scope" for the project.  In general I'm in favor of contributing back to the project as much as possible, as I think it's important to the community as a whole to have a good, general-purpose vector schema that everyone can use.  It seems like everyone working with hike/bike layers is off doing their own custom thing.  I'm perfectly happy to run off and make my own custom layers for my own use case, or borrow from some of the excellent approaches already done, but I'm only going to do that in cases where OpenMapTiles won't accept adding it to the main, community schema.

farfromrefug:
@ZeLonewolf dont take what i said the wrong way. The true reasonning for me is that putting everything in transportation makes jt harder for all of us, who are forking openmaptiles, to filter what we truly need.  
And the point with tile sizes was that the consequence of putting everything in transportation would be bigger tiles for everyone who did not take the time to go through transportation sql and clean it up( and it is not easy to maintain a fork with changes in sql files).
That s  it. I should not have said "my" and i apologize

ZeLonewolf:
@farfromrefug there is no need to apologize, I am hoping to come to a common understanding that can help the project move forward and learn what is and is not in scope.

> the consequence of putting everything in transportation would be bigger tiles for everyone who did not take the time to go through transportation sql and clean it up( and it is not easy to maintain a fork with changes in sql files).

I understand what you're saying.  If someone has forked OpenMapTiles _and modified the `transportation` layer in their fork_, then, I agree that this approach will be difficult to maintain.

I am suggesting that anyone that wants to have a modified version of the `transportation` layer (or any layer), can simply create an entirely new layer which reads from the same tables created by the `transportation` layer but with custom `WHERE` clauses that select _only_ the information that the user of the modified `transportation` layer cares about.  By taking that approach, someone with a fork of OpenMapTiles won't care what gets piled into the `transportation` layer, because their custom layer will define what data is extracted from `transportation`.

Are there other forks that are taking the approach of maintaining modified versions of the main layers rather than creating extensions?

farfromrefug:
But what you suggest means my transportation layer would have another name. But your idea is what i was thinking about all along. Your proposed routes for hiking / biking should be done exactly that way. Why not do it the other way around ;) it would answer your need and would not change anything for others, dont you think ?

frodrigo:
> Are there other forks that are taking the approach of maintaining modified versions of the main layers rather than creating extensions?

As I try to use new layers for new features, I also have forks for modified layer. Edit the layer in fork allow to integrate improvement from openmaptiles, and when possible propose patch to improve it base.

The model forks and contrib layers have to be more brain stormed, and documented.

ZeLonewolf:
> But what you suggest means my transportation layer would have another name. But your idea is what i was thinking about all along. Your proposed routes for hiking / biking should be done exactly that way. Why not do it the other way around ;) it would answer your need and would not change anything for others, dont you think ?

Of course, that is what I would have to do if my contributions are not accepted.  But the whole point of contributing back is to offer back useful functionality that's of general interest.

I note @frodrigo's open #1003 regarding contrib/addon layers seems to have not had any activity is nearly a year.  I'd be happy to help move that along as it's something that I'm going to have to take on at some point (regardless of the outcome of this hiking discussion).

ZeLonewolf:
Closing in order to take a different approach.

PR #1172, opened 2021-08-11
Transportation generalization table optimization and cleanup

ZeLonewolf:
This PR updates the `transportation` layer creation scripts to simplify the SQL and remove unneeded sub-selects, checks and conditionals, fix indexes, and improve inline documentation. 

Currently, there are two sequences of materialized view creations.  There is one from zoom 11 through 9, and a second one for zoom 8 through 4.  This PR removes that break in the sequence of transportation table materialized view creations, in favor of one in which high-zoom views are created first, and then each lower zoom is created from the zoom above.

Instead, the current generalized zoom 8 transportation table is built directly from `osm_transportation_linestring` rather than being built from the zoom 9 transportation table.  This means that when building the zoom 8 table, it must scan the entire transportation network rather than just selecting from the pre-filtered zoom 9 table.

This PR removes an unneeded sub-select in the build of the zoom 8 table, which appears to be a leftover from an old version of the SQL that did some sort of merge.  Once this PR is implemented all zooms from 11 through 4 will be linked via a progressive series of materialized views.

Lastly, this adds in missing materialized view refreshes for zooms 9-11, which appear to have been entirely missing, and as far as I can tell aren't getting updated in the current version of this code.

In addition, the following optimizations were added as part of this commit:

1. Updates the `osm_highway_linestring_highway_partial_idx` partial index to match the `SELECT..WHERE` clause actually present in `transportation/update_route_member.sql`, which is where it appears to be actually used, and update inline documentation to reflect this.

2. Collapses unnecessary sub-select block in `osm_transportation_merge_linestring_gen_z11`, and removes unnecessary ST_IsValid() call, which already provided in `mapping.yaml`, and update inline documentation to reflect these assumptions.

3. Updates `WHERE` blocks to remove unnecesary checks and further document assumptions.  The `highway=construction` check is unnecessary in all cases, because it is sufficient to check the `construction` key alone.  If `construction=*` is set, then `highway=construction` is implied.

4. Two indexes were added to `layers/transportation/update_route_member.sql` to improve route population performance.

In testing locally, I'm seeing performance improvements around 10% in the generation of the `transportation` layer, based on modifying `openmaptiles.yaml` to generate only the transportation layer and then repeatedly running `time make import-sql`, however, this timing might be impacted by docker, so I would ask for confirmation of acceptable performance.

In addition, this PR shortens the length of the transportation update SQL file by 30 lines, which makes it easier for contributors to work with.


ZeLonewolf:
Screen shot of the transportation layer still being generated normally:
image

TomPohys:
Great work! it looks awesome! Thanks for this update and clean-up!

TomPohys (approved):

PR #1174, opened 2021-08-14
Implement private road tagging

### LINKED ISSUE #1066: Filtering highways with access: no attribute
OpenMapTiles includes highways that have "access: no" attribute. The scheme omits this attribute and as the result private highways get rendered as normal highways and there is no way to filter them out with style changes.

What would be the best way to omit these highways?

ZeLonewolf:
Fixes #1066 

This PR adds a new field `access` in the transportation layer, which will be set to `no` if the `access` tag is either `no` or `private`.  While `private` is the more popular value by a 17:1 ratio, I went with `no` because it's smaller in the tiles.  In addition, the text `no` opens up the future possibility of other text-based access values such as `destination`, `customers`, or `permit`.

The screenshot below shows an example of access tagging for a road on a military base:

image


TomPohys:
Awesome! Thanks, it looks great!

TomPohys (approved):

PR #1175, opened 2021-08-14
Add support for cemetery tagged as grave_yard

### LINKED ISSUE #1057: Cemetery tagged as amenity=grave_yard are missing
`amenity=grave_yard` are missing in tiles, only the POI is present.

Example : [way 31855109](https://www.openstreetmap.org/way/31855109)
See in [Maputnik](https://maputnik.github.io/editor/#17/47.72471/-3.33599)

ZeLonewolf:
Fixes #1057

This PR adds `amenity=grave_yard` to the `landuse` layer.  A unification function was implemented which encodes all `class=grave_yard` as `class=cemetery` in the tiles, which adds these features for existing users of `class=cemetery` with no change.  The unification function can serve as a basis for any other tags that we might want to unify in the `landuse` layer.

Tile rendering for an `amenity=grave_yard`.  ([Location](https://www.openstreetmap.org/way/857383420))
image

Tile rendering for a `landuse=cemetery`.  ([Location](https://www.openstreetmap.org/way/385779531))
image


TomPohys:
Thanks for PR! It looks great! 

Maybe one suggestion, idea. Would it be possible to move the `landuse_unify` function into a separate SQL file e.g. `class.sql` like it is in the `poi` or `transportation` layer? Maybe it will be more visible. But it is just an idea.

ZeLonewolf:
@TomPohys great idea!  I made the change.  Then, in the future, if there are other types that need to be combined, they can all go into one place.

TomPohys (approved):

TomPohys:
Thanks a lot!

PR #1176, opened 2021-08-14
Implement feature IDs in aerodrome_label and mountain_peak

### LINKED ISSUE #804: Osm_id hided from tiles in mountain_peak and aerodrome
This PR https://github.com/openmaptiles/openmaptiles/pull/686 hide osm_id in generated tiles. 

osm_id should be available. It will also help #792  

Should be `key_field_as_attribute: true`

ZeLonewolf:
Fixes #804 

This PR adds a feature ID (`id` field in mvt) for point objects in the `aerodrome_label` and `mountain_peak` layers.  Since these layers are entirely node-based, the OSM id is used for the feature ID.  Note that we would need a different approach for generating **unique** feature ids on layers that have a mix of areas/lines/nodes, but this approach should be fine for point layers.

Feature id for an `aerodrome_label` object:
image

Feature id for a `mountain_peak` object:
image


TomPohys:
Thanks for this PR! I just looked at how to get `featureID` from vector tiles and for `mountain_peak` there is a `featureID` for each peak that is possible to get, same as for `POI` layer. Nevertheless, I have a problem with `aerodrome_label`, that aerodrome centroids should keep `osm_id` but `osm_id` not bubble into vector tiles.

ZeLonewolf:
@TomPohys thanks for the review.  If I understand your comment correctly, it sounds like we want to have the feature ID in the tile but still retain the osm_id in the table returned to the layer.

TomPohys:
I am sorry that my previous text was not written clearly enough. nyurik added the `key_field` function (https://github.com/openmaptiles/openmaptiles-tools/commit/cf4257c75e4e845864d809ef018a035c0a7c6b44) where the input is a positive integer (`osm_id` in general). It is not necessary to have `osm_id` as FeatureID in vector tiles and `osm_id` as a regular attribute. The value as `feature_id` could be gain with a different approach (as nyurik writes in https://github.com/openmaptiles/openmaptiles/issues/804#issuecomment-610030467)

It is working for `mountain_peak` but for `aerodrome_label` I was not able to find feature_id in vector tiles - I am using (streets for aerodrome and topo for peaks) https://maplibre.org/maplibre-gl-js-docs/example/queryrenderedfeatures/

ZeLonewolf:
@TomPohys I made some changes in a local branch, and below is what I'm currently generating.  Can you tell me if this is what you expect to see?

I assume that when you say `featureID` you mean the top-level `id` tag?

image
image

TomPohys:
Yes, the `FeatureID` is top-level `id`. If there already is `FeatureID` it is not necessary to have `propertiesL{id}` as it is the duplicity of the same `osm_id`. Prefered is `FeatureID` over `propertiesL{id}`.

There are a lot of `id` :smile: 
 - top-level `id` (FeatureID), 
 - `id` in `properties` (I would prefer if this could be `osm_id` insted of just `id`), 
 - `id` in `layer`. 
 
From current inspect by `queryrenderedfeatures` the `mountain_peak` has `FeatureID`. The `aerodrome_label` missing `FeatureID`. It would be superb if you could apply changes that you did in your local branch for `aerodrome_label`, but (I would prefer, but I am keen to discuss ) without adding `properties{id}`. `mountain_peak` from current inspect looks good to me (there is `FeatureID` and not `properties{id}`)

ZeLonewolf:
Thanks @TomPohys for being patient with me and explaining this.  I think we have a workable solution now, based on these screen shots:

image
image

TomPohys:
Awesome! Thanks for the additional changes!

TomPohys (approved):

nyurik:
I just looked at this change, and got really confused. The [change comment](https://github.com/openmaptiles/openmaptiles/commit/36bd917e050593c04ebee7c8455c1dcb8a346b7b) says "Implement feature IDs in aerodrome_label and mountain_peak", whereas the actual change simply removes quotes (a noop in YAML), and fixes grammar of the field description.  Was this intentional??

ZeLonewolf:
Ugh, we came full circle to a no-op.  Can you clarify what it means to have feature IDs in the tile for these objects?  It looks like `id` is actually set here.

ZeLonewolf:
Okay, so it turns out that no change was actually necessary, and these two layers _already_ had a working featureID.  Since both of the layers come entirely from OSM data, and not natural earth or some external source without IDs, using osm_id as the featureID is an acceptable approach here.

TomPohys:
I just found out, that not all `aerodrome` have `osm_id`. One example - https://www.openstreetmap.org/relation/12007076

`./quickstart.sh rhode-island`
`make psql`
result:
```bash
openmaptiles=# select osm_id, name from osm_aerodrome_label_point;
   osm_id   |             name              
------------+-------------------------------
  -12007076 | Quonset State Airport
  532546632 | 
  141518051 | Westerly State Airport
  141564591 | North Central State Airport
  141565041 | Newport State Airport
  368382660 | Albert Farms Airport
  368382677 | Boystown Fly-in-Ranch
  368382679 | Wing-Over Farm Airport
  368382681 | Keskinen Airport
  368382682 | Riconn Airport
  368382684 | Tiverton Airport
  368382744 | Richmond Airport
 1042055823 | Plouffe Landing Seaplane Base
 2139225910 | Coventry Airpark (Historical)
  209870429 | T. F. Green Airport
  495906627 | Block Island State Airport
(16 rows)
```
Quonset State Airport is a relation, but probably set ABS(osm_id) do the trick for `aerodrome`

ZeLonewolf:
Sure, we can do an absolute value - the only issue is that you won't have a perfect guarantee that the IDs will be unique (although it is extremely unlikely).  From my conversation wtih @nyurik, I understand that unqiue IDs aren't actually a requirement.

PR #1177, opened 2021-08-16
Add field for toll roads

### LINKED ISSUE #366: M6 Toll UK Motorway
In the UK the M6 motorway is split in two - 'M6' and 'M6 Toll'. M6 Toll is effectively a separate motorway.

At present the ref tag in OpenMapTiles for both is M6. Can this be changed to M6 Toll where appropriate.

OSM tags have the ref as M6 Toll but that is not coming through to the OpenMapTiles vector layers Ref.

ZeLonewolf:
Fixes #366

This PR sets `toll=1` in the `transportation` layer when a road is tagged as a toll road in OSM (with the tag `toll=yes`).  If a road is tagged with any other value of `tag=*`, the value is suppressed in the tile, since non-toll roads can be presumed as the default and therefore this PR should have only negligible impact in the tiles.

Support for toll road tagging is of interest in the American mapping community, because toll roads have historically been styled differently on American-style maps, for example:
image

Screen shot of a toll road being generated in the `transportation` layer tiles:
image


TomPohys:
Thanks, It looks good. Also `toll` tag probably does not create line segments for long highways.

TomPohys (approved):

PR #1182, opened 2021-08-18
Water layer river mapping bug fixes

ZeLonewolf:
This PR is a bugfix for the `water` layer.

* `waterway=stream`, `waterway=river`, `waterway=canal`, `waterway=ditch`, and `waterway=drain` are all linear features, not area features.  Thus, these objects are being unnecessarily mapped into the `osm_water_polygon` polygon table, and this PR removes these unneeded mappings.
* The combination `natural=water` + `water=river` is the most popular tagging for river areas.  However, the current mapping causes rivers tagged in this way to be rendered in the vector tiles as a `lake`.  This PR adds a check for the `water=river` tag and tags both variants of river areas as `class=river`.

`natural=water` + `water=river` river mapping:
image

`waterway=riverbank` river mapping:
image

Lake mapping for a `natural=water` (with no other tagging):
image


TomPohys:
Thanks @ZeLonewolf for this PR, it looks good, but I need to do some local tests.

eva-j (approved):
Thanks a lot for this fix, it works good.

TomPohys (approved):

PR #1185, opened 2021-08-20
Fix negative feature IDs

ZeLonewolf:
This PR removes the possibility of negative feature IDs for the `aerodrome_label` layer by defining the feature as the absolute value of the OSM id (imposm maps relations with negative numbers).  There is a very unlikely  scenario in which a relation and a way have the same ID (and are also in the same tile), however, unique IDs are not a strict requirement of MVT.

Noted in:
https://github.com/openmaptiles/openmaptiles/pull/1176#issuecomment-902503655

Positive feature ID for Quonset State Airport, which is mapped as a relation:
image


TomPohys (approved):

TomPohys:
Thank you!

Issue #271, opened 2017-05-31
Show track/path sooner


Duke74:
When using openmaptiles for hiking focused mapping, paths and tracks are displayed too late.

Making it difficult for the final user to find paths that are going around a mountain for example.

Paths and Tracks are currently displayed at zoom level : 14
Here is what it looks like (path and tracks are red):
![chrome_2017-05-31_15-59-25](https://cloud.githubusercontent.com/assets/1609064/26636262/7d105464-461c-11e7-8125-e62498cfae71.png)

In my opinion for the hiking map usecase it would be interesting to make it appear sooner (z11-12 ?):
It would allow us to make maps that looks like this:

![chrome_2017-05-31_15-59-06](https://cloud.githubusercontent.com/assets/1609064/26636368/db332418-461c-11e7-819e-cd9827cee9ee.png)
This map allow a bigger field of view making it easier to plan an hiking tour without having to scroll everywhere at zoom 14.
do you guys have the same opinion ?

To go arround the issue I'm generating my own vector tiles with the paths and tracks only and use it as an additional source in my style, it's working but it would be interesting to add this change to openmaptiles.

I noticed the issue for track and path but it may also apply to alpine_huts, viewpoints ...

carlos-mg89:
Totally agree with this request. Would there be any way of changing any configuration so this is something that could be fixed? The paths and tracks are simply not in the MBTiles of zoom level less than 14. So it's not an issue of my style (I've checked with other MBTiles), and I'm modifying it for other elements.

nnhubbard:
You are missing one thing. In `transportation/mapping.yaml` you would need to add `track` and `path` to the `sql_filter`, e.g.:

`sql_filter: highway IN ('motorway', 'trunk', 'primary', 'secondary', 'tertiary', 'motorway_link', 'trunk_link', 'primary_link', 'secondary_link', 'tertiary_link', 'track', 'path') AND NOT is_area AND ST_IsValid(geometry)`

carlos-mg89:
I've just applied the changes to those 2 files, and just fired `./quickstart.sh area_to_be_downloaded`. Would this be enough? Or anything else has to be done for the layers to be taken into account?

nnhubbard:
Yes it will be enough to include the data in the import. I don’t think the style need to be adjusted either,  but I can’t remember, not looking at my style sheet at the moment.

carlos-mg89:
It worked like a charm! Thanks @nnhubbard 

[...]

PR #1186, opened 2021-08-20
show tracks / paths at zoom 12 and 13

farfromrefug:
As discussed this PR changed the visibility of tracks/paths to be shown at lower level as done by most hikings maps

### DIFF
--- a/layers/transportation/transportation.sql
+++ b/layers/transportation/transportation.sql
@@ -360,12 +360,11 @@ FROM (
          WHERE NOT is_area
            AND (
                      zoom_level = 12 AND (
-                             highway_class(highway, public_transport, construction) NOT IN ('track', 'path', 'minor')
+                             highway_class(highway, public_transport, construction) NOT IN ('minor')
                          OR highway IN ('unclassified', 'residential')
                      ) AND man_made <> 'pier'
                  OR zoom_level = 13
                          AND (
-                                    highway_class(highway, public_transport, construction) NOT IN ('track', 'path') AND
                                     man_made <> 'pier'
                             OR
                                     man_made = 'pier' AND NOT ST_IsClosed(geometry)


nyurik:
@farfromrefug please update the first comment, describing what you are trying to do

ZeLonewolf:
I believe the intent is for this to fix #271

farfromrefug:
Btw i have to say i am not 100% sure this would go in the main repo. Feels weird to me to have tracks/paths showing before minor for non hiking maps. It.might a first step befote having configurable min zoom level

TomPohys:
Hi, thanks for PR. based on github-actions statistics, there is a raise in tile size by 28% and 19% for zoom 11 and 12. Could you please provide statistics how much will raise the whole mbtiles for some bigger areas (France, Switzerland or Europe)? Thanks

ZeLonewolf:
I was surprised to see how big the tile increase was - more than I expected.  I think the goal here is really to show hiking trails at lower zooms, but not specifically all paths.

If you look at the tag conflation of `highway=path`, [it's quite telling](https://taginfo.openstreetmap.org/tags/highway=path#combinations): 
* Only 5.2% of paths have a `name=*` tag
* Only 4.7% of paths have a `sac_scale=*` tag

Instead, I think it would make more sense at these zooms to only show the paths that have some type of tagging that indicates that they're a hiking trail.  This would eliminate 90%+ of paths added in the PR but would still leave behind paths that are important to show at lower zooms.

farfromrefug:
Interesting point. Now from an hiking app point of view you need them all. Let me explain. 
In my app style, i, by mistake, did exactly what you suggested. I had all data from zoom 12 but filtered on sacscale until zoom 13.
As you can see in the screenshots (zoom at bottom right) the most important paths (those leading to summits) dont have sac scale and thus are filtered until zoom 13. Thus showing only sacscale from zoom 12 is kind of useless. 
The more i think about it the less i think this should be in master
Screenshot_20210824-083945_Alpi_Maps
Screenshot_20210824-083941_Alpi_Maps

ZeLonewolf:
@farfromrefug I think "only `sac_scale`" is not enough, as you note.  However, any trail with a `name`, `sac_scale`, is part of a route relation, has a `colour`, `osmc:symbol`... and possibly other checks we can make, is probably a hiking trail, and that filter would STILL eliminate probably 90%+ of the path objects while still leaving behind the paths that make sense to show.

farfromrefug:
@ZeLonewolf at least not in France. Most deep montain trails are only path/path. From what i am seeing ( and it is sometimes tough to style) it is far from 90% of true hiking trails. A funny test would be to see what the closest paths from peaks are tagged. surely for like the Mont Blanc they have tags. But from what i see in the french Alps it is not the case

ZeLonewolf:
Surely there is some middle ground that can be achieved between "all paths at z12-13" and "no paths at z12-13" so that general purpose styles can still show some trails at these zooms.

farfromrefug:
@ZeLonewolf my point is if we  want to add path at zoom 12 because of hiking apps it seems to me it should either b all or none. as an hiking app devs the middle ground is kind of useless. And it does rise tile s size for everyone not only hiking apps devs. 
Just my opinion, personally dont really care as i use my own fork...

ZeLonewolf:
I think we want to add (important) paths up to zoom 12 because it's of general interest to general-purpose maps, not just hiking maps.  This exposes yet again the need to have switchable customizations so that every minor adjustment doesn't require a fork of the SQL layer.

TomPohys:
Thank you for the discussion. Let's close this in advance of #1190, where @ZeLonewolf using path filtering for z12 and z13.

Issue #469, opened 2018-06-27
Missing shops


nlehuby:
For now, we have a white-list of values for the `shop` tag to import in the poi layer.

Even if we have really many of them, and the most used according to `taginfo`, there is a few missing ones.

Why don't we import all the OSM objects the the `shop` key ? Would you accept a PR for that ?

klokan:
If it does not significantly increase size of tiles - we would be glad to marge any changes on POI from @openmaptiles/qwant-research team.

Please make PR.

ZeLonewolf:
Based on the discussion at https://github.com/openmaptiles/openmaptiles/pull/1188#issuecomment-903226853, it appears that the OpenMapTiles consensus is to continue considering new `shop=*` values on a case-by-case basis, in order to keep a handle on tile sizes and prevent vandalism.  Therefore, this issue should be closed as **wontfix** rather than open and labelled "help wanted".

PR #1188, opened 2021-08-21
Add locksmith shop

### LINKED ISSUE #566: No locksmith shop on tiles
I report this issue : https://github.com/QwantResearch/qwantmaps/issues/23

It seems that there is no shoemaker or locksmith on openmaptiles : https://github.com/openmaptiles/openmaptiles/blob/master/layers/poi/mapping.yaml#L117

Could openmaptiles include the shop=locksmith tags ?

Thanks !

ZeLonewolf:
Fixes #566

This PR adds support for `shop=locksmith`.

Example OSM object: https://www.openstreetmap.org/way/874850561

Rendered tile:
image


ZeLonewolf:
Note that #566 conflicts with #469.  We should really decide whether we want a white-list of shop values, or allow arbitrary values.

farfromrefug:
@ZeLonewolf how do you decide which shops to show?
I mean i recently changed in my fork to show all shops except vacant and unknown. I feel it was too hard to decide which ones to show or not (for example sea_food is missing too)

TomPohys:
Hi, thanks for the PR. Locksmiths are OK to add. I would like to not adding all shops with arbitrary values. There are a lot of names that are not useful at all. There is also big potential of vandalism with allowing `shop=*`

TomPohys (approved):

farfromrefug:
@TomPohys good point! I was mostly asking how you choose which ones to show

TomPohys:
@farfromrefug, if the shop's characteristics make sense (e.g. does not have like 10 records, or it is grey-zone ) to show, let's add it.

PR #1190, opened 2021-08-24
Selective rendering of tracks and paths at z12-13

### LINKED ISSUE #271: Show track/path sooner

ZeLonewolf:
Closes #271 

This PR adds track and path rendering at lower zooms than currently provided, and also achieves near-parity with openstreetmap-carto on track and path rendering.  A previously-abandoned attempt, with significant discussion, was #1169.

### Discussion

The following table shows the zoom levels that paths and tracks currently appear in both openstreetmap-carto and OpenMapTiles:

| Zoom Level | openstreetmap-carto | OpenMapTiles |
| -- | -- | -- |
| vz12 / rz13 | `highway=track` |
| vz13 / rz14 | `highway=path` |
| vz14 / rz15 | | <ul><li>`highway=track`</li><li>`highway=path`</li></ul>

In order to achieve parity between openstreetmap-carto and OpenMapTiles, it would be necessary to render paths in z13 and tracks in z12-13.  The naive approach approach of simply adding these features to these zooms was attempted in #1186 which proposed to render all tracks and paths up to z12, which would match openstreetmap-carto at z13 and exceed it at z12 (because openstreetmap-carto renders tracks but not paths at this zoom).  This resulted in a 20% increase in the tile size at z13 in the test area, which was an unacceptable increase in tile size.  The tile size increase was even larger at z12, however, this is partly because paths are also extended to z12, which exceeds that of openstreetmap-carto.

While the large increase in tile size is unworkable, it is also not desirable that OpenMapTiles has a feature gap with openstreetmap-carto for tracks and paths.  Therefore, this PR adds selective, progressive rendering of important paths at z12-13 and universal rendering of tracks at z12.  This brings OpenMapTiles to parity with openstreetmap-carto for tracks while bringing a similar, but more nuanced rendering of paths at z12-14.

### Selective Path Rendering
With selective path rendering, we select "more important" paths to render at zoom 13, and only the "most important" paths for zoom 12.  In order to identify whether a path is more/most important, we can look at the following tagging:
* `name` (unnamed paths are less important)
* `sac_scale`
* If the trail is a member of a hiking route relation, possibly with:
    * `network`
    * `osmc:symbol`
    * `colour` 

Of note in the discussion in #1186 was that there are likely important hiking trails that are tagged as unnamed `highway=path` in the map without any of the tagging described above.  Since there is no way to distinguish unnamed/untagged high-importance hiking trails from unnamed/untagged low-importance hiking trails, all we can do is establish minimal rules for importance based on tagging.  This provides mapper feedback to add this tagging, where applicable, to important hiking trails, so that they can appear in general-purpose styles at lower zooms.

In particular, the selective rendering of trails of national importance at lower zooms is of interest to the openstreetmap-americana project, and American mapping in general.  Long-distance trails such as the [Appalachian Trail](https://en.wikipedia.org/wiki/Appalachian_Trail), [Pacific Crest Trail](https://en.wikipedia.org/wiki/Pacific_Crest_Trail), or [Continental Divide Trail](https://en.wikipedia.org/wiki/Continental_Divide_Trail) have significant cultural importance and run through rural areas with low map feature density.  Therefore, it is cartographically appropriate to render these features at lower than usual zoom levels for walking paths.

Below is an example from Pennsylvania, USA, showing high-importance hiking trails being rendered at city-level zooms on a general-purpose (non-hiking-specific) map:

### Proposed Zoom Filter
This PR implements the following rules:

**Zoom 12**
* All `highway=track`
* Paths that are members of international & national hiking routes (`network=iwn`/`network=nwn`)

**Zoom 13**
* All `highway=track`
* Paths that are members of a hiking route, or have a name, `sac_scale`, `osmc:symbol`, or `colour`

**Zoom 14**
* All `highway=track`
* All `highway=path`

### Rendering Samples

Example location: https://www.openstreetmap.org/#map=12/41.5867/-71.7063

The following samples were generated from zoom 11-14 tiles from the us-northeast extract.

Zoom 11
Zoom 12
Zoom 13
Zoom 14


daliborjanak:
I’m happy with adding and moving roads to upper zooms. Also, the proposed zoom filter based on sac_scale, color, etc. is a really good idea. 

But please don’t add more attributes related to hiking - sac_scale and values network-* to this schema. The design of this schema is to be general-topographic. It would be better to add hiking stuff as a different layer with the behavior of https://github.com/openmaptiles/openmaptiles/pull/1003 turned off by default.

We did it as a separate layer on global field https://www.maptiler.com/maps/outdoor/ look at schema here https://docs.maptiler.com/schema/outdoor/

TomPohys:
Hi @ZeLonewolf, thanks for the PR. OpenMapTiles should provide the schema for the general-purpose map, or base map. Unfortunately, adding another hierarchy layer into the `transportation-highway-path` layer is out of the scope of a base map. Now, the transportation lines with key `highway` have hierarchy like `motorway`, `trunk`, `primary`, `secondary`, `tertiary`, `minor`, `track` and `path`. With your PR you add another hierarchy to `path` (`international`, `national`, `regional`, `local`) and another attribute `sac_scale`. 

Unfortunately, the `sac_scale` is not the only scale to differentiate the hiking path (https://wiki.openstreetmap.org/wiki/Proposed_features/cai_scale). `sac_scale` is also not used all over the world. On the other hand in Europe, there could be too much path visible in z12, which also leads to much bigger tiles. This has to be also tested.

Could you please keep the selecting logic for `path` (based on `network`, `sac_scale`, `osmc_symbol` and `colour`) but not add the `network` values and `sac_scale` attribute to the `path` (have the values in the database but not in tiles)?

I am preferring your approach than just move all the paths to zoom 12 (#1186). 

Thank you

ZeLonewolf:
Thanks for the detailed feedback @TomPohys.  I will work on making changes consistent with your description.  Before I do that, I am planning to complete #1207 to refactor the zoom 12-14 logic which I find confusing to read.  Separating this refactor will help ensure that this PR does not introduce unexpected bugs do to converting the nested `IF`/`AND`/`OR` logic to more readable `CASE` statements.

TomPohys:
Great, looking forward to it!

ZeLonewolf:
@TomPohys @daliborjanak

I've modified this PR to incorporate the cleanup in #1207 and remove `sac_scale` from the tiles.

I'm not quite sure of the best approach to incorporating hiking network information into the zoom determination without exposing new network_type values in the tiles.  I'm considering the following approaches:
1. Compute a `rank` value at a low level and ripple it up through the `transportation` and `transportation_name` layers
2. Suppress the hiking network types at the last moment before they are sent to the tiles
3. Make raw network values available to the `transportation_name` tables in addition to the current network_type values so that they can be used in the z12-14 zoom logic

I welcome any thoughts you have on best approaches here.

nnhubbard:
I have used `osmc:symbol` in a custom layer for my fork. When building out this feature, I realized it was very hard for styles to consume that tags data because of the format. Parts of it are the way color, the symbol, text, etc. 

I propose that the tag be split, which allows for much easier styling. For example, on iOS, it would be almost impossible to use the full `osmc:symbol` to create styles for a path and then extract the symbols. Instead, I did the following:

```
NULLIF(SPLIT_PART(osmcsymbol,':', 1), '') AS osmcsymbol_way_color,
NULLIF(SPLIT_PART(osmcsymbol,':', 2), '') AS osmcsymbol_background,
NULLIF(SPLIT_PART(osmcsymbol,':', 3), '') AS osmcsymbol_foreground,
NULLIF(SPLIT_PART(osmcsymbol,':', 4), '') AS osmcsymbol_foreground2,
NULLIF(SPLIT_PART(osmcsymbol,':', 5), '') AS osmcsymbol_text,
NULLIF(SPLIT_PART(osmcsymbol,':', 6), '') AS osmcsymbol_text_color,
```

This provides separate fields for `osmc:symbol` which makes it MUCH easier for styling. Thoughts?

ZeLonewolf:
> I propose that the tag be split, which allows for much easier styling. For example, on iOS, it would be almost impossible to use the full `osmc:symbol` to create styles for a path and then extract the symbols.

@nnhubbard it seems reasonable to me, however it's certainly outside the scope of this issue, and there is resistance to adding hiking-related data into the main style.

ZeLonewolf:
I have updated this PR consistent with the comments from @daliborjanak and @TomPohys.  Accordin

ZeLonewolf:
I have updated this PR consistent with the comments from @daliborjanak and @TomPohys. Accordingly, there are no new hiking network values and the symbol/colour tagging is not exposed to the tiles.  The tile size increase is very small while still adding "significant" paths and track roads to the map at lower zooms.

layers/transportation/transportation.sql
```diff
@@ -359,9 +359,17 @@ FROM (
          FROM osm_highway_linestring
          WHERE NOT is_area
            AND
-               CASE WHEN zoom_level = 12 THEN transportation_filter_z12(highway, construction)
+               CASE WHEN zoom_level = 12 THEN
+                         CASE WHEN transportation_filter_z12(highway, construction) THEN TRUE
+                              WHEN network IN ('hike-international', 'hike-national', 'hike-regional') THEN TRUE
```
TomPohys, line 364: Shouldn't this condition be same as in `transportation_name` layer?
https://github.com/openmaptiles/openmaptiles/blob/c4d2c80801688ba91095af6b941ead021e171148/layers/transportation_name/transportation_name.sql#L117
 
```suggestion
                              WHEN route_rank = 1 THEN TRUE
```

layers/transportation/update_route_member.sql
```diff
@@ -84,14 +89,20 @@ CREATE INDEX IF NOT EXISTS osm_route_member_network_type_idx ON osm_route_member
...
     osm_id,
-    DENSE_RANK() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index
+    DENSE_RANK() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index,
+    CASE
+         WHEN network IN ('iwn', 'nwn', 'rwn') THEN 1
+         WHEN network = 'lwn' THEN 2
+         WHEN COALESCE(osmc_symbol, colour) <> '' THEN 2
+    END AS rank
   FROM osm_route_member
   ON CONFLICT (id, osm_id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index;
```
TomPohys, line 107: I was unable to add `rank` values for hiking routes. This fixes it.

```suggestion
  ON CONFLICT (id, osm_id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index, rank = EXCLUDED.rank;
```

layers/transportation/transportation.sql
TomPohys, line 364: Is this just an old code, which should be replaced? 
```suggestion
                              WHEN network IN ('iwn', 'nwn', 'rwn') THEN TRUE
```

ZeLonewolf: Yes, my version is clearly wrong here.  I need to investigate this because that network value should be coming from the route relation so I'm not sure that network is populated in that way.  Going to send this back to draft took more closely at this.

TomPohys: Thank you.

As for `osm_transportation_name_linestring` table where is `route_rank` added in https://github.com/openmaptiles/openmaptiles/blob/949896a205636b25b9ca0305bcabe7ae028e2461/layers/transportation_name/update_transportation_name.sql#L57

then you are using `route_rank` directly in `layer_transportation_name()` function
https://github.com/openmaptiles/openmaptiles/blob/949896a205636b25b9ca0305bcabe7ae028e2461/layers/transportation_name/transportation_name.sql#L117


But there is no `route_rank` in `osm_highway_linestring`. And as you use this approach for populating `osm_route_member`, https://github.com/openmaptiles/openmaptiles/blob/949896a205636b25b9ca0305bcabe7ae028e2461/layers/transportation/update_route_member.sql#L102

I thought, that it could be used here as I suggested.

ZeLonewolf: The solution needed to be a bit more complex.  The most efficient approach was to use the `osm_transportation_name_network` table directly.  To achieve this, I moved that materialized view down to the `transportation` layer, added unique index for performance, and simply joined it to the z12-14 table merge in order to access the pre-calculated route rank.

ZeLonewolf:
This PR is updated to move the `osm_transportation_name_network` table down to the `transportation` layer (with follow-on effects) in order to take advantage of the pre-computed `route_rank` value.  I also fixed a bug in the z13 rank calculation and added a few unit tests to verify that route rank calculations are working properly.

TomPohys:
Hi, thank you very much! It looks awesome. 

Just two questions
 - in commit fd4c510 you add `layers/transportation_name/update_transportation_name.sql.save` what is its function?
 - should not be added `, rank = EXCLUDED.rank` on the line https://github.com/openmaptiles/openmaptiles/blob/83ee059cc2bb78583e53474bef24d10cadb6bfdc/layers/transportation/update_route_member.sql#L77 same as on the line https://github.com/openmaptiles/openmaptiles/blob/83ee059cc2bb78583e53474bef24d10cadb6bfdc/layers/transportation/update_route_member.sql#L107

I would like to merge this PR before https://github.com/openmaptiles/openmaptiles/pull/1295. It seems like that will be less effort to resolve conflicts.

ZeLonewolf:
Thanks @TomPohys for your comprehensive review as always!  I've posted a new updated to fix the issues noted.

TomPohys:
Thank you very much! It looks awesome!

TomPohys (approved):

PR #1200, opened 2021-08-28
Remove area checks in Makefile for `import-borders` & `generate-tiles-pg`

Falke-Design:
While importing multiple .pbf files into postgres I was confronted with some unnecessary problems:

1. I wanted to use the borders of the planet which I had already generated. But `make import-borders` needs a .pbf file which is not necessary if the line.csv file is already existing -> So I added a check if the line.csv is **not** existing it checks for the `area parameter` / .pbf file
2. `make generate-tiles-pg` is only possible with a specified `area parameter` BUT the area is not used / needed for generation -> So I removed the area check


Makefile
```diff
@@ -402,10 +402,14 @@ import-data: start-db
...
-	$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools sh -c \
-		'pgwait && import-borders $$([ -f "$(BORDERS_CSV_FILE)" ] && echo load $(BORDERS_CSV_FILE) || echo import $(PBF_FILE))'
+ifeq (,$(wildcard $(BORDERS_CSV_FILE)))
+        @$(assert_area_is_given)
+        @echo "Generating borders out of $(PBF_FILE)"
+else
+        @echo "Borders already exists. Useing $(BORDERS_CSV_FILE) to import borders"
+endif
+        $(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools sh -c \
+                'pgwait && import-borders $$([ -f "$(BORDERS_CSV_FILE)" ] && echo load $(BORDERS_CSV_FILE) || echo import $(PBF_FILE))'
```
TomPohys, line 412: Simple switch from spaces to tabs

```suggestion
ifeq (,$(wildcard $(BORDERS_CSV_FILE)))
	@$(assert_area_is_given)
	@echo "Generating borders out of $(PBF_FILE)"
else
	@echo "Borders already exists. Useing $(BORDERS_CSV_FILE) to import borders"
endif
	$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools sh -c \
		'pgwait && import-borders $$([ -f "$(BORDERS_CSV_FILE)" ] && echo load $(BORDERS_CSV_FILE) || echo import $(PBF_FILE))'
```

TomPohys:
Hi @Falke-Design, thank you for your PR. I just suggested switching from spaces to tabs.

TomPohys:
Thanks! It makes sense.

TomPohys (approved):

Issue #1151, opened 2021-07-14
`bc: command not found` Error quickstart.sh


Falke-Design:
**Problem:**
When I clone the latest version from git and run `./quickstart.sh` it throws following error:

```
./quickstart.sh: line 134: bc: command not found
xargs: echo: terminated by signal 13
```

**Solution:**
I simply removed in line 143 `bc`:
```
mem=$( grep MemTotal /proc/meminfo | awk '{print $2}' | xargs -I {} echo "scale=4; {}/1024^2" | bc )
```
to:
```
mem=$( grep MemTotal /proc/meminfo | awk '{print $2}' | xargs -I {} echo "scale=4; {}/1024^2")
```
I think this brokes the calucation / command **but** this value is only displayed in the console and has no other usages.

**Suggestion**
Can be checked if `bc` is available else the memory output will skipped?

**Server:**
OS: Ubuntu 20.04.1 LTS   minimal
Kernel: Linux 5.4.0-67-generic

TomPohys:
Thanks! PR is very welcome!

PR #1201, opened 2021-08-28
Check if bc command is available

Falke-Design:
On some systems the `bc` command is not installed, so I added a check if the command can be used.

It is a simply calculation from KB to GB.  If `bc` is missing, then it is displayed in KB.

fix: #1151


TomPohys (approved):

TomPohys:
Thank you, it looks good!

quickstart.sh
```diff
@@ -130,9 +130,15 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
       echo "ERR: Sorry this is working only on x86_64!"
       exit 1
     fi
+
     echo "      : --- Memory, CPU info ---- "
-    mem=$( grep MemTotal /proc/meminfo | awk '{print $2}' | xargs -I {} echo "scale=4; {}/1024^2" | bc )
-    echo "System memory (GB): ${mem}"
+    if [ -n "$(command -v bc)" ]; then
```
nyurik, line 135: thx, i have always used `which`, but i guess `command` is a more recommended way? (saw a comment about it, but don't have an authoritative source).  Either works.  Why do you use `-v` though?

Falke-Design: I don't know which of the commands would be better. I don't know `which` and i get `command -v` from an stackoverflow answer.

But i can verfiy that `-v` is good because it displays only the path of the program or blank. Without `-v` it displays an explanation of the program, which needs more time to display and it can happen that a program has no description, so it would be return blank too.

Issue #274, opened 2017-05-31
Missing natural data


Duke74:
A lot of data from osm key:natural is missing: http://wiki.openstreetmap.org/wiki/Key:natural

For example I'm interested in : 
- valley
- cliff
- bare_rock
- scree
- ridge
- arete

Is it a choice not to include this in openmaptiles, to optimize the tiles size ?
In this case should I use my own layers in order to add this data ?

stirringhalo:
That's a tricky question, we'll have to see how large our current tile sizes are and see if its feasible.

stirringhalo:
- [ ] test

stirringhalo:
Will do ridge, arete, cliff and valley in a separate layer and PR.

stirringhalo:
@klokan , @jirik  Btw, these features need to be added to the styles.

jirik:
BTW ridge, arete, cliff and valley are mapped as linear features (not polygon) and only cliff is currently rendered by osm-carto.

kylebarron:
`natural=spring` is also not currently included. I added it in my own fork to the POI layer. Is this something you're interested in a PR for?

kylebarron:
What about adding other point `natural` values to the `mountain_peaks` layer? I'm specifically thinking of `natural=spring` and `natural=saddle`.

1ec5:
[`natural=mud`](https://wiki.openstreetmap.org/wiki/Tag:natural=mud) and [`natural=dry_lake`](https://wiki.openstreetmap.org/wiki/Tag:natural%3Ddry_lake) would also be nice to have. [Large expanses of the Western United States](https://www.openstreetmap.org/relation/11485256) are covered in one of these tags. It can be jarring when they’re missing from a map that otherwise shows wetlands and other natural features.

PR #1202, opened 2021-08-30
Add enhanced mountain features

ZeLonewolf:
This PR adds support for several features in the `natural` key, including `saddle`, `ridge`, `arete`, and `cliff`.  These features all currently render in openstreetmap-carto, so adding these features would advance OpenMapTiles towards its goal of parity with the standard renderer, as indicated in https://github.com/openstreetmap/operations/issues/565#issuecomment-907303115.

This PR also adds the features requested in #274, with the exception of valleys, which I've left out of this PR because there are different complexities to mapping valleys that should be addressed as a separate (future) PR.

### Examples of mountain features in paper maps
These features are regularly found in American-style maps and thus is of interest to the openstreetmap-americana map style.  Below are examples of these mountain-related features found in general purpose maps:

### Approach
This PR extends the existing `mountain_peak` layer by adding other mountain features.  We may want to consider renaming this layer in version 4.0 to be more inclusive of other mountain features including the potential future addition of valleys.  However, the features added in this PR are associated with mountains and so their inclusion in this layer is the most appropriate location.

A new `osm_mountain_linestring` mapping maps the new linear mountain features, with similar ranking logic as is used for the existing `peak`/`volcano` features.  Additionally, `natural=saddle` is added to the `osm_peak_point` mapping and ranked using the formula for peaks.  Since saddles will have lower elevations than peaks, important saddles will be preempted by important peaks.

The new linestring features are rendered only at zoom 13-14, in order to match the zoom at which they appear in openstreetmap-carto.  However, it may be appropriate in a future PR to extend the rendering of these features as generalized linestrings at lower zooms.

### Test Renderings
Below is a test rendering showing aretes and peaks in Austria, just north of the Swiss border, followed by a screen shot of the [same location](https://www.openstreetmap.org/#map=14/46.8682/10.0863) in openstreetmap-carto:

**OpenMapTiles, zoom 14**

**openstreetmap-carto, zoom r14/v13:**


TomPohys:
Thank you for this PR and detailed explanation text. These natural features will enhance the `mountain` layer. 

You are right that if this `mountain_peak` layer is not just `mountain peaks` it should be renamed in v4 to e.g. `mountain`.

TomPohys (approved):

layers/mountain_peak/mountain_peak.sql
```diff
@@ -55,6 +55,39 @@ FROM (
      ) AS ranked_peaks
 WHERE zoom_level >= 7
   AND (rank <= 5 OR zoom_level >= 14)
+
+UNION ALL
+
+SELECT
+    -- etldoc: osm_mountain_linestring -> layer_mountain_peak:z13_
```
zstadler, line 62: `make generate-devdoc` says:

```
Creating layers/mountain_peak/etl_diagram.png
Warning: node layer_mountain_peak, port z13_ unrecognized
```

ZeLonewolf: Fixed in #1223

zstadler:
@ZeLonewolf,

You wrote [with my emphasis]:
> This PR also adds the features requested in #274, **with the exception of valleys, which I've left out of this PR because there are different complexities to mapping valleys** that should be addressed as a separate (future) PR.

Could you elaborate on the differences between `natural=valleys` and `natural=ridges`, which are included in this PR?
Are these differences related to the styling or the ability to include valleys in the OMT tiles?

Here are two example of how valleys show on USGS maps:

image
[Leuthold Couloir](https://www.openstreetmap.org/way/724854492)

image
[Alkali Canyon](https://www.openstreetmap.org/way/924000559)

ZeLonewolf:
Basically, the tagging of valleys is controversial so I chose to skip them in order to not hold up the other types. But I think it would be fine to add them.

PR #1203, opened 2021-08-30
Add the complete workflow "how to generate tiles" to readme

Falke-Design:
When I started with OpenMapTiles I was missing an explanation how I can generate tiles with the make commands.

Demo: https://github.com/Falke-Design/openmaptiles/tree/feature/readme_workflow#workflow-to-generate-tiles


nyurik (approved):
awesome, thanks!

Issue #1197, opened 2021-08-27
Can't stop script `make generate-tiles-pg`


Falke-Design:
It is not possible to stop the execution of `make generate-tiles-pg`  with `CTRL+C` . I need to wait until it is finished. or open a new terminal and kill the process.

![grafik](https://user-images.githubusercontent.com/19800037/131159224-eed429ee-d2d3-46e1-ab6f-60deda6549b6.png)

PR #1205, opened 2021-09-01
Fix Ctrl+C when generating tiles

nyurik:
Adding -T fixes the issue when users are unable to stop tile genartion with ctrl+c

For some reason it doesn't work without it, even though the process of building
and running both docker images is the same.


Falke-Design:
Issue: #1197

TomPohys:
Thanks!

TomPohys (approved):

PR #1207, opened 2021-09-02
Refactor transportation layer's z12-14 logic

ZeLonewolf:
This PR is a refactor.  The purpose of this PR is to refactor the logic in the z12-z14 transportation layer to increase readability and reduce code complexity.

In addition, this fixes bugs in the original filter code:
* Create filter parity between `service` and `service_construction` highway classes.
* Removal of unintended `class` values `minor_construction`, `path_construction`, `service_construction`, and `track_construction` from zoom 12 filter by explicitly listing the highway classes that are included.


layers/transportation/transportation.sql
```diff
@@ -358,26 +358,20 @@ FROM (
                 z_order
          FROM osm_highway_linestring
          WHERE NOT is_area
-           AND (
-                     zoom_level = 12 AND (
-                             highway_class(highway, public_transport, construction) NOT IN ('track', 'path', 'minor', 'service')
-                         OR highway IN ('unclassified', 'residential')
-                     ) AND man_made <> 'pier'
```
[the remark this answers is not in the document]
ZeLonewolf: Thanks, I need to look at this more closely.  The layer sizes changed so clearly I didn't achieve a 1:1 refactor here.

layers/transportation/transportation.sql
```diff
@@ -358,26 +358,23 @@ FROM (
...
+           AND
+               CASE WHEN zoom_level = 12 THEN
+                    CASE WHEN highway_class(highway, public_transport, construction) NOT IN ('track', 'path', 'minor', 'service') THEN TRUE
+                         WHEN highway IN ('unclassified', 'residential') THEN TRUE
+                    END
+
+                    WHEN zoom_level = 13 THEN
+                    CASE WHEN highway_class(highway, public_transport, construction) NOT IN ('track', 'path') AND
+                              service NOT IN ('driveway', 'parking_aisle') THEN TRUE
+                         WHEN man_made = 'pier' AND NOT ST_IsClosed(geometry) THEN TRUE -- Only linear piers, not pier areas
```
Falke-Design, line 370: `AND service NOT IN ('driveway', 'parking_aisle')` is missing here

layers/transportation/transportation.sql
```diff
@@ -358,26 +358,23 @@ FROM (
...
-                            man_made <> 'pier'
-                            OR
-                            NOT ST_IsClosed(geometry)
-                        )
-             )
+           AND
+               CASE WHEN zoom_level = 12 THEN
+                    CASE WHEN highway_class(highway, public_transport, construction) NOT IN ('track', 'path', 'minor', 'service') THEN TRUE
+                         WHEN highway IN ('unclassified', 'residential') THEN TRUE
+                    END
```
Falke-Design, line 365: `AND man_made <> 'pier'` is missing

layers/transportation/transportation.sql
```diff
@@ -358,26 +358,23 @@ FROM (
...
+
+                    WHEN zoom_level = 13 THEN
+                    CASE WHEN highway_class(highway, public_transport, construction) NOT IN ('track', 'path') AND
+                              service NOT IN ('driveway', 'parking_aisle') THEN TRUE
+                         WHEN man_made = 'pier' AND NOT ST_IsClosed(geometry) THEN TRUE -- Only linear piers, not pier areas
+                    END
+
+                    WHEN zoom_level = 14 THEN
```
Falke-Design, line 373:
```suggestion
                    WHEN zoom_level >= 14 THEN
```

Falke-Design (changes requested):

layers/transportation/transportation.sql
ZeLonewolf: I don't understand this comment - that exact line is present at line 369.  Can you clarify?

ZeLonewolf (commented):
Thanks for detailed review

layers/transportation/transportation.sql
Falke-Design: Yes, but line 369 is another `WHEN` then line 370. So it is needed here too

layers/transportation/transportation.sql
```diff
@@ -358,26 +359,23 @@ FROM (
...
-                 OR zoom_level >= 14
-                         AND (
-                            man_made <> 'pier'
-                            OR
-                            NOT ST_IsClosed(geometry)
-                        )
-             )
+           AND
+               CASE WHEN zoom_level = 12 THEN
+                    CASE WHEN highway IN ('unclassified', 'residential') THEN TRUE
```
Falke-Design, line 364: `AND man_made <> 'pier'` is missing

layers/transportation/transportation.sql
```diff
@@ -358,26 +359,23 @@ FROM (
...
+                         WHEN highway_class(highway, public_transport, man_made, construction) NOT IN ('track', 'path', 'minor', 'service', 'service_construction', 'pier') THEN TRUE
+                    END
+
+                    WHEN zoom_level = 13 THEN
+                    CASE WHEN highway = 'service' AND service NOT IN ('driveway', 'parking_aisle') THEN TRUE
```
Falke-Design, line 369: `highway = 'service'` is new? Does this break some current logic?

layers/transportation/transportation.sql
```diff
@@ -358,26 +359,23 @@ FROM (
...
+                         WHEN man_made = 'pier' AND NOT ST_IsClosed(geometry) THEN TRUE  -- Only linear piers, not pier areas
```
Falke-Design, line 370: `AND service NOT IN ('driveway', 'parking_aisle')` is missing

layers/transportation/transportation.sql
```diff
@@ -358,26 +359,23 @@ FROM (
...
+                         WHEN highway_class(highway, public_transport, man_made, construction) NOT IN ('track', 'path', 'service', 'service_construction', 'pier') THEN TRUE
```
Falke-Design, line 371: `AND service NOT IN ('driveway', 'parking_aisle')` is missing

Falke-Design (changes requested):
This now very complex too, I don't know if it makes sense to change from complex to complex and then add maybe some unwanted bugs

layers/transportation/transportation.sql
ZeLonewolf: Where are you saying it's missing?  It's covered in 369 (include all service roads other than driveway/parking_aisle) and line 371 (exclude service roads)

layers/transportation/transportation.sql
ZeLonewolf: Are you saying for the case of `highway=unclassified` + `man_made=pier`?  I don't see that being a necessary check.  If it's tagged as unclassified/residential, adding pier tagging shouldn't pull it out of z12.

layers/transportation/transportation.sql
ZeLonewolf: I don't think so?  You need that check, otherwise you capture `highway=` values other than `service`.  Another option is to do a `service <> ''` instead but I felt the explicit highway check made things more clear.

layers/transportation/transportation.sql
ZeLonewolf: `service` is a sub-tagging of `highway=service`, it's not clear to me why you would need a check outside of `highway=service` filters.

ZeLonewolf:
Okay, I agree I haven't really solved the "too confusing" problem here, let me have another go at it.

layers/transportation/transportation.sql
Falke-Design: Ah okay, I didn't know this, I only compared the old sql with the new one

layers/transportation/transportation.sql
Falke-Design: Ah okay, I didn't know this, I only compared the old sql with the new one

layers/transportation/transportation.sql
Falke-Design: Same here, I didn't know this, I only compared the old sql with the new one. So this looks valid with your explanation

layers/transportation/transportation.sql
Falke-Design: It was in the old sql checked. I don't know if it makes sense. But if you say it is not necessary, then I believe you.

ZeLonewolf:
Okay, I think I've got it sorted.  Hopefully this is easier to look at this time.  I was able to compact the changes and also come up with a cleaner solution to the `pier` and `service` filters.  I did find some bugs as noted in the top comment which I suspect is the reason for the -0.1% change at z12 (which was inadvertently bringing in construction roads that it shouldn't have).

TomPohys:
Great work! Thank you very much!

TomPohys (approved):

PR #1208, opened 2021-09-03
Remove landuse=park from landcover

ZeLonewolf:
The tag `landuse=park` is a rare tagging mistake of `leisure=park`.  The OSM wiki [lists this tag as a mistake](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dpark).  It is included in the landcover layer with a note that there were (at the time) 600 parks tagged in this way.  There are currently a bit over 100 objects tagged with this mistake.  Based on the wiki documentation and obvious decline of this tag, we should remove it so that we're not introducing mapping mistakes into the database and not inadvertently providing mapper feedback that encourages the use of mistake tagging.


TomPohys:
Good catch! Thank you!

TomPohys (approved):

PR #1213, opened 2021-09-06
Replace osmborder with imposm/SQL

### LINKED ISSUE #810: SQL query-based alternative to osmborder
Per the discussion at https://github.com/openmaptiles/openmaptiles-tools/issues/215#issuecomment-613870437 and on [slack](https://osmus.slack.com/archives/CH6GFAFRC/p1586970993074400) , the `boundary` layer may, or may not, be able to use SQL to perform the tasks currently done by [`osmborder`](https://github.com/pnorman/osmborder).

This issue is opened as a discussion platform. Please add the appropriate labels.

#### The advantages for such a change are:
- Remove the dependency on [`osmborder`](https://github.com/pnorman/osmborder), which is an archived, read-only repository.
- Use PostGIS to replace multiple `pbf` file scans using `osmium`: 3 scans on the full OSM data and 3 scans on the administrative boundaries data.

#### Schema generation flow
The current schema flow for the `boundary` layer is

import-borders

In the proposed approach, the schema flow for the `boundary` layer would be

no-import-borders

#### Evaluation steps
Following @nyurik advice, the steps to evaluate this approach _could_ be

1. Understand the current functionality of [`osmborder`](https://github.com/pnorman/osmborder), with the help of @pnorman, including:
   - If the same "way" is used multiple times by different admin-level relations, only include it once with the lowest admin-level?
   - Do not assume polygons are valid/closed -- work only with the line segments (ways)?
   - How is the `disputed` field calculated?
   - How is the `maritime` field calculated?
   - What other functionality, if any, is performed by  [`osmborder`](https://github.com/pnorman/osmborder)?
2. Use modified mapping and SQL code to create an additional materialized view that should be equivalent to the `osm_border_linestring` table imported from [`osmborder`](https://github.com/pnorman/osmborder).
   - Use SQL to compare them.
   - Keep the existing schema generation flow unchanged.
   - Note: the `dividing_line` field is not used by OMT and need not be re-implemented in SQL.
3. Modify the schema generation flow, including
   - Consolidating potential definition differences of `disputed` and `maritime` between the `osm_border_linestring` and `osm_border_disp_linestring` materilized views.
   - Performing performance measurements on speed and size.

### LINKED ISSUE #1156: Boundary layer - move data import into mapping.yaml
After discussion on Slack (osmus.slack.com), it would be good to move importing borders from omt-tools (using https://github.com/pnorman/osmborder) to `mapping.yaml`. It could be probably joined with the mapping we do now for disputed borders (maybe just remove there `claimed_by: [__any__]` ?) 
https://github.com/openmaptiles/openmaptiles/blob/master/layers/boundary/mapping.yaml#L89-L124
This way we would have only `osm_border_linestring` table, no `osm_border_disp_linestring` anymore.
___
Comment of @pnorman:

I was looking at if osmborder can be replaced by imposm functionality for work, and it turns out it can - at least partially. This can be done by creating a relation_member table with imposm,
```
tables:
  admin_members:
    type: relation_member
    columns:
    - name: osm_id
      type: id
    - name: member
      type: member_id
    - name: type
      type: member_type
    - name: admin_level
      key: admin_level
      type: integer
    - name: geometry
      type: geometry
    relation_types: [boundary]
    mapping:
      boundary: [administrative]
```
The SQL to generate tiles can be something like
```
SELECT DISTINCT ON (member, type)
geometry, admin_level
FROM osm_admin_members
WHERE type = 1
AND admin_level > 0
ORDER BY member, type, admin_level ASC;
```
___
This change enables to do attribute edits on import, requested e.g. in https://github.com/openmaptiles/openmaptiles/issues/623.

### LINKED ISSUE #1228: borders should be removed each time quickstart is run
When running `quickstart.sh` the `data/borders` folder is not removed if it exists. This can be a problem when a user is running `quickstart.sh` multiple times for different downloads such as Washington and then Arizona:

> ./quickstart.sh washington

This will correctly create borders for Washington in the borders folder.

> ./quickstart.sh arizona

This will use the old borders files that were created for Washington which causes incorrect data to be written for that state, and missing data as well.

**What Needs to be done**

The `data/borders` folder needs to be removed if it exists when running `quickstart.sh`.

ZeLonewolf:
Fixes #1156
Fixes #810
Fixes #1228

This PR replaces `osmborder`, which is no longer maintained, with `imposm` mappings and SQL code to generate borders.  Key features that were moved into the imposm/SQL layer:
1. Grouping by `osm_id` and aggregating by lowest `admin_level` value so that there's only one copy of ways that are members of multiple relations.
2. Filtering out of point features in boundary relations (typically `admin_centre` and `label` roles).
3. Move disputed boundary detection logic into SQL.

This will increase the database size slightly because of the limits of what imposm can do, as some of the filtering is done in the SQL layer after importing, rather than being done in `osmborder`.

**Rendering Samples**

Rendering sample from a disputed alpine boundary along the Italy/France border:

image

San Marino enclave boundary:

image

Italy maritime boundary:

image


ZeLonewolf:
Note that CI appears to be broken currently, with "permission denied" errors.

ZeLonewolf:
@TomPohys I can't seem to figure out why CI is not working here.  I am able to run `./quickstart monaco` locally and have it run.

Falke-Design:
Currently i have generated a borders line.csv file of the planet and import them into my openmaptiles where I have only the osm data for germany. With your change this not longer working. It will need for every generation the planet.osm.pbf which took much longer or i can't see all borders.

But with this changes i think it is possible to get the borders out of the Database where multiple .pbf files are inserted. Which was not possible before, because only one .pbf file can pe passed to import-borders.

Maybe the borders of the countries can be pre added to openmaptiles-tools and imported. For local boundaries, the one of the osm data are used and added to the country one.

ZeLonewolf:
That is a curious use case, where you want to see global boundaries but map data only for one country.  It seems like this could be accomplished with an overpass query to download all boundaries, and them compress that to a pbf file.  Or, you could download the planet file and use osmosis to extract only boundary data from it.  In any case, osmborder is no longer maintained, so there is a desire to move away from it in the project.

On a side note, there are also some more advanced things with boundaries that OpenMapTiles can't do today, such as [name labels on boundaries below the national level](https://www.openstreetmap.org/#map=14/50.5563/10.1028) that I'm not sure we can do without having the processing logic live in the SQL.

Falke-Design:
My goal is that I have no blank tiles on the lower zoom levels, which I had managed by adding https://github.com/lukasmartinelli/naturalearthtiles and additionally I want to show the borders of countries so it has more structure.

Maybe you are right and I have done this on a other way. I will check this when I'm home.

Falke-Design:
Ok my posts where **bullshit** 😄 I thought that the borders until z4 are generated from `import-borders` so I used the complete planet file. But in the `boundary.yaml` is the comment:
> Until z4 [Natural Earth data](http://www.naturalearthdata.com/downloads/) is used

I will create a PR to add this to the Readme.

Now I can appriciate this PR and this workflow because it makes it possible to get the boundary from multiple imported osm files. This was the other reason why I had imported the full planet borders.

layers/boundary/boundary_name.sql
```diff
@@ -32,7 +32,7 @@ CREATE TABLE IF NOT EXISTS osm_border_linestring_adm AS (
   			FROM (SELECT ST_Node(
                           ST_Collect(geometry)) AS geometry
     			FROM osm_border_linestring
-    			WHERE admin_level = 2
+    		        WHERE admin_level = 2 AND ST_Dimension(geometry) = 1
```
nyurik, line 35: lets keep the same tab-indentation as before for consistency. TBH, I prefer spaces myself, but since the rest of the code uses tabs, lets keep that

layers/boundary/mapping.yaml
```diff
@@ -1,124 +1,79 @@
...
     - key: admin_level
       name: admin_level
       type: integer
+    - key: disputed_by
+      name: disputed_by
+      type: string
+      from_member: true
     - key: claimed_by
       name: claimed_by
       type: string
```
nyurik, line 44: for ease-of-comparison, let's move this below (unless the order is important for some other reason?)
```suggestion
    - key: claimed_by
      name: claimed_by
      type: string
    - key: disputed_by
      name: disputed_by
      type: string
      from_member: true
```

nyurik (commented):
i really like this effort, thanks for working on it!  For simplicity, lets keep just the needed changes here (i.e. keep tabs as tabs, etc).  What else is needed here to change the status from draft to a PR?

layers/boundary/boundary_name.sql
ZeLonewolf: The boundary layer has a mix of tabs and spaces, so I'd been converting tabs to spaces on lines that I'm editing anyways.  On a  separate note, would you accept a PR to convert tabs to spaces across the layer? :-D

ZeLonewolf:
> What else is needed here to change the status from draft to a PR?

I still need to properly test left/right national boundary labels and disputed boundary tagging.  This PR Is a pain to test because I need to render multinational areas to prove that I haven't broken anything.  I've been using Italy as my example because it has enclave countries, and I also tried to use the Europe extract  but I ended up with a full disk.  Each of these big renders takes a long time, basically overnight, so there's a longer than usual lag time here  :)

My latest overnight test:

image

ZeLonewolf:
See now, this is looking much better:
image

Falke-Design:
I don't know if this a common way to name this but i think the name admin-left & right can be confusing. On the left side of the boundary it is correct named but on the right side of the boundary the left / right names should be switched. I think better names were "in" and "out" or something like that.

ZeLonewolf:
> I don't know if this a common way to name this but i think the name admin-left & right can be confusing. On the left side of the boundary it is correct named but on the right side of the boundary the left / right names should be switched. I think better names were "in" and "out" or something like that.

This discussion is a bit out of scope for this PR since this is a re-write that doesn't change functionality, but here goes.  "left" and "right" are with respect to the direction of an individual way, so there is no concept of "inner" and "outer" possible.  Unless you are thinking of whether a boundary member way is an "inner" or "outer" of its respective multipolygon -- that is not exposed to the layer.

I propose in #1216 to fix boundary labelling by changing the data representation to closed LineString objects, overlapping at each `admin_level`, and normalized with right-hand-rule winding.  With that implemented, the inner side of a polygon would always fall on the left side of each line segment -- the outers wind counter-clockwise and the inner wind clockwise.  This would also allow you to place full name labels at all boundary levels, not just 3-letter codes only on national boundaries.  With that implemented, we could remove the left/right scheme currently in the boundary layer (in V4.0, as that would be a breaking change) and possibly fold in the 3-letter codes into the new name layer if it's still desired to keep it.

ZeLonewolf:
Converting back to draft because I noticed a few minor issues that I should be able to resolve tonight.

ZeLonewolf:
> Note that boundaries are not always closed polygons, frequently due to errors in OSM data

Good point.  We'll have to segregate the unclosed boundaries and do some PostGIS gymnastics to handle those cases.  We should be able to make use of the code in `boundary_name.sql` which already deals with LineString left/right processing using projected line segments.

ZeLonewolf:
Ready for review.  I suspect there should be an extra `ST_LineMerge()` thrown in here somewhere, but I can't figure out how to do it without totally breaking things, and the renders seem like they look right.  So I think we're good to go!

Falke-Design:
If I want to import `liechtenstein` I get following error while running `import-sql`
```
psql:/sql/parallel/boundary.sql:105: ERROR:  relation "ne_10m_admin_0_countries" does not exist
LINE 39:     ne_10m_admin_0_countries AS ne
```

Are there any dependencies to other layers? Or should it work if I have only the `boundary.yaml` in the Tileset?

ZeLonewolf:
@Falke-Design I was able to run `./quickstart.sh liechtenstein` to completion with no issues with all layers except boundary disabled.

Is it possible we're running into the same issue as gravitystorm/openstreetmap-carto#4461 ?

ZeLonewolf:
@Falke-Design also, you will need to run `make import-data` in order to populate the Natural Earth tables.

ZeLonewolf:
> The problem was, that I forgot to run `import-data`. Is it possible to make the `import-sql` working without going wrong when `import-data` was not run? A check if the tables of `import-data` existing before they are used in a statement?
> 
> For testing the layers it would maybe easier

This is a good idea - though Natural Earth data is used in a number of layers, not just boundary, so this would need to be in multiple places.  I've documented this in #1220.

Falke-Design:
I have generated all borders of the planet: [Download](https://www.dropbox.com/s/qi75svhsf5j5jxb/borders.mbtiles?dl=0)

It looks good. The only thing that I saw is that the borders around the continents are not loaded on the low zoom levels. I think that are the marine / sea borders.

borders

ZeLonewolf:
Converting to draft to investigate the issue with low-zoom maritime/country borders.

ZeLonewolf:
Thank you @Falke-Design for the test render!

I've run the mbtiles you provided and reviewed the before and after code, and I believe it's actually exhibiting the correct behavior.

Here is a client render from the "bright" style: https://openmaptiles.github.io/osm-bright-gl-style/#3/25.89/-12.71

Maritime borders do not start until zoom 4, which is the same as the test mbtiles.  Maritime borders are ocean boundaries that are a distance away from each country's actual coastline (typically, 12 miles).  Thus, only the land parts of a country's border are rendered at low zooms (0-3), with the water layer used to "outline" the continents.

While coastline itself is rendered using the `water` layer (and not included in the boundary layer for `admin_level=2`), it is common that lower-level boundaries (`admin_level=4` and higher) will have the coastline be part of the boundary.  This is why you have the appearance of the coastline "disappearing" when you move down to the zoom in which lower level boundaries are no longer rendered.  Had you rendered the `water` layer also, this would make more sense visually.

During this review, I identified that I was unnecessarily mapping `natural=coastline` and `boundary=administrative` into columns.  These are not needed because `natural=coastline` is not part of the boundary logic and `boundary=administrative` is only needed as an imposm filter on import, so I've made those changes in order to reduce the size of the database.

This PR is now ready for review.

TomPohys:
Awesome work! It looks great! Thank you very much!

zstadler:
Thank you @ZeLonewolf for resolving an issue that was open more than a year!

PR #1215, opened 2021-09-07
Add boundaries `import-data` info to Readme

Falke-Design:
I was thought that the borders are completley imported over `import-borders` but until z4 the Natural Earth Data are used.
I added this info to the Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ make start-db
 ```
 
 Import external data from [OpenStreetMapData](http://osmdata.openstreetmap.de/), [Natural Earth](http://www.naturalearthdata.com/) and [OpenStreetMap Lake Labels](https://github.com/lukasmartinelli/osm-lakelines).
-
+The boundaries of the countries are used from Natural Earth until zoom level 4.
 ```bash
 make import-data
 ```


nyurik:
I don't think this is the right place for this info. README describes steps needed to setup everything. The specifics of what gets used where should probably go to the specific layer that uses that data, or possibly a detailed section on the structure of the data (if we have that?).

Falke-Design:
I would agree with you in principle, but I was not aware of this and had imported the borders for the whole planet. Because I didn't know that it is already covered by the Natural Earth data. And the info in the `boundary.yaml` is not enough because a newbie don't know how it is working with the layers.

TomPohys:
Hi, I do not think, that this is the right place to add info just about the `boundary` layer. Then it looks like, that only `boundary` layer up to zoom 4 is gaining data from NaturalEarth. 

But maybe something like "Please look on `etl_diagram.png` in every layer (e.g. [/layers/boundary/etl_diagram.png](https://github.com/openmaptiles/openmaptiles/blob/master/layers/boundary/etl_diagram.png)) to view the data source for every zoom. `osm_*` table from OSM and `ne_*` table from Natural Earth"

But thank you for pointing this out, it could be mention more explicitly.

Issue #1211, opened 2021-09-05
Explanation how to create own layer structures / schemas


Falke-Design:
Hi all,
I created a repo that should explain and help beginners (like I'm) to understand how own layers can be created. It was not easy for me to understand how this is working and to generate my first tile (also because the sample (skiing) does not work / run).

To the **Maintainers**: It would be nice if you can copy my repo and add it to the openmaptiles project (No need to fork).
@TomPohys you are currently one of the most active maintainers, what do you think?

To the **contributers**: There are many TODOs and I can't solve them all, because I don't understand how some parts are working and what happing there. Like I said I'm a beginner. It would be very nice if you can help! ❤

The Repo: https://github.com/Falke-Design/openmaptiles-layers

nyurik:
@Falke-Design can this be submitted as a pull request to the `openmaptiles-tools` repo as a `docs/` directory?  I don't think there is any reason to create another repo just for a few documentation files just yet, but it is tools repo that has all the scripts to process the yaml files. Docs and examples are always welcome! You may also want to link to it from the main repo's README.

Falke-Design:
@nyurik Sure I will do if you want. For the documentation part it will make complete sense. But one of the other thoughts was, that also some other samples and nice-to-have layers can be added to this repo. Which makes the openmaptiles-tools bigger and adds stuff which the most people don't need.

nyurik:
@Falke-Design i think we should separate those two things (both are valid).  Documentation should go into tools because that's where all the tooling is. Proposed layers should go into the OMT itself, and this is where we had a number of discussions of how to implement optional layers. I don't think we know how to do it just yet - there has been some discussion on the topic, and everyone recognizes it as a valid need, but clearly it should not be a part of documentation. Take a look at #970

PR #1219, opened 2021-09-10
Add layer / schema explanation

Falke-Design:
This should explain and help beginners (like I'm) to understand how own layers can be created.

There are many **TODOs** and I can't solve them all, because I don't understand how some parts are working and what happing there. It would be very nice if you can help!

Close: #1211

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,8 @@ make
 make import-sql
 ```
 
+More about layers and their creation: [Doc LAYERS](https://github.com/openmaptiles/openmaptiles/docs/LAYERS.md)
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -146,6 +146,8 @@ make
 make import-sql
 ```
 
+More about layers and their creation: [Doc LAYERS](https://github.com/openmaptiles/openmaptiles-tools/docs/LAYERS.md)
```
ZeLonewolf, line 149: This is not currently a valid URL.  Is there a PR somewhere that adds this file?

ZeLonewolf:
Since the openmaptiles-tools PR that adds this README is draft, this PR should be marked draft also.

PR #1222, opened 2021-09-11
Makefile: Replace static DB properties with the variables

Falke-Design:


nyurik:
I'm not too sure this will work -- the `PG*` variables do not exist on the host, only inside the container. Using them in the makefile may generate empty values. I suspect it will work if you let the expansion happen inside docker by using shell expansion instead, something like `--host="$${PGHOST}"` - but this needs to be tested

P.S. Considering that we want to get rid of the tm2source/mapnik altogether, i'm not even sure its worth even dealing with

Falke-Design:
The `PG*` vars are always existing because `.env` is included in the Makefile. And I don't think that we should suppot the old var style because they are not defined in the .env

nyurik:
good point about the .env file, i did not realize it was included (even though I was the one who added it there :) ).

I disagree about vars not defined -- anyone who uses their own .env file could still have those settings. So either we should remove their support everywhere in the tools, and officially declare them as unsupported, or we should consistently support them everywhere.

Falke-Design:
I added a check about the old var styles. If one of the old variable is existing, then it is used else the new one. Same as in the tools [scripts](https://github.com/openmaptiles/openmaptiles-tools/blob/3b2a0675294d2db35ce432f51a0cdddf9ea6b073/bin/import-sql#L10): 
grafik

Solution in the Makefile:
```
# use the old postgres connection values if they are existing
PGHOST := $(or $(POSTGRES_HOST),$(PGHOST))
PGPORT := $(or $(POSTGRES_PORT),$(PGPORT))
PGDATABASE := $(or $(POSTGRES_DB),$(PGDATABASE))
PGUSER := $(or $(POSTGRES_USER),$(PGUSER))
PGPASSWORD := $(or $(POSTGRES_PASSWORD),$(PGPASSWORD))
```


I think it is valid to support both variable names but I think also that [after 2 years](https://github.com/openmaptiles/openmaptiles-tools/commit/6d9ed08810c8d9f681b4bcfcbebe97d74ef864cc#diff-5f0abe7b60c5d1fbd319fb601dc71e2361fb893deae0a394b3d1261b2dab01ba) it is legit to remove the old var styles. Maybe for a major relase in future. 

A simple test which can added to the Makefile to see which value is taken:
```
.PHONY: test
test:
	@echo "HOST: $(PGHOST)"
	@echo "PGPORT: $(PGPORT)"
	@echo "PGDATABASE: $(PGDATABASE)"
	@echo "PGUSER: $(PGUSER)"
	@echo "PGPASSWORD: $(PGPASSWORD)"
```

nyurik (approved):
looks good to me!  I agree that in the next major we should probably remove support for the legacy PG vars

TomPohys (approved):

PR #1187, opened 2021-08-21
Add international airports at z8-9

ZeLonewolf:
Closes #1082

This PR moves the aerodrome/aerodrome_type consolidation (via template) into the lower-level `osm_aerodrome_label_point` table, and creates a partial index on class `international` for airports with an `iata` code.  The presence of an `iata` code is more significant than an `icao` code which exists at many smaller, general aviation airport.  Therefore, `international` + `iata` should be a sufficient filter.

Zoom 8 international airports in the NYC area (blue dots):
image

Zoom 10 in the same area:
image

PR #1223, opened 2021-09-11
Update missing devdocs

ZeLonewolf:
This PR fixes missing devdoc links introduced in #1187 and #1202.  Thanks @zstadler for finding the issue.


nyurik (approved):
thanks!

Issue #43, opened 2016-11-18
Define layer ID in Tileset YAML not in Layer YAML


lukasmartinelli:
The ID of the layer should be defined in the tileset not in the layer YAML.
This means you can pull in the layers as you want and still give them a custom name.

So not like this.

```yaml
tileset:
  layers:
    - layers/boundary/boundary.yaml
    - layers/highway/highway.yaml
    - layers/highway_name/highway_name.yaml
    - layers/building/building.yaml
    - layers/housenumber/housenumber.yaml
    - layers/place/place.yaml
```

More like this.

```yaml
tileset:
  layers:
    boundary: layers/boundary/boundary.yaml
    street: layers/highway/highway.yaml
    streetname: layers/highway_name/highway_name.yaml
    house: layers/building/building.yaml
    housenumber: layers/housenumber/housenumber.yaml
    place: layers/place/place.yaml
```

nyurik:
Layer ID is used by other layers to establish dependencies, so each layer has to be reliably referenced. Won't implement

PR #1224, opened 2021-09-12
Moved layer-id from layer definition into tileset

Falke-Design:
More informations in this PR: https://github.com/openmaptiles/openmaptiles-tools/pull/371

Fix: #43


nyurik:
Currently layers can require other layers to be present.  I think it would be conceptually wrong for a layer to reference another layer by ID, but not to have its own

Falke-Design:
I think the main idea behind #43 is good but you are right if there are references they need ids on the layer level. 
I suggest to close #43

nyurik:
Yep, makes sense, closed

PR #1230, opened 2021-09-14
Fix SQL for DIFF_MODE=true

zstadler:
Related to comments for b011b27

### DIFF
--- a/layers/transportation_name/update_transportation_name.sql
+++ b/layers/transportation_name/update_transportation_name.sql
@@ -48,7 +48,7 @@ FROM (
         NULLIF(rm3.network, '') || '=' || COALESCE(rm3.ref, '') AS route_3,
         NULLIF(rm4.network, '') || '=' || COALESCE(rm4.ref, '') AS route_4,
         NULLIF(rm5.network, '') || '=' || COALESCE(rm5.ref, '') AS route_5,
-        NULLIF(rm6.network, '') || '=' || NULLIF(rm6.ref, '') AS route_6,
+        NULLIF(rm6.network, '') || '=' || COALESCE(rm6.ref, '') AS route_6,
         hl.z_order
     FROM osm_highway_linestring hl
             LEFT OUTER JOIN osm_route_member rm1 ON rm1.member = hl.osm_id AND rm1.concurrency_index=1
@@ -353,7 +353,7 @@ BEGIN
 	    NULLIF(rm3.network, '') || '=' || COALESCE(rm3.ref, '') AS route_3,
 	    NULLIF(rm4.network, '') || '=' || COALESCE(rm4.ref, '') AS route_4,
 	    NULLIF(rm5.network, '') || '=' || COALESCE(rm5.ref, '') AS route_5,
-	    NULLIF(rm6.network, '') || '=' || NULLIF(rm6.ref, '') AS route_6,
+	    NULLIF(rm6.network, '') || '=' || COALESCE(rm6.ref, '') AS route_6,
             hl.z_order
         FROM osm_highway_linestring hl
                 JOIN transportation_name.network_changes AS c ON
--- a/layers/transportation_name/update_transportation_name.sql
+++ b/layers/transportation_name/update_transportation_name.sql
@@ -336,7 +336,7 @@ BEGIN
             CASE WHEN length(hl.name_en) > 15 THEN osml10n_street_abbrev_en(hl.name_en) ELSE NULLIF(hl.name_en, '') END AS name_en,
             CASE WHEN length(hl.name_de) > 15 THEN osml10n_street_abbrev_de(hl.name_de) ELSE NULLIF(hl.name_de, '') END AS name_de,
             slice_language_tags(hl.tags) AS tags,
-            rm.network_type,
+            rm1.network_type,
             CASE
                 WHEN rm1.network_type IS NOT NULL AND nullif(rm1.ref::text, '') IS NOT NULL
                     THEN rm1.ref::text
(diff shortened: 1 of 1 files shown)


zstadler:
@ZeLonewolf,

Please review the code.

layers/transportation_name/update_transportation_name.sql
```diff
@@ -348,13 +348,23 @@ BEGIN
             CASE WHEN highway IN ('footway', 'steps') THEN layer END AS layer,
             CASE WHEN highway IN ('footway', 'steps') THEN level END AS level,
             CASE WHEN highway IN ('footway', 'steps') THEN indoor END AS indoor,
-            route_1, route_2, route_3, route_4, route_5, route_6,
+	    NULLIF(rm1.network, '') || '=' || COALESCE(rm1.ref, '') AS route_1,
+	    NULLIF(rm2.network, '') || '=' || COALESCE(rm2.ref, '') AS route_2,
+	    NULLIF(rm3.network, '') || '=' || COALESCE(rm3.ref, '') AS route_3,
+	    NULLIF(rm4.network, '') || '=' || COALESCE(rm4.ref, '') AS route_4,
+	    NULLIF(rm5.network, '') || '=' || COALESCE(rm5.ref, '') AS route_5,
+	    NULLIF(rm6.network, '') || '=' || NULLIF(rm6.ref, '') AS route_6,
```
ZeLonewolf, line 356: Did you mean to change the pattern here?

zstadler: This change and the following change were adapted from [this code](https://github.com/openmaptiles/openmaptiles/blob/5e37a811fb7d407a1c7ec3884f742de9eba91f72/layers/transportation_name/update_transportation_name.sql#L46-L60)

```sql
        NULLIF(rm1.network, '') || '=' || COALESCE(rm1.ref, '') AS route_1,
        NULLIF(rm2.network, '') || '=' || COALESCE(rm2.ref, '') AS route_2,
        NULLIF(rm3.network, '') || '=' || COALESCE(rm3.ref, '') AS route_3,
        NULLIF(rm4.network, '') || '=' || COALESCE(rm4.ref, '') AS route_4,
        NULLIF(rm5.network, '') || '=' || COALESCE(rm5.ref, '') AS route_5,
        NULLIF(rm6.network, '') || '=' || NULLIF(rm6.ref, '') AS route_6,
        hl.z_order
    FROM osm_highway_linestring hl
            LEFT OUTER JOIN osm_route_member rm1 ON rm1.member = hl.osm_id AND rm1.concurrency_index=1
            LEFT OUTER JOIN osm_route_member rm2 ON rm2.member = hl.osm_id AND rm2.concurrency_index=2
            LEFT OUTER JOIN osm_route_member rm3 ON rm3.member = hl.osm_id AND rm3.concurrency_index=3
            LEFT OUTER JOIN osm_route_member rm4 ON rm4.member = hl.osm_id AND rm4.concurrency_index=4
            LEFT OUTER JOIN osm_route_member rm5 ON rm5.member = hl.osm_id AND rm5.concurrency_index=5
            LEFT OUTER JOIN osm_route_member rm6 ON rm6.member = hl.osm_id AND rm6.concurrency_index=6
    WHERE (hl.name <> '' OR hl.ref <> '' OR rm1.ref <> '' OR rm1.network <> '')
```

As far as I could see, neither `osm_highway_linestring` aka `hl` nor `osm_route_member` aka `rm` in the original code have the `route_1` to `route_6` fields.

Are there any other options for this code?

zstadler: Did you mean that both here and [the original code](https://github.com/openmaptiles/openmaptiles/blob/5e37a811fb7d407a1c7ec3884f742de9eba91f72/layers/transportation_name/update_transportation_name.sql#L51) should use `COALESCE(rm6.ref, '')` rather than `NULLIF(rm6.ref, '')`?

ZeLonewolf: I think I did this rather poorly on the first pass.  I think what we really want here is just a simple `concat()`, which ignores null arguments.  As currently written, it will drop values which have a ref but no network, which doesn't seem desirable.

ZeLonewolf:
Overall this seems reasonable, there's nothing obvious that jumps out at me as wrong based on a code inspection (other than my own misgivings about how I did the route concatenation stuff).  I would advise that you use `indiana` as your test area, because it has several roads that go up to `route_6`.

zstadler:
@ZeLonewolf ,

I've tested the latest code with the Indiana data, as you suggested. Although there are no syntax errors, the update code does not work properly. After the import of the update, all rows have NULL values in `route_6`, `route_5`, and `route 4`.

1. I've downloaded the 1-Aug data for Indiana and ran `./quickstart.sh` in `DIFF_MODE=true` with it. 
  - I use [DBeaver](https://dbeaver.io/) to examine the `osm_transportation_name_network` table
  - The table has 242,960 rows.
  - 150 rows have non-NULL `route_6` values and all of the 150 have non-NULL values for `route_1` to `route 5`. 
2. I've downloaded the changes since then and applied them using `make import-diff`.
  - The table now has 243,519 rows
  - All rows have NULL values in `route_6`, `route_5`, and `route 4`.
  - 16 rows have non-NULL `route_3` values and all of them have NULL in all other `route_*` fields.
  - Only 24 of the 264 rows where `route_2 is not null` also have non-NULL value in `route_1`.
3. I've ran a clean `./quickstart.sh` in `DIFF_MODE=true` and `DIFF_MODE=false` with the latest Indiana data. In both runs:
  - The table has 243,684 rows
  - 150 rows have non-NULL `route_6` values and all of the 150 have non-NULL values for `route_1` to `route 5`. 

I don't know how to make the update code work.
Perhaps @frodrigo, who I has implemented #987, could help.

### To reproduce
#### Commit your WIP into a separate branch or stash it

#### One-time setup:
```
git checkout master
git restore .
git pull master
git fetch origin pull/1230/head
git checkout -b DIFF_MODE-fix FETCH_HEAD
echo DIFF_MODE=true >> .env
echo MAX_ZOOM=14 >> .env
export area=north-america/us/indiana
```

#### Build 1-month-old tiles:
```
rm -fr data build cache
make download-geofabrik area=$area
docker-compose run --rm --user=$(id -u):$(id -g) openmaptiles-tools sh -c "wget -O data/$area.osm.pbf http://download.geofabrik.de/$area-$(date --date="$(date +%Y-%m-15) -1 month" +'%y%m01').osm.pbf"
./quickstart.sh $area
```
#### Update with the changes since then:
```
docker-compose run --rm --user=$(id -u):$(id -g) openmaptiles-tools sh -c "osmupdate --base-url=$(sed -n 's/ *\"replication_url\": //p' data/$area.repl.json) data/$area.osm.pbf data/changes.osc.gz";
make import-diff
make generate-tiles-pg
```

ZeLonewolf:
Thanks for the detailed procedure, @zstadler!  I've been trying to figure out the problem, and I haven't found the bug yet.  But I have found at least one bug with the route membership code though I don't think it's related.  I am continuing to work on this.

ZeLonewolf:
I found at least one problem.  There is no code present to update `concurrency_index` in the table `osm_route_member`.  There should be a trigger added to that table to re-compute `concurrency_index` (which is a calculated table) when rows are changed.  This would explain why the various `route_X` columns aren't getting populated.

https://github.com/openmaptiles/openmaptiles/blob/8f96a598af8691f38204e055955a7ee6bf1fa1f8/layers/transportation/update_route_member.sql#L83-L90

ZeLonewolf:
Posted a draft PR, #1239 with additional fixes that I think will resolve the issue.

zstadler:
> 
> 
> Posted a draft PR, #1239 with additional fixes that I think will resolve the issue.

Closing

Issue #1226, opened 2021-09-12
Integrity tests to check the import-diff flow


zstadler:
While the Integrity tests check the `import-osm` flow, they do not test the `import-diff` flow. As a result, the SQL code that performs table updates is not tested.

[Replace materialized view of water by tables with diff update #853](https://github.com/openmaptiles/openmaptiles/pull/853) and similar PRs have continuously increased the complexity of the code that is used only in flows based on `import-diff`. Currently, this code is not tested by the integrity tests, including for typing errors.

Such a test requires finding a source changefile that performs modifications in all OMT tables for a relatively small area.

It would be nice to also add performance measurements for the flow.

Related to comments for b011b27e52726361d352932ad923de7e05893cc4

ZeLonewolf:
This would be a great addition.  I have personally broken the update scripts without realizing it and and not noticed it until weeks later.

zstadler:
Here is a working example for such a test using `area=greater-london`

```
echo DIFF_MODE=true >> .env
rm -fr data build cache
make download-geofabrik area=greater-london
docker-compose run --rm --user=$(id -u):$(id -g) openmaptiles-tools sh -c "wget -O data/greater-london.osm.pbf http://download.geofabrik.de/europe/great-britain/england/greater-london-$(date --date="$(date +%Y-%m-15) -2 month" +'%y%m01').osm.pbf"
./quickstart.sh greater-london
docker-compose run --rm --user=$(id -u):$(id -g) openmaptiles-tools sh -c "osmupdate --base-url=$(sed -n 's/ *\"replication_url\": //p' data/greater-london.repl.json) data/greater-london.osm.pbf data/changes.osc.gz"
make import-diff
make generate-tiles-pg MAX_ZOOM=14
```

PR #1231, opened 2021-09-14
Integrity check to generate zoom 14 tiles

zstadler:
Setting the MIN/MAX zoom levels in the `env` section had no effect since the values in the `.env` file have higher precedence.

See the currently effective `maxzoom = 7` value in [a typical integrity check log]( https://github.com/openmaptiles/openmaptiles/pull/1223/checks#step:3:3546).


zstadler:
@TomPohys,

Please review and merge this PR so I can submit a PR for [Integrity tests to check the import-diff flow #1226](https://github.com/openmaptiles/openmaptiles/issues/1226)

TomPohys (approved):

TomPohys:
Great! Thank you!

ZeLonewolf:
Since the integrity test and performance test run in parallel, it's a no-brainer to allow the integrity test to run down to z14, since it's still shorter to run than the performance test.

TomPohys:
Hi @ZeLonewolf, if I understand correctly (not necessary to run integrity test up to zoom 14), there is e.g. `housenumbers` layer which is only in zoom level 14. So I think it is the correct way to test integrity up to zoom level 14.

ZeLonewolf:
@TomPohys I was saying that the new integrity test up to zoom 14 won't slow down the continuous integration :)

TomPohys:
Ohh, OK. I am sorry for not understand it correctly :)

PR #1233, opened 2021-09-17
BUGFIX: Ignore duplicate route concurrencies

ZeLonewolf:
While troubleshooting #1230, I discovered that there were cases where a way was a member of the same route relation more than once, such as: https://www.openstreetmap.org/way/17439235

This segment would end up showing duplicate highway shields in a rendered style, which is not desirable.  While I personally think this style of tagging is wrong, it's a relatively easy fix to remove those duplicates.  This PR replaces `ROW_NUMBER()` with `DENSE_RANK()` in the concurrency index generation code and adds a `DISTINCT` constraint on the concurrency join to handle multiple member rows with the same concurrency index.

Since I was in this file, I also fixed the copy/paste error in generating the route names.

Screenshot from database showing correct mapping:
image


TomPohys:
Awesome! Thanks a lot!

TomPohys (approved):

Issue #1178, opened 2021-08-16
Approach for bifurcated rendering of high-zoom features


ZeLonewolf:
Would it be possible to define layers or features that only render at zooms 15+, so that they could still be a part of the schema, but don't get rendered into the mbtiles?  Then, they could still be served by postserve in a standard way for high zoom rendering, but not bloat the tiles.  That way we can still work to standardize the schema for smaller features but not burden pre-rendered tiles.  This could be a useful approach for things like barriers, trees (#405), street furniture (#395), building names (#241, #279), and other minor features that don't make sense to pack into zoom 14 tiles.

Challenges include:
1. Updating the schema documentation to make it very clear which features are in the zoom 14 and below tiles, versus which features are "high zoom".  This gives the community a clear indication of the separation between "high zoom" and "low zoom" features.
2. The current offering of MapTiler Cloud currently serves the entire OpenMapTiles schema.  However, if high zoom tiles were added, MapTiler Cloud would need to either add a postserve/direct-from-DB offering, create a separate tier of access for high zoom, or exclude it.  If managed well, this could be a business opportunity for MapTiler to service high-zoom use cases.

High zoom (aka "micro-mapped") features are of high interest to the openstreetmap-americana project, and represent one of our biggest future challenges that might force us to diverge from OpenMapTiles and run our own fork of the schema.  However, I think that bifurcated high-zoom rendering, especially in dense urban areas, is of broad interest to general-purpose styles and we should look at this challenge as an opportunity to take OpenMapTiles to the next step in version 4.0.

daliborjanak:
We are discussing adding z15 and we are doing some tests on it in the meantime. We would like to add it to the schema but we need to plan it deeply. There are a few things that can be problems like bigger import and database. We didn't have plans to bifurcated rendering now. IMHO it is not an issue of this project

nyurik:
@daliborjanak do you plan to use z15 on selective areas (i.e. have a list of bboxes that should be z15), or do it everywhere?

frodrigo:
@daliborjanak the oposit approach, all in real time, is also an doable option. Bifurcated may not be an interesting way.

Low zoom rendering are already fast, as the data was already generalized in the data base (not the same strategy used in carto-osm/mapnik default approach).

I am trying to run this all real-time strategy in this project, OMT based: https://github.com/makina-maps/makina-maps

ZeLonewolf:
[...]

PR #1235, opened 2021-09-18
Added power facilities as a new layer

### LINKED ISSUE #139: Power=* features
Power distribution networks are missing.
power=pole/tower as pois (tilezen has this)
power=cable/line as line (not in tilezen)
http://wiki.openstreetmap.org/wiki/Power_lines

ache051:
Fixes #139 

Although [not used in USGS topo series](https://www.usgs.gov/faqs/why-are-there-no-power-lines-pipelines-libraries-trails-etc-us-topo-maps), power facilities including polygons (power plants etc.), lines (power lines etc.) and points (power poles etc.) are used as a whole in topographic maps of other countries ([NZ Topo50](https://www.linz.govt.nz/sites/default/files/media/pages-attachments/topo50-legend.png), [Austria ÖK50](https://www.bev.gv.at/pls/portal/docs/PAGE/BEV_PORTAL_CONTENT_ALLGEMEIN/0200_PRODUKTE/SCHNITTSTELLENBESCHREIBUNGEN/ZEICHENSCHLUESSEL_FUER_DIE_OESTERREICHISCHE_KARTE_1_50_000.PDF)) or sometimes used in parts ([Swiss Topo series](https://www.swisstopo.admin.ch/content/swisstopo-internet/en/home/products/maps/national/lk25/_jcr_content/contentPar/tabs_copy_copy_copy/items/dokumente_publikatio/tabPar/downloadlist_copy_co/downloadItems/554_1466074737273.download/symbols_en.pdf)). The inclusion of power facilities has been queried by a few in #139 but has not yet been looked at deeply. This PR addressed that by creating a new power layer.

PR will also be made to a GL style within the next few days.

etl_diagram
mapping_diagram


ZeLonewolf:
In openstreetmap-carto ([example](https://www.openstreetmap.org/#map=14/41.6011/-71.4946)), power lines and power poles are not shown until zoom level r14/v13, however, this PR adds power features at zoom level r11/v10 which seems like an excessive level of detail [at that zoom level](https://www.openstreetmap.org/#map=11/41.5516/-71.4416) for a general-purpose map style.

ache051:
I have had some internal struggle with what zoom levels the power facilities should be visible, and also whether it should be controlled via GL style or to preset it when vector tiles are produced. In this particular version I opted for controlling via style for two reasons:
1) I have a need for them at lower zoom levels
2) Using Denmark from z1-z14 as an example. The resultant mbtiles with power layered z10+ (356.8MB) is only 0.2MB larger than without (356.6MB), so I thought it would be a good inclusion.
I am aware that we should be conscious of mbtiles size and rendering time, so I am always open to discussion!
Happy to provide the Denmark files for assessment as well if there is interest.

ZeLonewolf:
I think the concern would be tile size inflation at particular zooms (probably z10 would be the most problematic), vice looking at the overall mbtiles size.  Once a maintainer approves running the CI, we'll have some metrics to look at.

It would think that by the time you get down to z10, you would want to suppress things like individual poles, which would be clutter at that point.  Additionally, that's a low enough zoom (256:1 over z14) where you would also want to have generalized geometries for the power lines.

ache051:
@ZeLonewolf Agree that starting from z10 is probably too much for general purpose basemaps. Looking at the openstreetmap-carto example you provided, another concern is that some minor power facilities (minor_line, pole, etc.) don't render until z15+. Meanwhile it seems openmaptiles schema has no new features appear beyond z14 for any layer.
Thus we can either 1) Follow the openmpatiles schema, and make all power facilities big and small to render from z13+ or only at z14, or 2) Follow openstreetmap-carto and have additional high zoom level (z15+) for smaller features.
I am happy to try out both ways or if there is any other suggestions?

About option 2, last time I ran the whole planet (without power facilities) from z1-z14 the resultant mbtiles file was ~110GB, so generating to z15 will result in a ~450GB file. Storage-wise it is not too bad, but it might take a long long time to render. Hopefully aspects of it will be tackled by discussion in issue #1178 that you raised.

ZeLonewolf:
Until the project adopts z15 tile rendering, all features intended to render at z14 or higher are rendered at z14.  At some point if z15 is added, the project will need to decide which current z14 features shouldn't actually start until z15.  So for features that are intended for higher zooms, they can just be included in z14.

My point about generalized geometries still remains.  At lower zooms, where you would show lines but not towers/poles, the geometries need to be generalized into simpler shapes appropriate for each zoom level.  That's what the `_gen` tables and `ST_Simplify` calls that you see in other layers are doing.  If we are including these objects at lower zooms, the unneeded detail needs to be removed to constrain tile size.

ache051:
I am examining other layers on how to implement generalization and get the routine to use the generalized tables. Any hints?

ZeLonewolf:
The `park` layer is probably the simplest example to start from.

ache051:
I have done some work and would like to report back to have some discussion:

**Scheme** 

- Features with "cable" tag should not be in the scheme as most of them are underground cables.
- Larger power facilities should be rendered from Z13+ and smaller ones at Z14 to minimize file size and mbtiles generation time (refer to next section).

**Styling**

- Polygons (station and substation) should reside with landuse polygons in terms of layering. In fact we might as well put it into the landuse layer. They should appear at Z13+.
- Points (pole and tower) should be on top of roads section in terms of layering. Towers should appear at Z13+ and poles at Z15+.
- Lines (minor_line and line) should be on top of the points in terms of layering. "Lines" should appear at Z13+ and minor_lines at Z15+.
- As you can see I am proposing to use styling to make certain features appear only at Z15+, beyond OMT's scheme which stops at Z14 (for the time being).
- I also propose to render lines and points together (i.e. lines with tower, minor_lines with poles) to avoid confusion with other features.

**Generalization**

- I have investigated this a little, but is generalization still needed if power facilities are shown only at higher zoom levels?

**Sample**

- I have produced a short video of what it looks like with all the points above implemented into OSM Bright style. Apologies for the jumpy cursor as refresh rate was set too low.

https://user-images.githubusercontent.com/10702936/135052253-6a92381a-4b05-420a-b00b-d4a065013df0.mp4

I have not push the changes to the branch yet as I am more than happy to have discussion on this work before doing so!

frodrigo:
I am also interested by a power layer.

But, maybe this layer could be the first optional layer ?

ZeLonewolf:
> But, maybe this layer could be the first optional layer ?

I do not thing power lines and towers should be an optional layer, it should be on a core layer.  It is a [stated goal](https://github.com/openstreetmap/operations/issues/565#issuecomment-907303115) of the OpenMapTiles project to create:
> "An “OSM Carto” type of style - with cartographical goal to demonstrate all included data"

Since openstreetmap-carto renders power lines and power towers, we should copy that behavior, ideally matching zoom levels.

For zoom levels lower than what osm-carto renders, _that_ would be an appropriate choice for a first optional layer.

ache051:
I also think that at least some of the power facilities should not be optional, if having topographic-oriented basemap is one of the main goals. Not only that power lines and towers appear in most 1:50K, 1:25K topo maps around the world, they can also be used as landmarks during navigation.
Obviously you good people are far more experienced in OMT and OSM than I am, so please go ahead if there is any suggestions or opinions? Or should I go ahead to push the modifications mentioned in my previous post to the branch?

ZeLonewolf:
@ache051 please run `make generate-devdoc`, and commit the resulting changes, in order to update the ETL diagrams.

nmcdtu:
@ache051 It is really nice and useful! However, the file `style.json` is missing. 
Could you please consider sharing it as well?

Issue #1220, opened 2021-09-10
Raise a helpful error if Natural Earth data is not imported


ZeLonewolf:
During testing of #1213, it was noted that running `make import-sql` without having previously run `make import-data` causes a somewhat cryptic error message.  Instead, we can add a check to the top of the SQL to verify that the Natural Earth tables are present, and raise a more helpful error message if they're missing.

nyurik:
We could also extend the yaml files to say which tables or functions each layer expects to be present in the DB before it is imported.  This way we can auto-generate the same checks for all layers, and prepend them to the sql files.

ZeLonewolf:
Hi, does openmaptiles/openmaptiles-tools#370 actually fix this issue?  Or are more changes needed on the openmaptiles side to update various yaml files with the Natural Earth dependencies?

Falke-Design:
No this is not fixed. We provided only the funcunality now we need to add the required things in the layer.yaml

ZeLonewolf:
@TomPohys please re-open ticket.

Falke-Design:
@nyurik now we had added a check for the required tables / functions but it is still no helpful error message:
![grafik](https://user-images.githubusercontent.com/19800037/133892964-9888ce2c-00e6-4ba1-85de-018ed1033cf7.png)

It would be better if we show a message like `You need first to create the table with 'make import-data'`

nyurik:
Good point @Falke-Design, i guess we could add a generic error message at the top (simpler than to have a help text for each required table, and deal with duplicate messages, etc).

```yaml
requires:
  helpText: "This layer requires tables generated with 'make import-data' command."
  tables:
    - osm_ocean_polygon
    ...
```


I guess we could wrap the currently-generated code with something like this:

```sql

do $$
begin
  SELECT 'osm_ocean_polygon'::REGCLASS;
exception when others then   -- I think we should use a specific error instead of "others" here
  RAISE EXCEPTION SQLERRM USING HINT = 'This layer requires tables generated with 'make import-data' command.' ;
end;
$$ language 'plpgsql';
```

I think SQLERRM is what contains statement-specific error, and HINT will help with additional information.  I haven't tested the above.

PR #1236, opened 2021-09-18
Add requires.tables to <layer>.yaml

Falke-Design:
**NOTE** this can only be merged after the next tools version is released.

Added required Postgres tables to the `<layer>.yaml` definition.
Close: #1220

PR of tools: https://github.com/openmaptiles/openmaptiles-tools/pull/370


nyurik (commented):
i think it would be cleaner to use the yaml list rather than json list - makes it easier to read.

TomPohys:
Hi @Falke-Design, I agree that tables should be in the YAML list, the same as you did for the test in [OMT-T](https://github.com/openmaptiles/openmaptiles-tools/blob/74fcd480ecbfab63f90814c8864e7c7dbffb1965/tests/testlayers/mountain_peak/mountain_peak.yaml#L9)

Could you please update it?

Thank you

Falke-Design:
@TomPohys should the `"` removed too? Now I changed the single items to a list too, is this ok?

TomPohys:
Thank you, it looks better now. A single item as a YAML list looks OK, and it has consistency through all files.

If the `"` are not necessary, let`s remove them.
Thank you

Falke-Design:
@TomPohys done

TomPohys (approved):

TomPohys:
Thank you!

PR #1237, opened 2021-09-18
Print error of `import-sql` at the end of the output

Falke-Design:
Because of the parallel processing of `import-sql` the error message is printed in the middle of the output. With this PR the error is displayed again at the end:
grafik

Issue: https://github.com/openmaptiles/openmaptiles-tools/pull/370


nyurik (approved):
no idea how this works, but seems ok :D

Falke-Design:
```
awk '1{print; fflush()} $$0~".*ERROR" {txt=$$0} END{ if(txt){print "\n*** ERROR detected, aborting:"; print txt; exit(1)} }'
```
Explanation:
- `1{print; fflush()}` means if true then print output -> `1{}` same as `if(true){}`
- `$$0~".*ERROR" {txt=$$0}` get first argument `$$0` (line of output) and check if it contains (regex string) `ERROR`. If true save line to var `txt`: `{txt=$$0}`
- `END{ ... }` if last line of output is reached
- `if(txt){print "\n*** ERROR detected, aborting:"; print txt; exit(1)}` if error was found in a line / var `txt` is existing print it out

TomPohys (approved):

TomPohys:
Awesome! Great improvement!

PR #1239, opened 2021-09-18
BUGFIX: Route concurrency data is wiped after route member updates

ZeLonewolf:
This PR is a bugfix for the `transportation_name` layer.  Currently, the updates process does not work, as noted by @zstadler in #1230.  The issue is that the computed `concurrency_index` column in `osm_route_member` is never updated after the initial updates.  Therefore, whenever route member ways are updated, they will lose their route concurrency index, and hence route concurrencies will be lost in the `transportation_name` tiles.  This PR adds the missing concurrency index update code to the transportation/network update triggers.

This PR as written also incorporates the fixes in #1230 and #1233.

The following SQL is a unit test to demonstrate that these triggers work correctly:

```
select count(*) from osm_route_member where concurrency_index > 5;
select count(*) from osm_transportation_name_network where route_6 is not null;

update osm_route_member set concurrency_index=NULL where concurrency_index > 5;

select count(*) from osm_route_member where concurrency_index > 5;
select count(*) from osm_transportation_name_network where route_6 is not null;
```

If working correctly, both pairs of `count(*)` values should return the same pair of numbers.

I have completed code changes, but marking this as draft so it can be tested as I'm not 100% sure how to prove that the update process is working.


zstadler:
I suggest that updating of the `concurrency_index` would be executed only on the changes included in the update, and not for the entire `osm_route_member` table. This would maintain the performance improvements that were made by the incremental update.

[`update_osm_route_member()`](https://github.com/ZeLonewolf/openmaptiles/blob/3e26a8251e618fce8612565d8d25f00df944b7b7/layers/transportation/update_route_member.sql#L47-L71) is responsible for this update.
All the operations are limited to relations with `c.osm_id` included in `transportation_name.network_changes AS c`:

```
CREATE OR REPLACE FUNCTION update_osm_route_member() RETURNS void AS
$$
BEGIN
    DELETE
    FROM osm_route_member AS r
        USING
            transportation_name.network_changes AS c
    WHERE network IN ('omt-gb-motorway', 'omt-gb-trunk')
      AND r.osm_id = c.osm_id;


    INSERT INTO osm_route_member (osm_id, member, ref, network)
    SELECT r.*
    FROM gbr_route_members_view AS r
             JOIN transportation_name.network_changes AS c ON
        r.osm_id = c.osm_id;


    UPDATE
        osm_route_member AS r
    SET network_type = osm_route_member_network_type(network)
    FROM transportation_name.network_changes AS c
    WHERE network != ''
      AND network_type IS DISTINCT FROM osm_route_member_network_type(network)
      AND r.member = c.osm_id;
END;
$$ LANGUAGE plpgsql;
```

ZeLonewolf:
Thanks for the review @zstadler!

> I suggest that updating of the `concurrency_index` would be executed only on the changes included in the update, and not for the entire `osm_route_member` table. This would maintain the performance improvements that were made by the incremental update.

Restricting the concurrency update to actual changes was already done in the original  PR:
```
WHERE rm.member IN
      (SELECT DISTINCT osm_id FROM transportation_name.network_changes)
```

Any change to route members might change the concurrency index ordering.  Therefore, it's necessary to re-calculate the concurrency index for all relation memberships of the changed members.  Further, the `DENSE_RANK()` function that computes this requires a partition to assign rank numbers, so this is not possible to do on a row-by-row basis, which is the reason for the sub-select approach.

However, with your suggestion to move the concurrency index calculation into the `osm_route_member` trigger, I was able to merge that with the existing `osm_route_member` update code that deals with network type, so that should improve performance.

I've updated the PR with this change.

zstadler:
Thanks @ZeLonewolf for your comment.

I seem to have missed the fact that the recalculation is restricted to network members (ways) that were affected by the change, 
as captured in `transportation_name.network_changes`:

```
    WHERE rm.member IN
      (SELECT DISTINCT osm_id FROM transportation_name.network_changes)
```

and that it includes both additions and removals of members from relations:
```
    INSERT INTO transportation_name.network_changes(osm_id)
    VALUES (CASE WHEN tg_op IN ('DELETE', 'UPDATE') THEN old.member ELSE new.member END)
```

TomPohys (approved):

TomPohys:
Nice work! Thank you both!

PR #1241, opened 2021-09-19
Suppress rendering of underground buildings

### LINKED ISSUE #1227: Avoid including underground buildings
It would be better to avoid adding buildings with `layer<0`, `level<0` or `location=underground` to the tiles
Currently, they all show as above-ground structures.

[This undergound structure](https://www.openstreetmap.org/way/480727271), for example, [is shown](https://www.maptiler.com/maps/#streets//vector/19/9.8325249/53.4721039/-135.20/59.5) as a regular building:

image

ZeLonewolf:
Closes #1227

This PR does the following:
1. Suppresses underground buildings from rendering, when such buildings are tagged `location=underground`.  Based on the wiki documentation, the `layer` tag does not determine whether a feature is above of below ground.
2. Unifies the building and building relation imposm exclusion mappings.

The following underground building feature in Luxembourg is used as the test object:
https://www.openstreetmap.org/way/582873794

**Before:**
image

**After:**
image

### DIFF
--- a/layers/building/mapping.yaml
+++ b/layers/building/mapping.yaml
@@ -65,6 +65,7 @@ tables:
         building: ["no","none","No"]
         building:part: ["no","none","No"]
         man_made: ["bridge"]
+        location: ["underground"]
     type: polygon
 
   # etldoc: imposm3 -> osm_building_relation
--- a/layers/building/mapping.yaml
+++ b/layers/building/mapping.yaml
@@ -60,6 +60,8 @@ tables:
       aeroway:
       - terminal
       - hangar
+      location:
+      - underground
     filters:
       reject:
         building: ["no","none","No"]
@@ -158,4 +160,12 @@ tables:
       type: member_type
     mapping:
       type: [building]
+      location:
+      - underground
+    filters:
+      reject:
+        building: ["no","none","No"]
+        building:part: ["no","none","No"]
+        man_made: ["bridge"]
+        location: ["underground"]
     type: relation_member


layers/building/mapping.yaml
```diff
@@ -60,11 +60,14 @@ tables:
       aeroway:
       - terminal
       - hangar
+      location:
+      - underground
```
zstadler, line 64: There is no need to add `location` here. It is not needed for the `reject` filters below, much like the filter on the `man_made` tag.

Similarly, for building relation table below.

zstadler:
Currently, this PR only partially addresses #1227. It maintains the inclusion of buildings with `level<0` or `layer<0` tags in the `building` layer.

Based on the [wiki](https://wiki.openstreetmap.org/wiki/Key:level), `level<0` should also be use to identify underground building structures.

About 10% of the `layer` tags are placed on buildings. 

There are [62516 buildings  (click "Run" and be patient)](https://overpass-turbo.eu/s/1bkm) that have `layer=-1` and no `location` or `level` tags.

image

There are [2092 buildings](https://overpass-turbo.eu/s/1bkp) that have `location=underground` and no `layer` or `level` tags.

image

There are [1000 buildings](https://overpass-turbo.eu/s/1bkG) that have `level=-[1-6]` and no `layer` or `underground` tags.

image

In practice mappers probably choose negative `layer` values for underground structures because:
1. `layer<0` is well associated with other underground elements, such as tunnels.
2. An element with `layer<0` will not be rendered when overlapping with another element without a layer tag, as the [wiki warns](https://wiki.openstreetmap.org/wiki/Key:layer#Things_to_avoid). Other tags do not necessarily achieve this effect.

ZeLonewolf:
With regard to the `layer` tag, the [wiki entry for Key:layer](https://wiki.openstreetmap.org/wiki/Key:layer) states:

> Negative values do not imply that object is underground, use `location=underground` for this purpose.

That was the reason I limited this PR to the underground tag, which should be an uncontroversial indicator of undergroundness.

The use of `level=*` + `building=*` seems like a tagging mistake based on a reading of the wiki - level is supposed to be used for features internal to a building.  I'm not sure we should encourage the use of that tag on buildings, and the very small count of 1000 compared to other combinations and buildings as a whole seems small enough to ignore.

layers/building/mapping.yaml
ZeLonewolf: Thanks, I misunderstood what a building relation was and I agree those filters aren't necessary.

I found that I needed to add location=underground to the mapping (but not to the _columns_, or the filter wouldn't work.  I could be wrong about that and I'd be happy to re-test this to confirm it.

zstadler:
@ZeLonewolf,

In Theory There Is No Difference Between Theory and Practice, While In Practice There Is.

IMO, merging this PR, which handles about 3% of underground buildings, is not worthwhile.

ZeLonewolf:
@zstadler could you share some analysis that indicates that `building=*` + `layer=-1` consistently indicates an underground building?  I think you are making an assumption here that `layer<0` is used in place of `location=underground`, but I have not found this to be the case.

I examined the data in my state, and this is what I found.  There were 9 buildings that were marked with `layer=-1`.  Here's the overpass query I used:

```
[out:json][timeout:2500];
{{geocodeArea:"Rhode Island"}}->.a;
nwr[building][layer=-1](area.a);
out body;
>;
out skel qt;
```

Of the 9 objects returned, six of them were buildings that were definitely above ground:

https://www.openstreetmap.org/way/233329544
https://www.openstreetmap.org/way/230776599
https://www.openstreetmap.org/relation/4731990
https://www.openstreetmap.org/way/489667338
https://www.openstreetmap.org/way/887531032
https://www.openstreetmap.org/way/896370103

Two of them were parking garages with underground levels, but above-ground expression (i.e. they're definitely visible on satellite, appear as buildings in the real work, and thus should be rendered:

https://www.openstreetmap.org/way/764664384
https://www.openstreetmap.org/way/760594431

One of them was a parking garage under a building, and appears to be legitimately underground (and even has the tag `parking=underground`):
https://www.openstreetmap.org/way/765290068

So in my state, only 11% (one object) was actually below ground.  This is a pretty poor success rate.  That is why I suggested limiting this to `location=underground` which is unambiguous in declaring that the object is underground.

layers/building/mapping.yaml
ZeLonewolf: I've confirmed from re-testing that it is required to have a mapping in order to apply a reject filter.  Otherwise, the filter does nothing.

zstadler:
I see your point. Mappers were using # the `layer` tag to handle overlaps of buildings with roads and other buildings.
Apparently, in Practice the abuse of `layer<0` is far more diverse than I expected.

ZeLonewolf:
The other tag we might consider is `parking=underground`, however, I'm not confident on how it's used, or whether it even needs to be excluded.

PR #1243, opened 2021-09-23
SQL migration for new version of imposm

ZeLonewolf:
Fixes openmaptiles/openmaptiles-tools#363

This PR migrates the SQL in the `transportation` and `transportation_name` layer to use the new imposm3 mappings which now map a separate primary key ID.

Additionally, devdocs are updated so that CI will pass :)

**EDIT**: I can't explain why CI doesn't pass :(  It works locally.


nyurik:
@ZeLonewolf see the RAW log, and search for "error" from the bottom (shift+F3).  I see this relevant tidbit:

```
2021-09-23T02:31:01.2730673Z psql:/sql/parallel/mountain_peak.sql:279: NOTICE:  Finished layer mountain_peak
2021-09-23T02:31:01.4012870Z psql:/sql/parallel/water__waterway.sql:1: NOTICE:  Processing layer water
2021-09-23T02:31:04.9084447Z psql:/sql/parallel/transportation__transportation_name.sql:260: NOTICE:  relation "osm_highway_linestring_gen_z11_osm_id_idx" already exists, skipping
2021-09-23T02:31:05.1050842Z psql:/sql/parallel/transportation__transportation_name.sql:270: ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification
2021-09-23T02:31:05.1060627Z xargs: sh: exited with status 255; aborting
2021-09-23T02:31:40.6667583Z psql:/sql/parallel/housenumber.sql:123: NOTICE:  Finished layer housenumber
2021-09-23T02:31:48.4717356Z psql:/sql/parallel/poi.sql:247: NOTICE:  trigger "trigger_flag" for relation "osm_poi_point" does not exist, skipping
2021-09-23T02:31:48.4719975Z psql:/sql/parallel/poi.sql:248: NOTICE:  schema "poi_point" does not exist, skipping
2021-09-23T02:32:35.2169592Z psql:/sql/parallel/water__waterway.sql:37: NOTICE:  table "osm_ocean_polygon_gen_z11" does not exist, skipping
```

ZeLonewolf:
@nyurik I see the error, but it's indicating that I'm not hitting the "new" openmaptiles-tools.  Is there something I have to change in the github workflow to do that?

nyurik:
I think i know the issue -- it crashes in the performance test, which tests two versions -- the one in master-tools and the one in the PR. I suspect it crashes when it tries to get performance of the main branch and fails because master-tools doesn't work.  Just an idea, would have to dig deeper but it's past midnight and i'm a zombie

ZeLonewolf:
That would certainly explain the issue.  The SQL can't be compatible with both versions without a bunch of SQL gymnastics.

nyurik:
Actually this should work if you create a pr against the master branch,  and include the env change to version latest

ZeLonewolf:
Replaced by #1246

PR #1245, opened 2021-09-23
Integrity check for DIFF mode

### LINKED ISSUE #1226: Integrity tests to check the import-diff flow

zstadler:
This test applies 2-3 months of weekly updates on the `europe/monaco` extract from [Geofabrik](http://download.geofabrik.de/europe/).

It is worth noting that the contents of the updates may vary, and some SQL update flows may not be tested if the relevant changes did not occur during that period.

Resolves #1226


ZeLonewolf:
I was thinking about your note about update variability, and I'm wondering if it makes sense to start creating a library of fake sample data for the purpose of creating static unit tests.

zstadler:
That's not an easy task...

ZeLonewolf:
Rome wasn't built in a day :)

What I'm protesting wouldn't be a replacement for this PR, it would be in addition to it.

.github/workflows/integrity.yml
```diff
@@ -36,3 +36,36 @@ jobs:
...
+          docker-compose run --rm --user=$(id -u):$(id -g) openmaptiles-tools sh -c "
+            while ! osmupdate --keep-tempfiles --base-url=$(sed -n 's/ *\"replication_url\": //p' data/$area.repl.json) data/$area.osm.pbf data/changes.osc.gz ; do 
+              sleep 2;
+              echo Sleeping...;
+              sleep 630;
+            done"
+          echo Downloading updates completed
+          echo Importing updates
+          make import-diff
+          echo Generaing new tiles
```
ZeLonewolf, line 70: Typo, should be "Generating"

zstadler: Thanks @ZeLonewolf.
 Corrected.

zstadler:
Yes. When someone creates the first static unit test, it would be a separate PR.

TomPohys:
This is a great improvement in CI testing. Thank you very much!

TomPohys (approved):

PR #1246, opened 2021-09-24
Upgrade omt-tools and migrate SQL

ZeLonewolf:
Fixes openmaptiles/openmaptiles-tools#363

This PR migrates the SQL in the transportation and transportation_name layer to use the new imposm3 mappings which now map a separate primary key ID.

### DIFF
--- a/layers/transportation/update_route_member.sql
+++ b/layers/transportation/update_route_member.sql
@@ -60,15 +60,16 @@ BEGIN
              JOIN transportation_name.network_changes AS c ON
         r.osm_id = c.osm_id;
 
-    INSERT INTO osm_route_member (id, network_type, concurrency_index)
+    INSERT INTO osm_route_member (id, osm_id, network_type, concurrency_index)
     SELECT
       id,
+      osm_id,
       osm_route_member_network_type(network) AS network_type,
       DENSE_RANK() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index
     FROM osm_route_member rm
     WHERE rm.member IN
       (SELECT DISTINCT osm_id FROM transportation_name.network_changes)
-    ON CONFLICT (id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index;
+    ON CONFLICT (id, osm_id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index;
 
 END;
 $$ LANGUAGE plpgsql;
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
I recommend that #1245 be merged before this PR so that we can have a unit test for the update process before touch update code.

TomPohys:
Thank you for working on this! Could you please just not merging into the `master` branch but into `master-tools` branch? Thank you.

ZeLonewolf:
@TomPohys I attempted this in #1243, however master-tools will not pass CI as long as the master branch still has the older imposm.  So I'm not sure how to get around that but this PR at least holds the SQL fix ready to go for when omt-tools gets upated I guess?

See comments by @nyurik: https://github.com/openmaptiles/openmaptiles/pull/1243#issuecomment-926027617

Perhaps it might be possible to do a micro-release of omt-tools that upgrades imposm and does nothing else?

TomPohys:
Oh, sorry I missed the #1243. Let`s wait for https://github.com/openmaptiles/openmaptiles-tools/pull/378 and then we will release new tools. Is it OK for you?

nyurik:
@TomPohys I will postpone run-psql until 6.1 - i am not too happy with the ctrl+c and error handling in it. I have released 6.0, so we can proceed with this merge.

ZeLonewolf:
Since we're seeing the increase as low as z1, we can probably rule out the constraint changes in the transportation table, since transportation objects aren't rendered at that low of a zoom.

ZeLonewolf:
Great analysis @nyurik!  Because this PR also updates the SQL, I'm concerned about inadvertently breaking something on the SQL side with the transition to dual key.  Ideally we could upgrade the postgres version without upgrading the imposm version as a separate step in order to re-baseline the tile size.  But, maybe I'm overcomplicating things.

nyurik:
Digging deeper: running `psql` in both above shells:  The POI `Holiday Inn London Regents Park` with `osm_id = 1082519651` is actually fake -- the SQL changes original with `osm_id * 10 + 1`. Deleting last digit is found in v6 result, but it does NOT exist in the v5.3 one with this query: `select * from osm_poi_polygon where osm_id = 108251965;`.  Same result with `select * from osm_poi_polygon where osm_id = 3241840;`.

Checking the `building` layer has similar result - v6 has that data but v5.3 does not (unlike POI, the original osm_id is used):  `select * from osm_all_buildings where osm_id = 33632838;`

PR #1248, opened 2021-09-27
BUGFIX: Park polygon update fails for z4

ZeLonewolf:
This PR fixes a bug in the park polygon update code introduced in #1160.  Because park polygons at zoom 4 are dissolved, the centroid is not present in the table at that layer.  Thus, a separate update trigger is needed to avoid errors.  This PR adds the separate trigger for the zoom 4 park layer.

This PR likely resolves an unidentified blocker for #1245


TomPohys (approved):

TomPohys:
Thank you, good catch!

PR #1252, opened 2021-09-29
YAML-only user-defined fields

zstadler:
Allow developers of forked repositories to add fields by modifying the YAML files of the transportation layer.

Benefits:
- Avoid the need to modify SQL files
- Reduce the effort of merging upstream changes

The following example is also embedded as comments in the files.

Add the required OSM tags in `mapping.yaml`:
```
tags:
  include:
    - maxspeed
    - maxheight
    - tracktype
```

Extend `fields`  in `transportation.yaml`:
```
    maxspeed:
      description: |
          Original value of [`maxspeed`](http://wiki.openstreetmap.org/wiki/Key:maxspeed) tag, if exists.
    maxheight:
      description: |
          Original value of [`maxheight`](http://wiki.openstreetmap.org/wiki/Key:maxheight) tag, if exists.
    tracktype:
      description: |
          Original value of [`tracktype`](http://wiki.openstreetmap.org/wiki/Key:tracktype) tag, if exists.
```

Extend `query`  in `transportation.yaml`:
```
    query: |
      (SELECT geometry,
      -- Optional user-defind tags, such as
      NULLIF(tags->'maxspeed', '') AS "maxspeed",
      NULLIF(tags->'maxheight', '') AS "maxheight",
      NULLIF(tags->'tracktype', '') AS "tracktype",
      --
      class, subclass, network, oneway, ramp, brunnel, service, access, toll, layer, level, indoor, bicycle, foot, horse, mtb_scale, surface FROM layer_transportation(!bbox!, z(!scale_denominator!))) AS t
```


frodrigo:
But what about the complementary database size required to store tags?

zstadler:
> But what about the complementary database size required to store tags?

If I understand the [results of the performance tests](https://github.com/openmaptiles/openmaptiles/pull/1252#issuecomment-929958490) above, the PG DB is 0.4% larger and the tiles file is 0.2% larger.

layers/transportation/mapping.yaml
```diff
@@ -360,9 +368,7 @@ tables:
     - *layer
     - *level
     - *indoor
-    - name: is_area
-      key: area
-      type: bool
+    - *area
```
ZeLonewolf, line 371: This change doesn't appear related - can it be a separate PR?

ZeLonewolf:
I think this PR is moving in the right direction, but there is a significant tile size increase at some zooms, the worst being +25% on z7 tiles.  The "overall" tile size increase in small because there's no size increase in zoom 12-14, which represent the majority of the tiles by count (for every zoom 11 tile there's 4 z12 tiles, 16 z13 tiles, and 64 z14 tiles and so forth).

I have not actually run/debugged this PR but I'm pretty sure the issue is that there are unnecessary tags being included in the tags hstore.  The [imposm documentation](https://imposm.org/docs/imposm3/latest/mapping.html) indicates that an optional `include` parameter can explicitly list the fields that are mapping into the tags hstore.  This was done in the commented-out example mapping, but not in the "default" mapping in this PR.  Therefore, the tags likely contain a copy of all mapped tags, including ones that are already mapped to other columns and are thus redundant.

nyurik:
Just an idea -- would it be possible to have this override configurable in the [openmaptiles.yaml](https://github.com/openmaptiles/openmaptiles/blob/master/openmaptiles.yaml) file instead of modifying the "production" layer files?  One of the biggest pains with overrides is to merge upstream changes.  If something like this can be made generic enough to be usable declaratively, forks don't have to deal with merging upstream into production code, and layers can safely take such extensions into account, e.g. by declaring placeholders, e.g. `%%EXTENSION_FIELDS%%` in SQL.  Plus tools can verify that such extension is even possible (i.e. if placeholder doesn't exist, it will raise an error rather than silently ignore yaml).  Some ideas:

```yaml
tileset:
  layers:
    # ...
    - layers/aeroway/aeroway.yaml

    - layer: layers/transportation/transportation.yaml
      extend:
        fields:
          maxspeed:
            description: Original value of [`maxspeed`](http://wiki.openstreetmap.org/wiki/Key:maxspeed) tag, if exists.
          maxheight:
            description: Original value of [`maxheight`](http://wiki.openstreetmap.org/wiki/Key:maxheight) tag, if exists.
          tracktype:
            description: Original value of [`tracktype`](http://wiki.openstreetmap.org/wiki/Key:tracktype) tag, if exists.
        tags:
          - maxspeed
          - maxheight
          - tracktype
        query:
          - NULLIF(tags->'maxspeed', '') AS "maxspeed",
          - NULLIF(tags->'maxheight', '') AS "maxheight",
          - NULLIF(tags->'tracktype', '') AS "tracktype",

    - layers/building/building.yaml
    - layers/water_name/water_name.yaml
  # ...
```

ZeLonewolf:
The ```NULLIF(tags->'maxspeed', '') AS "maxspeed"``` seems superflous.  I was thinking we could have an ```{extended_fields}``` macro that gets put into the query SQL just like the ```{name}``` field.  Then you could just list the tags once under "extend" and have everything else auto-populated with openmaptiles-tools.

nyurik:
> The `NULLIF(tags->'maxspeed', '') AS "maxspeed"` seems superflous. I was thinking we could have an `{extended_fields}` macro that gets put into the query SQL just like the `{name}` field. Then you could just list the tags once under "extend" and have everything else auto-populated with openmaptiles-tools.

Sure thing - I simply copy-pasted the original to explain the idea, but the actual YAML structure / extension mechanism should be polished further.  Just need to make it generic enough to cover the most common cases.

layers/transportation/mapping.yaml
zstadler: Please feel free to do so.

zstadler:
Thanks @nyurik and @ZeLonewolf for the interface improvement suggestions. I like them both.

From an implementation viewpoint:
- The extension syntax requires modifications in `generate-tm2source`, `generate-imposm3`, and `generate-sql`.
- The underline SQL implementation in each layer would be based on the principles of this PR.
- Introducing this configuration in `openmaptiles.yaml` requires that the SQL of all existing layers will be extended.
- Special care need to be taken for handling the `transportation_name` and `waterway` layers that require the `transportation ` and `water` layers, resp.

zstadler:
> I think this PR is moving in the right direction, but there is a significant tile size increase at some zooms, the worst being +25% on z7 tiles. The "overall" tile size increase in small because there's no size increase in zoom 12-14, which represent the majority of the tiles by count (for every zoom 11 tile there's 4 z12 tiles, 16 z13 tiles, and 64 z14 tiles and so forth).

This is probably a bug in the modified SQL code. Since the PR did not introduce any new fields, it is not expected to have any impact on the resulting tiles. However, I'm clueless as to the source of this increase. Any help would be greatly appreciated!

layers/transportation/update_transportation_merge.sql
```diff
@@ -111,7 +115,7 @@ WHERE (highway IN ('motorway', 'trunk', 'primary') OR
        construction IN ('motorway', 'trunk', 'primary'))
        AND ST_IsValid(geometry)
        AND access IS NULL
-GROUP BY highway, network, construction, is_bridge, is_tunnel, is_ford
+GROUP BY tags, highway, network, construction, is_bridge, is_tunnel, is_ford
```
nyurik, line 118: is this even possible to group-by an hstore field?  What does this mean??

ZeLonewolf: Yes, it is 100% possible to group by an hstore.  It will group on all key/value pairs being identical, which is why it's critical that you have a very strong handle on which tags are put into the hstore.

nyurik: I wonder how fast this is, and if it would make sense to put it last vs first in the list (again, not certain of performance implications)

nyurik:
Lets continue this conversation in https://github.com/openmaptiles/openmaptiles-tools/issues/377 where we can define the exact details of layer modification

zstadler:
> I think this PR is moving in the right direction, but there is a significant tile size increase at some zooms, the worst being +25% on z7 tiles. The "overall" tile size increase in small because there's no size increase in zoom 12-14, which represent the majority of the tiles by count (for every zoom 11 tile there's 4 z12 tiles, 16 z13 tiles, and 64 z14 tiles and so forth).

1. The increase in size is a result of the grouping by `tags` where `tags` contained much more info than I assumed.
2. I would not worry too much on the size increase of z7 tiles. For every tile in zoom level 7 there are 64K tiles in zoom levels 8-14...

nyurik:
tile size is important at every zoom because that's what determines how fast users see the map. We shouldn't put anything unneeded in a tile if it won't be used, no matter which zoom its at.

zstadler:
I agree. Yet the size increase is not a result of additional data, but rather a result of additional fragmentation of the same data.

ZeLonewolf:
> I agree. Yet the size increase is not a result of additional data, but rather a result of additional fragmentation of the same data.

That is a huge problem.  Fragmentation prevents generalization and low-zoom display logic from happening.  For example, I posted a fix #1141 which was causing highway shields to not appear due to inappropriate fragmentation.  Instead of a long stretch of road that can be generalized at lower zooms, you end up with a pile of small road bits that can't be generalized out.  This is fine if there's a rendering purpose to the data in the tags - i.e. if you _want_ the road broken up into bits whenever `maxspeed` or whatever tag changes so that you can render them differently.

If we're increasing fragmentation, this is _not_ a no-change PR.

zstadler:
What's creating the additional fragmentation is the fact that other column's data is replicated in the `tags` field. From reading the SQL code of OMT for several years now, this seems to have been the assumption all along.
I'm experimenting with a fix for this problem.

layers/transportation/update_transportation_merge.sql
```diff
@@ -29,16 +30,37 @@ SELECT (ST_Dump(ST_LineMerge(ST_Collect(geometry)))).geom AS geometry,
...
+       toll,
        layer
-FROM osm_highway_linestring_gen_z11
+FROM (
+        SELECT
+               geometry,
+               NULLIF((tags - slice_language_tags(tags)) - ARRAY[
+                    'highway', 'network', 'construction', 'bridge', 'tunnel', 'bicycle', 'foot', 'horse', 'mtb:scale', 'access', 'toll', 'layer', 
+                    'short_name', 'ramp', 'ford', 'oneway', 'area', 'service', 'usage', 'public_transport', 'ref', 'level', 'indoor', 'man_made', 'surface'
+                    ], ''::hstore) AS tags,
```
ZeLonewolf, line 42: 👏

zstadler:
This discussion has been stalled for a while.
Closing

PR #1253, opened 2021-10-01
[EXPERIMENTAL] Convert park layer to tag-based approach

ZeLonewolf:
This is an experimental PR intended to support ongoing discussions about how to add customizability to OMT layers.  As an experiment, this PR only modifies the `park` layer.

This PR moves tags not involved in `WHERE` clauses into the `tags` hstore object.  The advantage of this approach is that it eliminates the duplication of tag data in both the table columns as well as in the `tags` hstore.

The disadvantage of this approach is that the new tag `include` keys will apply across all layers rather than just this one.  If we wish to use this approach, it should be replicated across all layers in order to achieve the space-saving benefits.

(as an aside, this PR also fixes the fact that `name_en` and `name_de` weren't correctly hooked up, however, these fields will likely be removed in a future version so it's probably not even worth extracting to a separate PR to fix.)

### DIFF
--- a/layers/park/mapping.yaml
+++ b/layers/park/mapping.yaml
@@ -1,3 +1,9 @@
+tags:
+  include:
+  - landuse
+  - leisure
+  - boundary
+
 generalized_tables:
   # etldoc: osm_park_polygon_gen_z5 -> osm_park_polygon_gen_z4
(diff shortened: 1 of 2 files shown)


ZeLonewolf:
This is not quite the right approach.

PR #1259, opened 2021-10-08
Remove mention of now-gone `make import-borders`

flother:
After commit b4b897999d69 the `make import-borders` rule no longer exists, but it's still mentioned in the README a couple of times. This PR removes those references.

Some [cleanup files are also mentioned in `.env`](https://github.com/openmaptiles/openmaptiles/blob/4d6945b935ac5c7cbfa922fa0414d732f69e00e1/.env#L42-L44). I'm not sure if those lines can be removed too, because the variables are used in [`bin/import-borders`](https://github.com/openmaptiles/openmaptiles-tools/blob/3796c4eab03dd8a317cea2d2c9b0c7dbdce35ac9/bin/import-borders#L8-L10) in the `openmaptiles/openmaptiles-tools` repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -287,6 +287,7 @@ destroy-db:
 	$(DOCKER_COMPOSE) rm -fv
 	docker volume ls -q -f "name=^$(DC_PROJECT)_" | $(XARGS) docker volume rm
 	rm -rf cache
+	mkdir cache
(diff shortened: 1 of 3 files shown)


TomPohys:
Thanks!

PR #1265, opened 2021-10-14
Migrate transportation layer to hstore for extra tags

ZeLonewolf:
This PR collapses several fields in the `transportation` layer which are currently exposed in table columns into a `tags` hstore object.  This is preliminary work intended to support the ability to add arbitrary extra fields to layers without having to modify a long sequence of cascading tables, as discussed in openmaptiles/openmaptiles-tools#377 and #1252.  The intent of this PR is to reduce layer complexity for developers and increase flexibility of OMT users to add fields to layers without impacting performance.

The following approach is taken:

1. An in-place edit of the `tags` column is executed on the `osm_transportation_linestring` table in order to reduce the keys in the `tags` hstore to only those which are actually used.  For example, the `highway` tag is excluded because it's mapped to a column.  This happens in the new function `highway_linestring_strip_tags()`.
2. On update of a row in `osm_transportation_linestring`, the `tags` will be re-sliced.
3. Instead of column-mapped tags, the `tags` hstore is passed through the generalized tables and used as the final `GROUP BY` field.  This will ensure that adjacent geometries with matching tags will be grouped together.  The slicing operation in step 1 ensures that extra tags do not disrupt grouping.
4. Presentation logic is pushed to the top of the layer function, where tag values are extracted and presented to the tiles.
5. In `update_transportation_merge.sql`, the `tags` hstore is narrowed via subtract operators during the sequence of cascading tables.  The preserves the logic which determines which extra tags are presented at which zoom level.

Future efforts to further generalize this approach may consider:
1. Automated approaches to modify `update_tags.sql` which modify the list of inserted tags, at compile time.
2. Automated approaches to modify `update_transportation_merge.sql` by adding maxzoom capabilities to added extra tags.  That way, an end user can say "I want to add this tag to the tiles, but only up to zoom X".


ZeLonewolf:
@nyurik - a cursory check seems to indicate that value retrievals are a trivial operations (using \timing in the psql client):

image

The far greater concern in my mind is (a) actually getting this to work (as you can see from the CI and tile size comparisons, it's still wrong) and (b) whether there are any performance problems from the initial cull of the imposm-generated hstore or from grouping by the hstore.  However, so far I've been very impressed with postgres's ability to optimize seemingly complicated SQL.

ZeLonewolf:
Here are some initial performance metrics comparing this PR to the current master.  However, I caution that since this PR is incomplete and shows significant tile size differences, these may not be comparable.  But at least it gives us a sense that performance should be in the ballpark.  The test area is **new-york** which has a significant road network, generated down to zoom 14 on a bare-metal ubuntu box.  Numbers generated with `time ./quickstart.sh new-york`.

**master** branch:
```
real    10m41.512s
user    0m13.437s
sys     0m2.417s

real    10m35.676s
user    0m13.383s
sys     0m2.179s
```

This PR, with tags packed into an hstore:
```
real    10m32.645s
user    0m13.495s
sys     0m2.385s

real    10m32.156s
user    0m13.574s
sys     0m2.232s
```

PR #1266, opened 2021-10-15
Remove unnecessary ramp tagging

ZeLonewolf:
Unblocks #1265 

This PR removes unnecessary `ramp` tagging from the `transportation` layer.  The `ramp` tag is used for two situations.  First, it is set for all `highway=***_link` tags to indicate a highway ramp.  The second situation is when the `ramp` tag is set on a highway object, which indicates the presence of a ramp, such as one that would be used for a wheelchair.  `ramp=0` is a reasonable default in these situations; by removing cases of `ramp=0` and presenting only `ramp=1` to the tiles, we can save considerable size in the tiles.  Additionally, there appears to be a bug in which all objects tagged `highway=steps` are inexplicably tagged `ramp=1`.

The changes as follows:

1. Remove `ramp=0` where it appears in the tile.
2. `highway=steps` no longer assumes `ramp=1` in cases where `ramp=*` is not explicitly set.

In the current behavior, `ramp=0` is tagged only in at z13+ but suppressed in lower zooms:
image
image
image

This PR unifies `ramp=0` suppression between z13+ and z12-.


zstadler:
@ZeLonewolf,
Could you check if using `NULLIF(..., 0)` will further reduce the tile size?

ZeLonewolf:
@zstadler on further inspection, all of the `ramp=0` tagging appears to be spurious (only appears at z13+) so I've modified the `CASE` statement to only have the possibility of returning 1 or NULL.

PR #1268, opened 2021-10-16
Remove empty name_en and name_de values from waterway

ZeLonewolf:
This PR fixes a bug in the `waterway` layer in which empty `name_en` and `name_de` tags were being added to the tiles.


TomPohys:
Thank you @ZeLonewolf for the summary. I think we are both on the same page with option 1. The empty tags should not be in the tiles.

TomPohys (approved):

PR #1282, opened 2021-10-29
Update QUICKSTART.md

nnhubbard:
Updated the output of `make help`.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -415,33 +415,59 @@ the current output:
 
 ```
 ==============================================================================
- OpenMapTiles  https://github.com/openmaptiles/openmaptiles
+OpenMapTiles  https://github.com/openmaptiles/openmaptiles
+
 Hints for testing areas
-  make download-geofabrik-list         # list actual geofabrik OSM extracts for download -> <<your-area>>
+  make list-geofabrik                  # list actual geofabrik OSM extracts for download -> <<your-area>>
   ./quickstart.sh <<your-area>>        # example:  ./quickstart.sh madagascar
 
 Hints for designers:
-  make start-postserve                 # start Postserver + Maputnik Editor [ see localhost:8088 ]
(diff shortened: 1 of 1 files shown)


nyurik (approved):
thanks!

QUICKSTART.md
```diff
@@ -415,33 +415,59 @@ the current output:
...
+  make bash                            # start openmaptiles-tools /bin/bash terminal
+  make generate-bbox-file              # compute bounding box of a data file and store it in a file
+  make generate-devdoc                 # generate devdoc including graphs for all layers [./layers/...]
+  make generate-qa                     # statistics for a given layer's field
+  make generate-tiles-pg               # generate vector tiles based on .env settings using PostGIS ST_MVT()
+  make generate-tiles                  # generate vector tiles based on .env settings using Mapnik (obsolete)
+  make test-sql                        # run unit tests on the OpenMapTiles SQL schema
+  cat  .env                            # list PG database and MIN_ZOOM and MAX_ZOOM information
+  cat  quickstart.log                  # transcript of the last ./quickstart.sh run
+  make help                            # help about available commands
```
ZeLonewolf, line 441: You know... if they're looking at this, they've already figured out what `make help` does! 😂

nyurik:
I think we should rework the whole system a bit to make each doc string be attached to each target. Then we won't have the problem of targets being added/deleted without being documented.

```Makefile
.PHONY: foo
help:  ## Sample help string for the foo command
	@echo "foo"

.PHONY: help
help:  ## Show this help message
	@awk 'BEGIN {FS = ":.*##"; printf "Usage: make \033[36m<target>\033[0m\n\nTargets:\n"} /^[a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-10s\033[0m %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
```

This magic command will find all make targets, and attached help strings, and print them as a nice colored table.  Try copy/pasting this snippet into a `Makefile` and run `make help`

PR #1038, opened 2020-11-09
Brunnel in zoom level 8 and smaller

eva-j:
Following https://github.com/openmaptiles/openmaptiles/pull/1027, solving https://github.com/openmaptiles/openmaptiles/issues/999.

Adding brunnel attributes to merge tables of `transportation` layer, now it is possible to style/filter tunnels, bridges and fords in all zoom levels. Tested on Switzerland and Europe. Changes in generation speed and filesize minimal.

PR #1283, opened 2021-10-29
Remove bridge/tunnel attributes from tiny bridges

ZeLonewolf:
**EDIT**

This PR now removes bridge/tunnel/ford attributes from the `transportation` layer from sections of bridge/tunnel etc that are too small to be meaningfully visible at a particular zoom level.  This allows those segments to be generalized into geometries along with the surrounding sections of road, thereby dropping lots of useless little pieces of geometry in lower-zoom tiles.

**ORIGINAL**

This PR removes bridge/tunnel/ford attributes from the `transportation` layer for zooms 4-8.  At these zooms, bridges and tunnels are so small that it doesn't make sense to create differentiated rendering.  For comparison, openstreetmap-carto does not begin rendering bridges and tunnels until zoom v11/r12, so this change still preserves bridge and tunnel attributes two zooms lower than openstreetmap-carto.

For reference, here is a zoom 9 map of openstreetmap-americana:
image

Here is an equivalent zoom 8(raster) map of openstreetmap-carto:
image

The images above represent the lowest zoom level that bridge/tunnel attributes are retained.  Below this zoom, it is unlikely that a style might want differentiated bridge/tunnel based on feature density.

In addition to saving space by removing these attributes, it also allows for improved generalization of transportation layer LineString objects, which can be better simplified when they are not disrupted by tiny bridge sections.


eva-j:
This is actually reverting PR we did last year https://github.com/openmaptiles/openmaptiles/pull/1038.
There were people complaining they cannot hide tunnels or tunnel labels.

ZeLonewolf:
Perhaps we could limit the brunnel suppression to bridges that are too small to show, which is the real space waster.

ZeLonewolf:
Initial result removing tiny bridges from z10-11:
image

ZeLonewolf:
Screen shot of the Lake Ponchartrain causeway at zoom 6 showing bridge atttributes:

image

TomPohys:
Thanks.. removing attribute from very small bridges (less than 3 pixels) make sense! When I tested it on the Czech Republic I found some bridges were split and merging with the surrounding sections of the roads did not work so well (maybe out of the scope of this PR). Example of not merged road https://www.openstreetmap.org/way/327298025 with https://www.openstreetmap.org/way/327298027. The bridge has `layer=1` which causes not merging the road. 

When I looked at another bridge the roads were also split. e.g. 

- in OSM - [bridge](https://www.openstreetmap.org/way/919166440#map=19/49.59755/17.26720), [1st section](https://www.openstreetmap.org/way/919166439#map=19/49.59755/17.26720), [2nd section](https://www.openstreetmap.org/way/1041226999#map=19/49.59755/17.26720), [3rd section](https://www.openstreetmap.org/way/1082993698#map=19/49.59755/17.26720), [4th section](https://www.openstreetmap.org/way/1041226978#map=19/49.59755/17.26720), [5th section](https://www.openstreetmap.org/way/614258068#map=19/49.59755/17.26720)
- in a database, I assume that differences in `layer`, `z_order`, and `network` cause splits.

bridge 
1st_section
2nd_section

In Prague, I find a bridge where this logic is not working 
 - this is OK - https://www.openstreetmap.org/way/397818112
 - this not - [1st section](https://www.openstreetmap.org/way/476039093#map=18/50.07528/14.41252), [2nd section](https://www.openstreetmap.org/way/1112129282#map=18/50.07528/14.41252), [3rd section](https://www.openstreetmap.org/way/997236200#map=18/50.07528/14.41252)

Would be possible to implement this logic after merging the roads? Or merging the bridges first and then deciding if it is long enough? I understand that this could be just an edge case, but still (another e.g. in Dubuque [1st](https://www.openstreetmap.org/way/132777539#map=15/42.5139/-90.6356), [2nd](https://www.openstreetmap.org/way/39350807#map=15/42.5139/-90.6356) section)

ZeLonewolf:
Adding the `level` suppression logic seems to make no difference at z10, presumably because there are only a small number of bridges that get suppressed at z10 but not at z11.

ZeLonewolf:
I think this is the best I can do easily for z9-10 with how the update functions are set up with triggers and ids. There is probably a way to further improve tile size but it would involve changing the way the update functions work and I'm concerned about breaking things. So I think this is ready for review at this stage.

TomPohys (approved):

PR #1290, opened 2021-11-04
Update "Verifying that updates still work"

zstadler:
Convert the instructions into a single script that anyone can copy-paste to a terminal.
Add `make destroy-db` to the initialization part.


CONTRIBUTING.md
```diff
@@ -48,26 +48,34 @@ It is recommended that you create a [unit test](TESTING.md) when modifying the b
...
+The verification requires the script to append temporary changes to the `.env` file. Please restore the original version from git using `git checkout .env` or remove these changes before submitting a PR.
 
 ```
+(
+set -e
+echo '# Temporarily changes for Verifying that updates still work' >> .env 
+# Ensure DIFF_MODE is active
+echo DIFF_MODE=true >> .env
+# Ensure all zoom levels are tested
+echo MAX_ZOOM=14 >> .env
```
ZeLonewolf, line 64: The above echo commands will add a line to .env on each run, which is not desirable.

Instead, perhaps replace with an inline sed script, similar to how it's done in the Makefile:

https://github.com/openmaptiles/openmaptiles/blob/be37f3a5659d785668aec7ec50901bca3021e137/Makefile#L619-L620

ZeLonewolf:
Would this be better as an actual shell script?  It could even accept area and MAX_ZOOM as arguments for bonus points.

CONTRIBUTING.md
zstadler: In what sense this is not desirable?
This is a temporary yer required change. By explicitly appending text to the file, it will be easier for the developer to notice this change and remove the trailing lines before making a commit.

To further assist in this, the trailing lines also have a header:
https://github.com/openmaptiles/openmaptiles/blob/a4380366be14f3f11dc6c9575a42ad22160e8aaa/CONTRIBUTING.md#L60

zstadler:
> Would this be better as an actual shell script? It could even accept area and MAX_ZOOM as arguments for bonus points.

I would not recommend running this test with a different MAX_ZOOM. A developer may thing a change does not impact some zoom levels, yet one of the goals of testing is to check the developer's assumptions.

CONTRIBUTING.md
ZeLonewolf: > In what sense this is not desirable?

It leaves behind a .env file with the same variable set twice to different values.  So if you don't remember to go back and reset things, you're left with a confusing config file.  Particularly, if you inspect the .env file you might reasonably look at those variables in their original location and not think to scroll down to the end where there's a second copy of the variable with a different value.  I'm not sure what the value is of appending the new variable versus changing the existing one and having the setting in one place.

zstadler: For this test to perform its task, the `.env file must be changed, and these changes must be reverted before submitting a PR.

It leaves us to consider how the changes are to be made and what are the goals of this changes. 
- Both the existing and your proposed approach fulfill the functional goal of changing the values of `DIFF_MODE` and `MAX_ZOOM`.
- The "bad coding style" and the comment inserted into the .env file serve another goal: clarify that these changes are not meant to be committed into the repository. For example, `git diff .env` will clearly show that "something is wrong" with the .env file
   image
   
   rather than the developer forgetting that 2 good-looking changes were meant to be removed.
   image

I suggest we agree to disagree.

TomPohys: Thank you both for the discussion! I understand the pros and cons of both versions. I hope that there will be no problem with addition lines in `.env` file when it is mentioned explicitly. The copy-paste function with "hard-coded" `DIFF_MODE` and `MAX_ZOOM` is maybe a little safer than if a user has to make a decision which zoom to use. Nevertheless, users could make changes in copy-paste code too and skip the part where are adding lines into `.env`.

CONTRIBUTING.md
```diff
@@ -48,26 +48,34 @@ It is recommended that you create a [unit test](TESTING.md) when modifying the b
...
-1. Change .env to set `DIFF_MODE=true`.  In addition, if you're testing changes that impact zooms higher than the default zoom of 7, you should also change `MAX_ZOOM` to be up to 14.
-2. Set the test area to the appropriate geofabrik extract, for example:
+**Note:**
+
+The verification requires the script to append temporary changes to the `.env` file. Please restore the original version from git using `git checkout .env` or remove these changes before submitting a PR.
 
 ```
+(
+set -e
+echo '# temporary changes for verifying that updates still work' >> .env
```
TomPohys, line 60: It is possible to add an empty line before the comment? It will be more visible in `.env` file.

```suggestion
echo -e '\n# temporary changes for verifying that updates still work' >> .env 
```

zstadler:
> 
> 
> Would this be better as an actual shell script? It could even accept area and MAX_ZOOM as arguments for bonus points.

@TomPohys,

Would you prefer that the code is moved out of `CONTRIBUTING.md` and into a separate script?

TomPohys:
In the first step, it is OK to keep in `CONTRIBUTING.md`. If it will be unsatisfactory, it could be moved.

TomPohys (approved):

TomPohys:
Thank you. It looks good.

Issue #406, opened 2018-01-11
Waterway outside river polygon on z7-9


daliborjanak:
Rivers are outside river polygons on zooms 7, 8 and 9. This is because rivers are generalized too much. See screenshots:

rivers - red colour
lakes - blue

Zoom 7-9 -river is outside polygon
![image](https://user-images.githubusercontent.com/1891090/34824113-763eb56a-f6cd-11e7-9ebc-6b60eb7a38cf.png)

Zoom 10 - river is in polygons
![image](https://user-images.githubusercontent.com/1891090/34824161-aa10e4bc-f6cd-11e7-851a-98026f009ad3.png)

Solutions:
Reduce generalization of rivers on critical zooms

jirik:
The reason is rivers are taken from NE, but lakes are taken from OSM at Z6-Z8. Related to #310.

Solution is to unify switch between NE and OSM to one zoom level (in this case at least for waters).

jirik:
@daliborjanak @klokan I can fix this one by showing generalized OSM for Z6-Z8 instead of NE, but another problem appear. Some river parts are missing:
![image](https://user-images.githubusercontent.com/2271850/38741367-196f0230-3f3a-11e8-8d7a-8f3239963eb7.png)
We already have the same problem problem in the schema for Z9+:
https://openmaptiles.org/inspect/#9.19/47.4760/7.5466
If I fix issue described by Dalibor, more rivers appear for Z6-Z8, but more river parts will be missing (see below). E.g. at Z6 most of river Aare is missing.

What's your opinion about it? Should it be fixed despite of missing river parts?

Suggested changes: https://github.com/klokantech/openmaptiles/commit/ac1d70fa3a869a9b86c969008bd5bc68dec8d4fa

Z9 (current, OSM)
![image](https://user-images.githubusercontent.com/2271850/38743302-0d8486de-3f3f-11e8-8353-7ecdb52a7d64.png)

Z8 (suggested, OSM)
![image](https://user-images.githubusercontent.com/2271850/38743279-f4b4ecc0-3f3e-11e8-8e09-a93d83419351.png)

Z7 (suggested, OSM)
![image](https://user-images.githubusercontent.com/2271850/38743344-26ef6774-3f3f-11e8-88dc-2c30cef8b9ed.png)

Z6 (suggested, OSM)
![image](https://user-images.githubusercontent.com/2271850/38743408-4c2a0544-3f3f-11e8-9d6b-b7c34397a789.png)

[...]

Issue #320, opened 2017-08-21
Lake and river related issues 


WBMol:
There are a couple of visual complications when dealing with water and lakes, at least in the Netherlands. Though these are mostly specific to our use case, I think more people would benefit from possible solutions.

**Zoom inconsistency**
First off, water is either 'ocean' or 'lake', and its classification is inconsistent among zoom levels. A related issue can be found here: https://github.com/openmaptiles/openmaptiles/issues/310

A prominent example can be found in the Zeeland province (south west NL). Zoom 0-8 shows this: 
![image](https://user-images.githubusercontent.com/13695103/29509441-935f6d8c-8658-11e7-99dd-26d7b74ae40c.png) where only the ocean and a custom lake (IJselmeer) has been coloured blue. 

Zoom 9+ turns the third from bottom ocean part into a lake, or in fact, the ocean part disappears. The lake polygons are present at lower zooms already (more or less duplicating the ocean): 
![image](https://user-images.githubusercontent.com/13695103/29509548-f07c99ea-8658-11e7-9fba-d62649066287.png)

**Scale rank of 'lakes'**
This would ordinarily not pose a problem, as you can use lake data to colour that part blue again. The problem here is, however, that a large lake like IJsselmeer can not be seperated from the smallest ponds and water bodies. And the Netherlands is full of the latter, which adds distractions at zooms 9+. Idealy, some sort of scale rank metadata to optionally filter these out would help a lot. Or, maybe even better, label ditches and ponds as such, instead of 'lakes'. 

An example of our base layer with the 'lakes' enabled between Amsterdam and Utrecht: 
![image](https://user-images.githubusercontent.com/13695103/29510014-9457e424-865a-11e7-9396-f7edf0a8b9eb.png)

Our current workaround is manually adding polygon data of lakes. This introduces new problems of course, like minor dataset inconsistency or missing lakes depending on the source dataset, resolution/quality issues etc. 

**Names of lakes**
Large lakes have no name attributes, though they would be nice to have. 

**Interrupted rivers**
While on the topic of water, I'd like to mention how some rivers are interrupted. Here an example of the city of Utrecht, which has a nice canal network. At lower zooms, there are gaps. Example at zoom 12 vs 13:
![image](https://user-images.githubusercontent.com/13695103/29510803-65e59f34-865d-11e7-9c7d-57b75208073a.png)
[...]

PR #1291, opened 2021-11-04
Waterway zoomlevel 6 - 8 corresponding with water

eva-j:
This PR is solving https://github.com/openmaptiles/openmaptiles/issues/406, partially also https://github.com/openmaptiles/openmaptiles/issues/320.
It is changing source of data for zoomlevel 6 - 8 in `waterway` layer to OSM (relation `waterway = river`), this way it corresponds with `water` layer.
Testing on Europe, z0 - z9 tiles increase size from 20.7 MB to 21.4 MB. We can discuss the level of generalization, but I think this size change is not critical.

____

Preview
z6 before:
z6_master

z6 after:
z6_new

____
z7 before:
z7_master

z7 after:
z7_new

____
z8 before:
z8_master

z8 after:
z8_new


TomPohys (approved):

TomPohys:
Awesome, it looks great! Thank you @eva-j.

TomPohys (approved):

TomPohys:
Thank you for the fix!

PR #1292, opened 2021-11-05
Convert transportation_name generalization to materialized views

ZeLonewolf:
This PR converts the `transportation_name` layer import and update code to use materialized views rather than a table and view configuration.  This will drastically simplify the update code and reduce the potential for errors.

Setting as draft for now in order to verify performance is acceptable and write unit tests to verify that the update code is working as expected.


nyurik:
@ZeLonewolf is this a revert of #996 ?

frodrigo:
Not at all!
This revert the ability to do quick update on the data.

We need to drop way all still present materialized view, not add more.

@ZeLonewolf look at #814

ZeLonewolf:
Abandoning this attempt for now.

PR #1294, opened 2021-11-06
Add unit tests for the transportation layer

ZeLonewolf:
This PR adds unit tests for the transportation layer, and fixes several bugs in the unit testing code.  It verifies the correct zoom association to the various highway classes, and verifies that updates to highway attributes are populated.

Additionally, the documentation for unit tests in the CONTRIB guide is updated to note the specific make commands for developers to run.

PR #1295, opened 2021-11-06
BUGFIX: Fix name-based way fragmentation in transportation_name

ZeLonewolf:
I discovered this bug while investigating issues with the updates process related to #1190 #1292, and #814.

The `transportation_name` layer produces slightly different `tags` hstore values in the `osm_transportation_name_linestring` table during the initial import versus when running an update.  As currently written, the import code produces null-value keys in the `tags` column, while the update code suppresses them.  This PR removes that difference and makes the import code use same method that is currently used in the update code.

With a test case I've written, the import code produces a tags hstore that looks like this:
`"name"=>"OpenMapTiles Secondary 2", "name:de"=>NULL, "name:en"=>NULL, "name_int"=>"OpenMapTiles Secondary 2", "name:latin"=>"OpenMapTiles Secondary 2"`

...while the update code produces a tags hstore that looks like this:

`"name"=>"OpenMapTiles Secondary 2", "name_int"=>"OpenMapTiles Secondary 2", "name:latin"=>"OpenMapTiles Secondary 2"`

Note the missing NULL values.

This bug causes a small amount of space wastage after an update is run, because the update matching code detects the `tags` value as different, resulting in a duplicate copy of the tags value if that row is updated.  This causes duplicate objects and breaks GROUP BY clauses that expect to group same-tagged features together.  I've tested this by inspection of a generated mbtiles, database spot checks, and the unit test code included in this PR.

This PR incorporates the changes in #1294, which were necessary for testing this PR.


ZeLonewolf:
Not ready for prime time yet.

layers/transportation_name/transportation_name.sql
```diff
@@ -28,9 +28,9 @@ CREATE OR REPLACE FUNCTION layer_transportation_name(bbox geometry, zoom_level i
 AS
 $$
 SELECT geometry,
-       name,
-       COALESCE(name_en, name) AS name_en,
-       COALESCE(name_de, name, name_en) AS name_de,
+       tags->'name',
```
TomPohys, line 31: Just `SELECT tags->'name'` return unnamed column. Maybe it will be better to explicit the name.
```suggestion
       tags->'name' AS name,
```

layers/transportation_name/transportation_name.sql
```diff
@@ -28,9 +28,9 @@ CREATE OR REPLACE FUNCTION layer_transportation_name(bbox geometry, zoom_level i
 AS
 $$
 SELECT geometry,
-       name,
-       COALESCE(name_en, name) AS name_en,
-       COALESCE(name_de, name, name_en) AS name_de,
+       tags->'name' AS name,
+       COALESCE(tags->'name_en', tags->'name') AS name_en,
+       COALESCE(tags->'name_de', tags->'name', tags->'name_en') AS name_de,
```
TomPohys, line 33: Shouldn't these be `name:en` and `name:de`?
```suggestion
       COALESCE(tags->'name:en', tags->'name') AS name_en,
       COALESCE(tags->'name:de', tags->'name', tags->'name:en') AS name_de,
```

When as it is now `tags->'name_en` and `tags->'name_de'` are taken from `osm_transportation_name_linestring_genX` and `osm_transportation_name_linestring`. 

Where `osm_transportation_name_linestring` is created from `osm_transportation_name_network` and `osm_shipway_linestring` where is using new function `transportation_name_tags` which creating tags `name`, `name:en` and `name:de`?

TomPohys:
Thank you for this PR. It looks awesome. Just a little comment for clarification if `name_en` shouldn't be `name:en`.

layers/transportation_name/update_transportation_name.sql
```diff
@@ -294,10 +269,7 @@ BEGIN
     FROM (
         SELECT hl.geometry,
             hl.osm_id,
-            CASE WHEN length(hl.name) > 15 THEN osml10n_street_abbrev_all(hl.name) ELSE NULLIF(hl.name, '') END AS name,
-            CASE WHEN length(hl.name_en) > 15 THEN osml10n_street_abbrev_en(hl.name_en) ELSE NULLIF(hl.name_en, '') END AS name_en,
-            CASE WHEN length(hl.name_de) > 15 THEN osml10n_street_abbrev_de(hl.name_de) ELSE NULLIF(hl.name_de, '') END AS name_de,
-            slice_language_tags(hl.tags) AS tags,
+            transportation_name_tags(hl.geometry, hl.tags, hl.name, hl.name_en, hl.name_de) AS tags,
```
ZeLonewolf, line 272: This line is causing the current unit test failure:
`column "tags" is of type hstore but expression is of type text`

ZeLonewolf:
The unit test CI is actually failing here.  There is something wrong with the unit test execution that is causing the failure to not bubble up into CI.

ZeLonewolf:
All issues are resolved!  The new test cases added in this PR also detected an issue with sac_scale not getting updates in the transportation table sequence; this  was corrected in https://github.com/openmaptiles/openmaptiles/pull/1295/commits/e53122d7465caf5c740d8b0ad0766fd29575911a.

Ready for review.

TomPohys:
Awesome! Thank you very much!

TomPohys (approved):

PR #1303, opened 2021-11-13
Implement feature IDs and non-locking updates in the transportation layer

ZeLonewolf:
This PR implements feature IDs and non-locking updates for the highway portion of the transportation layer.

Feature IDs are implemented as the OSM id, or the smallest OSM id of a way in a grouped/generalized feature.

Non-locking updates are implemented by changing the REFRESH MATERIALIZED VIEW sequence to REFRESH MATERIALIZED VIEW CONCURRENTLY, with supporting unique indexes that are required for concurrent materialized views.  This will allow updates to happen without locking reads on the the materialized views.

Sanity-check screens shot of an objects in the transportation/transportation_name layer is provided below to show no change in the output data tiles.

**High zoom**:
image

**Low zoom**:
image

For consistency, in the SQL layer, columns are named `osm_id` when a record corresponds 1:1 with an OSM feature, and `id` for derived IDs, in this case min(osm_id).

In addition, an unnecessary `ST_Dump` call was removed from the z11 materialized view, which was causing extra records to be created in that view.  This was replaced with simple line merging, which groups together connected features with identical tagging.

### DIFF
--- a/layers/transportation/transportation.sql
+++ b/layers/transportation/transportation.sql
@@ -11,7 +11,7 @@ $$ LANGUAGE SQL IMMUTABLE
 CREATE OR REPLACE FUNCTION layer_transportation(bbox geometry, zoom_level int)
     RETURNS TABLE
             (
-                osm_id    bigint,
+                id        bigint,
                 geometry  geometry,
                 class     text,
                 subclass  text,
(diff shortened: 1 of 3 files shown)


ZeLonewolf:
Needs re-work

ZeLonewolf:
Marking draft to experiment with performance improvements

ZeLonewolf:
Abandoning this attempt - does not seem possible to achieve the stated goal and keep it performant.

PR #1305, opened 2021-11-14
Remove fragmentation from transportation layer z12-14

### LINKED ISSUE #1304: Tile size inflation in z12-14 transportation layer due to fragmetation
There is currently significant tile size inflation in z12-14 due to way fragmentation.  At these zooms, no geometry merging is performed, therefore each way in OSM is presented as a separated feature in the tile.  Consider a road section as follows:

image

In this case, the mapper has split the road into three segments, in order to tag the middle section as having a different number of lanes.  (For simplicity sake, the additional lane tagging that would also be present is omitted here).

However - OpenMapTiles does not present the `lane=*` tag to the tiles.  Therefore, what is presented to the `transportation` layer is simply this:

image

Similarly, the same issue occurs in the `transportation_name` layer, which would see:

image

This geometry can be merged, and in doing so will eliminate the repetition of having three database rows for an object that can be represented as a single feature.

I noticed this issue in the implementation for #1303, where I fixed a bug in the merging code which dropped the tile size by as much as 30% in some layers, which is a significant difference.  However, the fix for #1303 only impacted the existing materialized views, which are only present in zoom 11 and lower.

In order to fix this, an additional materialized view is needed to hold the z12-14 merged geometries as well as an investigation of the `transportation_name` layer to ensure that it's using merged geometries wherever practicable.  This will add a small amount of size to the database in exchange for a reduction in tile size.

ZeLonewolf:
Closes #1304
Blocked by #1303

This PR fixes fragmentation in the high zooms (12+) of the transportation layer by adding an intermediate materialized view which merges geometry with identical tagging.  A full description of the problem can be found in #1304.

This PR is held in draft until #1303 is merged so that the tile size impact can be assessed.


jonasmatthias:
What is the status of this approach?

I noticed that https://github.com/openmaptiles/openmaptiles/pull/1303#issuecomment-968358403 was abandoned. But it seems to me that solving https://github.com/openmaptiles/openmaptiles/issues/1304 is still relevant.

ZeLonewolf:
I was not able to accomplish the fix within acceptable performance constraints, but feel free to look at my attempt if it's useful to take another stab at it.

PR #1307, opened 2021-11-18
Remove blank name tags from park layer

ZeLonewolf:
This PR performs the same fix on the park layer that #1268 applied to the waterway layer.  It removes blank name, name_en, and name_de from the tiles.


TomPohys (approved):

TomPohys:
Thank you!

PR #1308, opened 2021-11-18
Implement non-blocking updates in the park layer

ZeLonewolf:
This PR implements non-blocking updates in the park layer.

The approach was to use the `REFRESH MATERIALIZED VIEW CONCURRENTLY` feature in postgres.  In order to achieve this, a unique index was added on the z4 dissolved park area table.  The `ST_Union` / `ST_Dump` sequence was changed to an explicit cluster DB scan (an equivalent operation) so that a unique osm ID could be generated from each dissolved polygon.

Below is a screen shot from Idaho, USA showing that the dissolved z4 still works as expected.

image


TomPohys:
Thank you very much, it looks good!

TomPohys (approved):

PR #1313, opened 2021-11-25
Add expressway tagging to the transportation layer

### LINKED ISSUE #1148: Expose expressway tagging to the transportation layer
Unblocks ZeLonewolf/openstreetmap-americana#22

It is common in North American-style maps to differentiate roads that do not meet all of the criteria for `highway=motorway` but still have expressway-like characteristics.  This need is described in depth in ZeLonewolf/openstreetmap-americana#22. with pictures.

In order to achieve this, I propose that OpenMapTiles consume the tags `motorroad=yes` and `expressway=yes` (combined >200K tags) using the following rule:

1. If a highway class is `trunk`, `primary`, `secondary`, `tertiary` or `unclassified`; and
2. The highway is tagged `motorroad=yes` or `expressway=yes`

Then:

That feature will be rendered in the tile as `subclass=motorroad`  Subclass is currently unused for these road classes.  Expressways versus non-expressways is an appropriate subdivision of these road classes.

The value `motorroad` is selected as the value because is the more common of the two tags `expressway=yes` and `motorroad=yes`.  However, if this is seen as too close to `highway=motorway`, we can use `expressway` as the value instead.

Alternately, if we do not wish to use subclass in this way, we could create a separate `motorroad`/`expressway` column as a boolean to expose this data.

ZeLonewolf:
Fixes #1148 

This PR adds expressway tagging to the `transportation` layer, by setting `expressway=1` for non-motorway roads tagged `expressway=yes`, and omitting the tag otherwise.  Additionally, I've added a few unit tests to verify that the expressway tagging is being imported and updated into the intermediate tables.

Here is an example of expressway tagging on US-1 in Rhode Island, USA:

image


TomPohys (approved):

TomPohys:
Thank you! It looks great!

PR #1320, opened 2021-12-02
Remove `STRICT` statement

### LINKED ISSUE #1319: Missing shipway and aerialway names
After bugfix #1295 there are missing `shipway` and `aerialway` names added in #1309 and #1179

The problem was detected after introducing the function `transportation_name_tags`.

After some digging, it looks like functions `transportation_name_tags`, `get_basic_names` and `get_latin_name` shoud not have the [`STRICT`](https://www.postgresql.org/docs/current/sql-createfunction.html#:~:text=RETURNS%20NULL%20ON%20NULL%20INPUT%20or%20STRICT%20indicates%20that%20the%20function%20always%20returns%20null%20whenever%20any%20of%20its%20arguments%20are%20null.) statement, which returns NULL if the input is NULL

> RETURNS NULL ON NULL INPUT or STRICT indicates that the function always returns null whenever any of its arguments are null. 

The `geometry` is used by [osml10n_get_name_without_brackets_from_tags](https://github.com/openmaptiles/openmaptiles-tools/blob/2b94a8fa869a868d64b3559435db5f5929f4d6f6/sql/zzz_language.sql#L72) in `get_latin_name` function

And [osml10n_get_name_without_brackets_from_tags](https://github.com/giggls/mapnik-german-l10n/blob/90b301117d0a46e8f86915cb73f6a99b83ed58ff/plpgsql/get_localized_name_from_tags.sql#L332) function is `STABLE` not `STRICT` (geometry is NULL by default)

If these functions (`transportation_name_tags`, `get_basic_names` and `get_latin_name`) have to stay in `STRICT` mode, then the `transportation_name_tags` needs to pass `geometry` in 

https://github.com/openmaptiles/openmaptiles/blob/7838221b163cdcecf7eab81ec18fa73757f024e5/layers/transportation_name/update_transportation_name.sql#L48

https://github.com/openmaptiles/openmaptiles/blob/7838221b163cdcecf7eab81ec18fa73757f024e5/layers/transportation_name/update_transportation_name.sql#L72

TomPohys:
Remove `STRICT` statement from `transportation_name_tags` function.

Concurrently with remove `STRICT` from `get_basic_names` and `get_latin_name` in OMT-T

FIX #1319

### DIFF
--- a/layers/transportation/highway_name.sql
+++ b/layers/transportation/highway_name.sql
@@ -9,7 +9,4 @@ SELECT hstore(string_agg(nullif(slice_language_tags(tags ||
                      ''), ','))
                      || get_basic_names(tags, geometry);
 $$ LANGUAGE SQL IMMUTABLE
-                STRICT
                 PARALLEL SAFE;
-
-


ZeLonewolf (approved):
LGTM!

TomPohys:
> LGTM!

Thanks!

PR #1333, opened 2022-01-02
Fix version check in quickstart.sh

ePirat:
When docker compose returns a version with v prefix,
the version check would incorrectly treat it as version component
of 000, leading to always failing the docker-compose version check
even if the version is recent enough.

This happens at least with docker-compose 2.2.1 where
  docker-compose version --short
returns v2.2.1 with the leading v.


guillaumerose:
I had the same issue and opened https://github.com/docker/compose/pull/9096

TomPohys:
Thank you! Sorry for the delay!

PR #1334, opened 2022-01-02
Fix paths and tracks in transportation layer z12 and z13

lazaa32:
This PR fixes a bug that causes that `track` lines disappear at z13. This bug was introduced in https://github.com/openmaptiles/openmaptiles/pull/1190, which adds rendering of paths and tracks at z12 and z13.

Before this PR:
z12: lines with `route_rank = 1` are added (no matter what `highway`).
z13: lines with `route_rank BETWEEN 1 AND 2` and `highway = 'path'` are added.
-> tracks with `route_rank=1` are added at z12 but not at z13.
GIF of z11, z12, z13, z14
master

After this PR
z12: lines with `route_rank = 1` and `highway IN ('path', 'track')` are added.
z13: lines with `route_rank BETWEEN 1 AND 2` and `highway IN ('path', 'track')` are added .
-> only tracks and paths are added at z12 and z13 (which was IMHO the goal of https://github.com/openmaptiles/openmaptiles/pull/1190)
GIF of z11, z12, z13, z14
pr

GIF showing the difference at z13.
diff_z13

@ZeLonewolf could you please check this PR as it is related to your work? I hope that I understood your goal to add rendering of paths and tracks correctly.

### DIFF
--- a/layers/transportation/transportation.sql
+++ b/layers/transportation/transportation.sql
@@ -376,15 +376,14 @@ FROM (
            AND
                CASE WHEN zoom_level = 12 THEN
                          CASE WHEN transportation_filter_z12(hl.highway, hl.construction) THEN TRUE
-                              WHEN n.route_rank = 1 THEN TRUE
+                              WHEN hl.highway IN ('track', 'path') THEN n.route_rank = 1
                          END
                     WHEN zoom_level = 13 THEN
                          CASE WHEN man_made='pier' THEN NOT ST_IsClosed(hl.geometry)
-                              WHEN hl.highway = 'path' THEN (
-                                                        hl.name <> ''
-                                                     OR n.route_rank BETWEEN 1 AND 2
-                                                     OR hl.sac_scale <> ''
-                                                   )
+                              WHEN hl.highway IN ('track', 'path') THEN (hl.name <> ''
+                                                                   OR n.route_rank BETWEEN 1 AND 2
+                                                                   OR hl.sac_scale <> ''
+                                                                   )
                               ELSE transportation_filter_z13(hl.highway, public_transport, hl.construction, service)
                          END
                     WHEN zoom_level >= 14 THEN
--- a/layers/transportation/transportation.sql
+++ b/layers/transportation/transportation.sql
@@ -376,15 +376,15 @@ FROM (
            AND
                CASE WHEN zoom_level = 12 THEN
                          CASE WHEN transportation_filter_z12(hl.highway, hl.construction) THEN TRUE
-                              WHEN hl.highway IN ('track', 'path') THEN n.route_rank = 1
+                              WHEN n.route_rank = 1 THEN TRUE
                          END
                     WHEN zoom_level = 13 THEN
                          CASE WHEN man_made='pier' THEN NOT ST_IsClosed(hl.geometry)
-                              WHEN hl.highway IN ('track', 'path') THEN (
-                                                                   hl.name <> ''
-                                                                OR n.route_rank BETWEEN 1 AND 2
-                                                                OR hl.sac_scale <> ''
-                                                                )
+                              WHEN hl.highway = 'path' THEN (
+                                                        hl.name <> ''
+                                                     OR n.route_rank BETWEEN 1 AND 2
+                                                     OR hl.sac_scale <> ''
+                                                   )
                               ELSE transportation_filter_z13(hl.highway, public_transport, hl.construction, service)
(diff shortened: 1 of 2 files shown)


layers/transportation/transportation.sql
```diff
@@ -376,15 +376,15 @@ FROM (
            AND
                CASE WHEN zoom_level = 12 THEN
                          CASE WHEN transportation_filter_z12(hl.highway, hl.construction) THEN TRUE
-                              WHEN n.route_rank = 1 THEN TRUE
+                              WHEN hl.highway IN ('track', 'path') THEN n.route_rank = 1
```
ZeLonewolf, line 379: Instead of this, add an entry for track to `layers/transportation/class.sql` in the function `transportation_filter_z12()`

layers/transportation/transportation.sql
```diff
@@ -376,15 +376,15 @@ FROM (
...
-                              WHEN hl.highway = 'path' THEN (
-                                                        hl.name <> ''
-                                                     OR n.route_rank BETWEEN 1 AND 2
-                                                     OR hl.sac_scale <> ''
-                                                   )
+                              WHEN hl.highway IN ('track', 'path') THEN (
+                                                                   hl.name <> ''
+                                                                OR n.route_rank BETWEEN 1 AND 2
+                                                                OR hl.sac_scale <> ''
+                                                                )
```
ZeLonewolf, line 387: This change is not correct, track should have universal rendering at z13, not conditional rendering, which will be fixed with the change noted in above in `transportation_filter_z12()`

ZeLonewolf (changes requested):
This PR should just be a one-line change to `transportation_filter_z12()` as noted in my review comments.

layers/transportation/transportation.sql
lazaa32: `track` added to `transporation_filter_z12` function.

layers/transportation/transportation.sql
lazaa32: Reverted.

lazaa32:
> This PR should just be a one-line change to `transportation_filter_z12()` as noted in my review comments.

This adds all `track` lines at z12 and z13.
@TomPohys can you please confirm this is OK?

ZeLonewolf (commented):
Pending confirmation with screen shots, this looks correct.

lazaa32:
@ZeLonewolf this PR atm with your suggestions added +2.7% at z12 and +1.8% at z13 which is quite a lot.

I need to clarify one thing. Was your intention to:
- at z12
  -  add the most important **tracks** (`route_rank=1`)
- at z13
  - add more important **tracks** (`route_rank between 1 and 2`)
  - add more important **paths** (`route_rank between 1 and 2`)
  
Or was your intention to:
- at z12
  -  add all **tracks**
- at z13
  - add all **tracks**
  - add more important **paths** (`route_rank between 1 and 2`)
 
Because with your review suggestion the latter is now implemented.
Thanks for the clarification.

ZeLonewolf:
The intent is for all tracks to render, which would be parity with osm-carto, and is what I described in my writeup for #1190.  The 2.7% increase, in my opinion, is a necessary growth in order to provide that parity.  The zoom filtering that I describe in that ticket is:

**Zoom 12**
* All highway=track
* Paths that are members of international & national hiking routes (network=iwn/network=nwn) (rank=1)

**Zoom 13**
* All highway=track
* Paths that are members of a hiking route, or have a name, sac_scale, osmc:symbol, or colour (rank=1,2)

**Zoom 14**
* All highway=track
* All highway=path

lazaa32:
After discussion with @TomPohys and @daliborjanak:
Adding all tracks at z12 and z13 increases tiles size too much and it adds too many lines at some places.
diff_z12

We decided to use the same filter (`route_rank=1` at z12 and `route_rank between 1 and 2` at z13) both for tracks and paths.

farfromrefug:
@lazaa32 a little bit out of subject. what tool.did you use to do the gif sliding comparaison? seems very useful

lazaa32:
MapLibre Compare plugin
https://github.com/maplibre/maplibre-gl-compare

farfromrefug:
@lazaa32 awesome thanks ! i am migrating mapbox mbview to maplibre. will try to add support for that plugin and and drag and drop support of mbtiles. really need something like this!

TomPohys (approved):

TomPohys:
Thank you!

PR #1353, opened 2022-02-10
Use `docker-compose up -d` for on-going processes

### LINKED ISSUE #1352: Use `docker-compose up -d` for on-going processes
Make targets `start-tileserver`, `start-maputnik`, and `import-osm` initialize on-going processes rather than perform a one-time task. It is more appropriate to execute them using `docker-compose up -d`. With the current `docker-compose run` execution, the user needs to run `make` as a background process. This PR also allows the use of `docker-compose logs` rather than the need to handle `stdout` and `stderr` redirection and avoids the dependency of `stdin` handling for `docker-compose run`

This is a follow-up for [a comment](https://github.com/openmaptiles/openmaptiles/commit/0cae7b9fd639f9cbad2a1ce86bbfbcaca4682670#r40912721) on an earlier commit.

> According to the [documentation of `imposm run`](https://imposm.org/docs/imposm3/latest/tutorial.html#run), invoked by `import-update`, it runs forever and repeatedly downloads daily/hourly/minutely updates from the server and updates the PostGIS DB. 
IMO, it be more appropriate to `up` it as a docker-compose service rather to `run` it with a shell command.

The comment received a 👍 from @TomPohys.

zstadler:
Resolves #1352

Make targets `start-tileserver`, `start-maputnik`, and `import-osm` initialize on-going processes rather than perform a one-time task. It is more appropriate to execute them in a detached container. With the current non-detached execution, the user needs to run `make` as a background process and redirect `stdin`, `stdout` and `stderr`.

This is a follow-up for [a comment](https://github.com/openmaptiles/openmaptiles/commit/0cae7b9fd639f9cbad2a1ce86bbfbcaca4682670#r40912721) on an earlier commit.

> According to the [documentation of `imposm run`](https://imposm.org/docs/imposm3/latest/tutorial.html#run), invoked by `import-update`, it runs forever and repeatedly downloads daily/hourly/minutely updates from the server and updates the PostGIS DB. 
IMO, it be more appropriate to `up` it as a docker-compose service rather to `run` it with a shell command.

The comment received a 👍 from @TomPohys.


TomPohys:
Hi, thank you for PR. Unfortunately, `start-tileserver` and `start-maputnik` runs through `docker` and not through `docker-compose`.

zstadler:
Hi @TomPohys,

The PR aims at preserving the use of `docker` for `start-tileserver` and `start-maputnik`.
The description was edited for clarity and `docker up -d` was corrected to `docker run -d`.

zstadler:
By the way, I cannot think of a reason why these two services are using `docker` while other services, including `postserve`, are using `docker-compose`.

TomPohys:
I think there will be no problem converting them into `docker-compose`. Do you have time to add it to this PR? 

Maybe also add `stop-tileserver`, `stop-maputnik`, and `stop-update-osm` into `Makefile`?

zstadler:
Will do!

docker-compose.yml
```diff
@@ -54,6 +54,10 @@ services:
       - ./build:/mapping
       - ./cache:/cache
 
+  update-osm:
+    <<: *openmaptiles-tools
```
zstadler, line 58: This line instructs the yaml parser to insert a copy of the `openmaptiles-tools` entry above.

TomPohys:
Thanks a lot. It looks much cleaner with all targets in `docker-compose`.

TomPohys (approved):

PR #1361, opened 2022-02-20
Coalesce duplicate route concurrencies

ZeLonewolf:
# Problem description

#1128 introduced route relation concurrency information in OpenMapTiles via the `route_X` attributes.  The original implementation assumed that there would be a single route relation for each `network` and `ref` pair.  However, it is increasingly common practice to tag a separate route relation for each direction of a route in order to provide awareness to routers and other data consumers of the directionality of a route.  This standard and growing practice is described on the [OSM wiki page on route directions](https://wiki.openstreetmap.org/wiki/Route_directions).  Thus, the naïve implementation of #1128 caused duplicate entries to be added as `route_X` attributes in the case where separate route relations were used for directional routes.

The following super-relation is an example of a route modeled with multiple route relations based on direction:
https://www.openstreetmap.org/relation/162114

Below is the resultant rendering in openstreetmap-americana, showing an incorrect shield duplication of Ohio, USA state route 104:
image

# Solution description
This PR adds grouping when computing route concurrency information, such that duplicate entries are coalesced in a predictable way.  Since this grouping is done only within a route membership join of a single member way, the computational complexity should be trivial.

Below is a re-rendering of the same region on openstreetmap-americana after implementing this change:
image


layers/transportation/update_route_member.sql
```diff
@@ -62,18 +62,19 @@ BEGIN
 
     INSERT INTO osm_route_member (id, osm_id, network_type, concurrency_index, rank)
     SELECT
-      id,
-      osm_id,
-      osm_route_member_network_type(network) AS network_type,
+      min(id),
+      min(osm_id),
+      osm_route_member_network_type(MIN(network)) AS network_type,
```
zstadler, line 67: Is there a special reason for using `min(..)` in lines 65-67 and `MAX(...)`  in lines 70-72?
Similar mismatch also between lines 99-101 and lines 104-106.

layers/transportation/update_route_member.sql
```diff
@@ -62,18 +62,19 @@ BEGIN
...
-           WHEN network = 'lwn' THEN 2
-           WHEN osmc_symbol || colour <> '' THEN 2
+           WHEN MAX(network) IN ('iwn', 'nwn', 'rwn') THEN 1
+           WHEN MAX(network) = 'lwn' THEN 2
+           WHEN MAX(osmc_symbol) || MAX(colour) <> '' THEN 2
       END AS rank
     FROM osm_route_member rm
     WHERE rm.member IN
       (SELECT DISTINCT osm_id FROM transportation_name.network_changes)
+    GROUP BY member, network, ref, network_type
```
zstadler, line 77: IMHO, `GROUP BY ... network` allows replacing `MAX(network)` and `MIN(network)` with plain `network` in lines 76, 70, 71, 101, 104, and 105

layers/transportation/update_route_member.sql
ZeLonewolf: On further review, I've change the approach to use a unique constraint, which avoids these min/max shenanigans.

layers/transportation/update_route_member.sql
ZeLonewolf: Updated to change approach so this is no longer necessary.

zstadler:
The route relation concurrency information seems to be specific for the OSM Americana style. Wouldn't is be better to have this information as a private layer of an Americana fork of this repository?

ZeLonewolf:
> The route relation concurrency information seems to be specific for the OSM Americana style. Wouldn't is be better to have this information as a private layer of an Americana fork of this repository?

Route concurrencies are a global feature for any renderer that wishes to display pictoral route shields. Concurrent pictoral route symbols are show in paper maps all over the world. The fact that OSM Americana was the first style to figure out how to do it technically doesn't take away from it being a general-purpose feature.

TomPohys (approved):

TomPohys:
Looks great! Thank you for investigating!

Issue #68, opened 2016-11-29
Show Islands


lukasmartinelli:
Have islands in the place layer.
http://wiki.openstreetmap.org/wiki/Tag:place%3Disland

stirringhalo:
Good to close this issue?

PR #1362, opened 2022-02-22
Add documentation on islands in place layer

flother:
Islands have been part of the OpenMapTiles schema for more than five years (see commits d664a2fe and 01db9a4e and issue #68), but they're not documented in [the layer definition](https://github.com/openmaptiles/openmaptiles/blob/master/layers/place/place.yaml) or [on openmaptiles.org](https://openmaptiles.org/schema/#place). This PR fixes that by updating `layers/place/place.yaml`.


TomPohys (approved):

TomPohys:
Thanks!

PR #1363, opened 2022-02-24
Allow non-default `PGPORT`

zstadler:
Also

- Remove `.env-postgres` and the need to synchronize it with `.env`
- Allow use of environment variables in addition to `.env`

Resolves #1354
Related to https://github.com/openmaptiles/openmaptiles-tools/pull/403

PR #1364, opened 2022-02-26
Restore controlability with environment variables

zstadler:
Since PR #922 the contents of `.env` are included in `Makefile` in order for `make` to by aligned with the `docker-compose` settings.

```
# Make all .env variables available for make targets
include .env
```

The down-side of employing the `include` mechanism is that the settings  in `.env` now take higher precedence than the shell environment variables. As a result, controlling the OpenMapTiles flow because more difficult. For example, tests for `DIFF_MODE=true` had to modify the contents of `.env` in order to work:
https://github.com/openmaptiles/openmaptiles/blob/b0e7f7884cc826d3176505ea840bb41f12463c75/.github/workflows/integrity.yml#L45-L47 and
https://github.com/openmaptiles/openmaptiles/blob/b0e7f7884cc826d3176505ea840bb41f12463c75/Makefile#L629-L630

Users were also faced with similar difficulties.

This PR restores the ability to control `make` and `quickstart.sh` using environment variable while keeping the use of the `.env` at a lower priority.

The result is restoring the ability to easily adjust the OpenMapTiles flow using environment variables, such as:
```
PGPORT=54321 DIFF_MODE=true ./quickstart.sh monaco
```

#### Notes: 
1. This PR depends on #1363
2. This PR includes some clean-up of `Makefile`


TomPohys (approved):

TomPohys:
Thanks! Looks good!

PR #1371, opened 2022-04-02
Update aerodrome_label class in DIFF_MODE=true

zstadler:
The class field was not updated if needed during the import of changes in DIFF_MODE=true.


TomPohys:
Awesome! Thanks!

Issue #1370, opened 2022-03-29
[schema][poi] add ATM and parcel lockers to schema to POI layer


ttomasz:
### Description

I saw that ATM were considered to be included in POI layer in #15 and I would like to propose opening the discussion again.  
Additionally with new tagging for parcel lockers I think it would be beneficial to add them too.  
I'm bundling them in this request together due to similarities in how name/brand/operator tag is used which I'll explain in a further section.

I'd like to see these at the highest zoom level (currently 14) since they would be relevant only when looking at a small area and for styles that want to show shops etc.

#### ATM
These are commonly used machines and people might want to see them on the map as they try to navigate to them.
There are about 190k features mapped in OSM which is quite significant but shouldn't cause the tiles to get too big and processing shouldn't be affected since they are pretty much just nodes.

I didn't see any more suitable class so I think using "atm" as class would be fine.

tag: `amenity=atm`
proposed class: `atm`

#### Parcel lockers
These are functionally similar to post boxes which are already in "post" class. They are highly popular in some countries e.g. Poland and similarly people might want to see them on the map when they try to find them so they can use them.
Recently proposal was approved by community changing their tagging from subclass of vending machine to their own tag under amenity.
There are few thousand of them mapped currently (plus another 10k with old tagging waiting to be retagged) but total number of machines in Poland is something around 18k and they are gaining popularity in Germany (GLS or DHL have their own machines), USA (Amazon is starting to set up their parcel hubs) and other countries.
Similarly to ATMs they are pretty much just points.

tag: `amenity=parcel_locker`
proposed class: `post` (or whatever `post_box` is in)

### name vs brand/operator
ATMs, parcel lockers, and post boxes have a property in common: they do not rely on `name` tag, they mostly use `brand` or `operator` values and sometimes - mostly parcel lockers - `ref`.

Not sure how this could be solved here... We could overwrite `name` with `brand` or `operator` value but not sure how that plays with `name:XX` for different languages. Adding additional fields like `brand` or `operator` doesn't seem feasible.

ttomasz:
@TomPohys I created PR for ATMs

ttomasz:
@TomPohys I created the second PR with parcel locker features

PR #1375, opened 2022-04-07
adds ATM to POI layer

ttomasz:
issue: #1370 

This PR adds ATM to POI layer.
If ATM isn't provided with name tag we'll try to use operator or network instead.

I'll  prepare PR to website repo and maybe one of the styles.


TomPohys:
Thank you very much. The small tile size raise on z14 was expected.

TomPohys (approved):

PR #1377, opened 2022-04-24
Remove unneeded waterway='' check in water.yml

ZeLonewolf:
This PR removes an unnecessary check in water.yml.

generated SQL before:
```
SELECT CASE
           WHEN waterway='riverbank' THEN 'river'
           WHEN "waterway" = 'dock' THEN 'dock'
           WHEN "water" = 'river'
               OR "waterway" = 'riverbank'
               THEN 'river'
           WHEN "waterway" = '' THEN 'lake'
           ELSE 'lake'
           END;
```

generated SQL after:
```
SELECT CASE
           WHEN waterway='riverbank' THEN 'river'
           WHEN "waterway" = 'dock' THEN 'dock'
           WHEN "water" = 'river'
               OR "waterway" = 'riverbank'
               THEN 'river'
           ELSE 'lake'
           END;
```

The `waterway=''` test is purely redundant because it returns the same value as the fall-through ELSE case: `lake`.


TomPohys (approved):

TomPohys:
Thanks! Good catch! Seems reasonable.

TomPohys (approved):

PR #1384, opened 2022-05-05
Bugfix: fix update_osm_route_member() function

### LINKED ISSUE #1374: network_type in osm_route_member is null after updates
After running daily update via `import-update` command, values of column `network_type` in `osm_route_member` table become null in updated rows.
Tested on  `us-south-220101.osm.pbf`.

After running daily import-update (till 22-01-02) there are 79111 osm_id in `transportation_name.network_changes` (dumped as `future_network_changes` table)
```
SELECT count(osm_id) FROM transportation_name.network_changes;
 count 
-------
 79111
```

Before running `import-update` (just after initial `import-osm` and `import-sql`), there are 87938 rows in `osm_route_member` that are about to be updated and most of them have `network_type`
```
SELECT count(member) FROM osm_route_member WHERE member IN (SELECT osm_id FROM future_network_changes);
 count 
-------
 87938

SELECT count(*), network_type FROM osm_route_member WHERE member IN (SELECT osm_id FROM future_network_changes) GROUP BY network_type;
count | network_type
-------+---------------
 15320 |
  3132 | us-interstate
 40972 | us-highway
 28514 | us-state
```

After daily `import-update` (till 22-01-02) there are 88003 rows in the table `osm_route_member` with corresponding osm_id and most of them do not have `network_type`
```
SELECT count(member) FROM osm_route_member WHERE member IN (SELECT osm_id FROM future_network_changes);
 count 
-------
 87938

SELECT count(*), network_type FROM osm_route_member WHERE member IN (SELECT osm_id FROM future_network_changes) GROUP BY network_type;
 count | network_type
-------+---------------
 63865 |
  2429 | us-interstate
  9512 | us-highway
 12197 | us-state
```

Table `osm_route_member` is mapped here:
https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/mapping.yaml#L430
But it gets altered and column `network_type` is added later during `import-sql`
https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/network_type.sql#L28

I suppose that during daily import-update all mapped columns are updated, but column `network_type` is not and values in updated rows become null.

There is a trigger that executes `update_osm_route_member`, but there are already null values in `network_type` column.
https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/update_route_member.sql#L47-L80

This also causes null values of column `network` in tables `osm_transportation_name_linestring` and `osm_highway_linestring` and thus wrong `ref_length` and `ref`. That's how we found there is a problem when some highway shields started to display e.g. `I 16` instead of `16`.

Any ideas? @ZeLonewolf

lazaa32:
Fixes #1374 

This PR fixes `update_osm_route_member()` function. Before this the column `network_type` became empty after daily-update.
This PR adds `network_type` into upsert together with `rank` and `concurrency_index` columns.
This bug had occurred once already and had been fixed in https://github.com/openmaptiles/openmaptiles/pull/1239 but then some columns were added and these new columns were not added into the upsert.

### DIFF
--- a/layers/transportation/update_route_member.sql
+++ b/layers/transportation/update_route_member.sql
@@ -75,7 +75,8 @@ BEGIN
     WHERE rm.member IN
       (SELECT DISTINCT osm_id FROM transportation_name.network_changes)
     ON CONFLICT (id, osm_id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index,
-                                           rank = EXCLUDED.rank;
+                                           rank = EXCLUDED.rank,
+                                           network_type = EXCLUDED.network_type;
 END;
 $$ LANGUAGE plpgsql;
 


TomPohys (approved):

TomPohys:
Thanks a lot! Good catch!

PR #1389, opened 2022-05-13
Change typecheck to perform within the current namespace/schema

### LINKED ISSUE #1388: Check for existing type in SQL files does not check if type exists in current schema
When doing a type check like `SELECT 1 FROM pg_type WHERE typname = 'city_place'` postgres will return true even if the type does not exist within the current namespace/schema. When using multiple schemas this can be problematic since the sql file will not create the type (because the check returns true) but adding a column with the type will fail.

To reproduce: have a type named city_place, route_network_type or public_transport_stop_type in another schema in the database, the import will fail. This is not a problem when running the stack normally or against a "clean" DB, but I run it against a non-docker DB that has other data in other schemas.

## Error:

```
ERROR:  type "route_network_type" does not exist
LINE 2:                 ADD COLUMN network_type route_network_type
                                                ^
QUERY:  ALTER TABLE osm_route_member
                ADD COLUMN network_type route_network_type
CONTEXT:  PL/pgSQL function inline_code_block line 4 at SQL statement
```

SahAssar:
This PR fixes #1388 by doing a test cast for the type and if it fails adding the type. The description of the problem is in that issue (It seemed like guideline was to open a issue with a description of the problem/bug).


TomPohys:
Awesome! Thank you!

TomPohys (approved):

PR #1390, opened 2022-05-14
Add POI parcel locker

### LINKED ISSUE #1370: [schema][poi] add ATM and parcel lockers to schema to POI layer

ttomasz:
closes #1370 

This PR adds parcel locker to post class.

Similarly to ATMs from PR #1375 if name is not present we try to use `brand` or `operator`. Here we also try to add `ref` to brand or operator since it can disambiguate parcel lockers that are next to each other (which happens).


TomPohys:
Awesome, thank you, and sorry for the delay!

TomPohys (approved):

Issue #798, opened 2020-03-25
housenumber and POI overlap


golubev:
Hello :wave:

We ran into an issue with housenumbers and POIs at our OMT maps. It is critical for us to show both the housenumbers and the POIs.

It seems that POIs and housenumbers pretty often share the same coordinates and they fight for being displayed on the map as they overlap. We tried to use `text-offset` and `text-allow-overlap` mapbox style properties but that didn't seem to look nice, especially on small buildings.

Have anybody faced the same issue? Perhaps, somebody can give an advice for a workaround or a solid fix. Thanks in advance :hand: 

I noticed that they both are being placed into buildings polygons centroids. I think that this is causing them to overlap:

https://github.com/openmaptiles/openmaptiles/blob/8fe9af23279067201450ec1e06dbb33e296ce085/layers/housenumber/housenumber_centroid.sql#L7-L13

https://github.com/openmaptiles/openmaptiles/blob/ff99a02aae287ed0546abe705c981e71c8618c33/layers/poi/update_poi_polygon.sql#L8-L14

eva-j:
Related to https://github.com/openmaptiles/openmaptiles/issues/766

daliborjanak:
Duplicate to #766

golubev:
If https://github.com/openmaptiles/openmaptiles/issues/766 is resolved - will we see on a map both a POI and a house number (without overlapping) for a house that has a corresponding feature in `osm_poi_polygon`?

Now for such houses OMT places both the POI and the house number in a centroid and they overlap.

eva-j:
@golubev You are right, for polygon POIs it is not resolved. For me this is style issue, I don't think there can be found a good solution on data side.

golubev:
I think that this can be resolved on the data side as well.

As well as we can play with offsets in style we may start placing houses numbers and polygon-POIs in different points by giving them different coordinates. But that would be quite tricky to implement for the non-convex polygons )

PR #1391, opened 2022-05-15
Tile duplicate housenumber filtering

ttomasz:
related to: #798 #766 

This PR introduces simple filtering of duplicate housenumbers.

Simple means that filtering is done withing the tile.
Duplicates are defined as same housenumber, street, block_number[1].

Duplicates are usually caused by POIs. People like to add addresses to them. Most POIs have names so to prioritize addresses we pick features without names first.

Formula is: `row_number() OVER(PARTITION BY concat(street, block_number, housenumber) ORDER BY has_name ASC) == 1`

Example before/after:
image
image

This is kind of change would require significant testing but I wanted to float the idea here for review.

---
[1] it's used in japan instead of street names, see: https://sive.rs/jadr


TomPohys:
Hi, sorry for the delay! I like the approach. After some performance tests, it looks excellent. 

The 765/6 or 2420 in the down left corner (14.82/50.11377/14.57381)
PR            |  Current state
:-------------------------:|:-------------------------:
image | image

TomPohys:
1 second after merge I realize, that I do not click on Approved.

ttomasz:
Well I thought if technical stuff was ok to kind of ask around in community if they know of cases where this could break but I guess just merging works too :sweat_smile:

PR #1393, opened 2022-05-19
Added various asian languages to YAML file

bgo-eiu:
This includes the addition of some widely spoken Asian languages to the YAML file: Punjabi, Nepali, Balochi, Pashto, Persian, Uyghur, Dzongkha, Gujarati, Maithili, Bengali, Pashto

### DIFF
--- a/openmaptiles.yaml
+++ b/openmaptiles.yaml
@@ -29,9 +29,13 @@ tileset:
   languages:
     - am # Amharic
     - ar # Arabic
+    - as # Assamese
     - az # Azerbaijani, Latin
+    - bal # Balochi
     - be # Belorussian
     - bg # Bulgarian
+    - bn # Bengali (Bangla)
+    - bo # Lhasa Tibetan
     - br # Breton, Latin
     - bs # Bosnian, Latin
     - ca # Catalan, Latin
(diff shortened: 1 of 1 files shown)


TomPohys:
Hi @bgo-eiu, thank you for the PR, and sorry for such a delayed reply. Could you please add statistics for each of the added languages? I think, there is no need to add languages like `name:bal` if there are just 27 occurrences (https://taginfo.openstreetmap.org/keys/name%3Abal)

You can search them in https://taginfo.openstreetmap.org/ and search text will be e. g. `name:bal`

jachym:
Please add Kurdish language too.

ZeLonewolf:
> there is no need to add languages like `name:bal` if there are just 27 occurrences

Conversely, if there's only 27 occurrences, then there's hardly any harm in including them.  We're doing quite a bit of work in Americana to render the user's preferred language, starting in ZeLonewolf/openstreetmap-americana#578.  Also, as noted in ZeLonewolf/openstreetmap-americana#586, many of the top languages spoken in the United States are not included in OpenMapTiles.  I'm hoping the OpenMapTiles project might consider simply allowing all language-variant `name:xxx` tags to be rendered in the tiles.  This should be a small data set as the vast majority of objects are tagged only in one language anyways.

PR #1401, opened 2022-06-13
Add support for county seats (capital=6)

ZeLonewolf:
Unblocks ZeLonewolf/openstreetmap-americana#384

This PR adds support for `capital=6` in the place layer, which is used for tagging county seats in the US.  A county seat is the capital of a county.  In addition, I reorganized the code a bit to be simpler.

As noted in ZeLonewolf/openstreetmap-americana#384, we would like to render county seats along with state and national capitals, as this is typical styling on American-style maps.

Render output:

image

### DIFF
--- a/layers/place/capital.sql
+++ b/layers/place/capital.sql
@@ -2,8 +2,8 @@ CREATE OR REPLACE FUNCTION normalize_capital_level(capital text)
     RETURNS int AS
 $$
 SELECT CASE
-           WHEN capital IN ('yes', '2') THEN 2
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
> You'd probably want to add 3 and 5 as well - these cities should be at least as important as 4 and 6. Even if you have no special rendering for level 3, you would still want to render them the same as level 4.

Good idea.  I've added this support in fe99d4d.

TomPohys (approved):

TomPohys:
Thanks! It looks great!

PR #1415, opened 2022-08-18
Class shipway

### LINKED ISSUE #1373: [BUG] Unify ferry tagging in transportation and transportation_name layers
In ZeLonewolf/openstreetmap-americana#266, it was noted that ferry lines are tagged differently in the `transportation` and `transportation_name` layers.  I suggest a fix of changing the attributes to `class=ferry` for both layer.  See discussion below:

> Shouldn't the class/subclass on transportation_name match the class/subclass on corresponding lines in transportation?

That would be a sensible assumption, but currently in the tiles transportation has class=ferry while transportation_name has subclass=ferry for some reason.  Seems like this may be a mistake, but hard to say since [the schema docs](https://openmaptiles.org/schema) don't say anything about ferries at all.

image

_Originally posted by @zekefarwell in https://github.com/ZeLonewolf/openstreetmap-americana/issues/266#issuecomment-1089588133_

TomPohys:
In `transportation_name` layer for `shipway` used Instead of `subclass` use `class` attribute for `ferry`.

class_ferry

resolve #1373


TomPohys:
Bring back the `subclass` to keep this PR without breaking changes. 
image

daliborjanak (approved):

PR #1034, opened 2020-11-05
Rename generic layer.sql to actual_layer_name.sql

lazaa32:
Following layers had generic `layer.sql` name.

- `aerodrome_label`
- `aeroway`
- `housenumber`
- `mountain_peak`
- `park`
- `place`
- `poi`
- `transportation`
- `transportation_name`
- `water_name`

This PR renames `layer.sql` to `aerodrome_label.sql`, `aeroway.sql`...

PR #1420, opened 2022-09-21
OpenMapTiles style

lazaa32:
This PR adds OSM OpenMapTiles style (inspired by OSM Carto) into OMT repo. The purpose of this style is to display every feature encoded in a vector tile.

The PR consists of:

- new `style` directory with svg icons, tileserver `config.json` file and `style-header.json`
- style.json snippets - split per layers - e.g. `aerodrome_label.json`, `aeroway.json`, `poi.json`, `water.json` etc
- new make targets: `make download-fonts` to download fonts, `make build-style` to build a complete style.json from snippets, `make build-sprite` to generate sprite files from svg icons.
- updated README

The goal of this is to have a style that shows all the features in vector tiles. Every PR, that modifies the schema in a way that might add any new features to vector tiles - e.g. adding new OSM tags, should also contain a modification in relevant style snippet so the change is propagated into the style and visible visually.

Related to a PR to OMT-tools:


pnorman (changes requested):
This change takes OpenStreetMap Carto's CC0 cartography and relicenses it as CC BY. While CC0 allows you to do basically anything you want with the cartography, this means that any improvements you do cannot be used by upstream. 

Additionally, most of the icons are OpenStreetMap Carto icons, modified only by moving the colour from MSS to the SVG. It's misleading to claim copyright on other's PD works.

LICENSE.md should be updated to add OpenStreetMap Carto contributors to the copyright line.

ZeLonewolf:
I was under the impression that there was a plan to move OMT back to CC0.  If that's on track, it would seem that the licensing issue would be moot.

nyurik:
If the code/knowledge was taken from Carto, than (while not required), it is a good form to mention that Carto stuff came from Carto.  Also agree about re-licensing PD work -- while technically not a legal issue (?), it is a bad form and should not be done in a public project like this.

Lastly, if OMT moves to CC0 (which I am highly hopeful!), the above still applies -- again, because we want to play nice and giv credit where credit is due.

nyurik:
More thoughts:
* I think it is bad to have differently named styling files in each sub-dir.  Makes it harder to process, and must be kept in sync with some other file, etc.  Instead, I think they should all be the same, named after the style they represent, e.g. `osm-bright.json`
* I do not like the "order" field. You currently have it as a hack to keep it identical with the original style. It will not be manageable in the long run -  keeping track of all order numbers across 20 files is non-scalable, and is a bad design. Instead, we need to have well defined groups, possibly with a number, e.g. `0-low` and `9-high`. The strings could really be anything - as long as all style files use the same pattern, and can be sorted.  This way, each styling rule will be combined in this order:  1) group-name, 2) order of the layers listed in the main tileset yaml file, 3) order within the style file.

nyurik:
and another thing (this *is* a big change after all, so i'm sure i will spot more later):

I think putting all icons into the same dir is also not scalabale. The whole idea of breaking the style files into multiple subdirs is to have a flexibiliy of what is included. Having all images in the same dir makes it harder to manage and track who needs what. If two layers use the same image (weird, but ok), make one depend on another.

lazaa32:
> I think putting all icons into the same dir is also not scalabale. The whole idea of breaking the style files into multiple subdirs is to have a flexibiliy of what is included. Having all images in the same dir makes it harder to manage and track who needs what. If two layers use the same image (weird, but ok), make one depend on another.

I wanted to have all svg icons separated and all in one dir since the tool to generate sprites files takes a one dirname as an input. But I did not think of the scenario when an user has just some layers enabled - then it is better to generate the sprite files just from icons that actually matter.

lazaa32:
> * I think it is bad to have differently named styling files in each sub-dir.  Makes it harder to process, and must be kept in sync with some other file, etc.  Instead, I think they should all be the same, named after the style they represent, e.g. `osm-bright.json`

First I had thought about keeping the name the same across the layers - e.g. `style-snippet.json` but then I remembered this https://github.com/openmaptiles/openmaptiles/pull/1034. We used to have generic `layer.sql` and then we renamed them to actual `water.sql`, `poi.sql` etc. But we might rename it to make it more understandable.

> * I do not like the "order" field. You currently have it as a hack to keep it identical with the original style. It will not be manageable in the long run -  keeping track of all order numbers across 20 files is non-scalable, and is a bad design. Instead, we need to have well defined groups, possibly with a number, e.g. `0-low` and `9-high`. The strings could really be anything - as long as all style files use the same pattern, and can be sorted.  This way, each styling rule will be combined in this order:  1) group-name, 2) order of the layers listed in the main tileset yaml file, 3) order within the style file.

I know we have discussed this in Firenze, but the OpenMapTiles style has about 200 layers. Sorting them into a few groups seemed a bit difficult. I couldn't think of a good grouping system as we have to keep the rendering order of all classes, subclasses, patterns, outlines, casings etc.

lazaa32:
The idea behind this PR is really to have just one main reference style that shows all features and if there are any OSM tags to be added to/removed from the schema, the modification must be propagated into the style as well.
I am happy to open a separate discussion about more styles but I'd like to get this done for the OMT style first, as this was announced to be part of the v3.14 release and it's the only thing that blocks the release at the moment.

nyurik:
I didn't mean that we should or shouldn't add more. I meant that having non uniform style filename is inconvenient for processing, and will block us from implementing more styles if we decide to do that. 

Let's name them all osmbright.json, or defaultstyle.json, or whatever.

klokan:
In this PR our team releases all the work on a complete OpenStreetMap vector style under an open-source licence.
It has been a lot of work. Now anybody can modify it, build upon, improve it, regenerate tiles and use everything even for commercial purposes - unlike with other complete vector tile OSM Carto implementations (i.e. ESRI).

Kindness is boosting cooperation and dialogue, let's try to stay at that level to find a way forward and to join this PR blocking 3.14 release, please...

@pnorman thank you very much for your alert. The credit to the original OSM Carto style is clearly provided already - we don't claim the work is only ours, it is heavily inspired by OSM Carto Mapnik style - we build here on the shoulders of amazing people and entire OSM community, including you - and you and others indeed deserve the credit (despite you waived such rights by selecting CC0 licence on the original OSM Carto style).
The original OSM Carto implemented for Mapnik renderer deserves more visibility here...
Let's (please @lazaa32):
1) add links to style/README.md - bringing people to the OSM Carto repository https://github.com/gravitystorm/openstreetmap-carto and it's license https://github.com/gravitystorm/openstreetmap-carto/blob/master/LICENSE.txt on places where already give credit to the original style and where we mention "OSM Carto" style.
2) create new README.md under style/icons/ - linking the exact location of the source of icons and describing modifications done

The inclusion of this style does not have any effect on the existing cartography inside of vector tiles and schema and does not give any reason for change of the licence of the entire repo as part of this PR (while a possible switch is considered in discussion with OpenStreetMap Foundation board).

@nyurik could you move the discussion WHY to have multiple style to a separate discussion thread (or a ticket) - with resolution post 3.14 release, please? 
No change to this PR is really required currently - modifications and refactoring can be done later on, if clear reasons are communicated, and if there is consensus and resources for doing and maintaining such work.
The reason why only ONE OFFICIAL OpenMapTiles style is here is that this is THE style where content of vector tiles should be demonstrated. This way, it is very clear what must be modified in future PRs to demonstrate new schema modifications. The alternative open styles have their own repositories linked from https://openmaptiles.org/styles/
 
The release 3.14 is going to have only one official style, called "OSM OpenMapTiles".

nyurik:
@klokan please see the https://github.com/openmaptiles/openmaptiles/pull/1420#issuecomment-1255470636 -- this is NOT about introducing multiple styles in this PR. This is about several bad design choices (in my opinion) that I think we can easily and quickly fix now, but they might be much harder to fix later on if not.  The biggest issues I see:
* each layer has a different file name for the style file -- it would make it easier to process if they all had the same file name
* the "order" field is too easy to break, and will cause numerous issues going forward. See [this comment](https://github.com/openmaptiles/openmaptiles/pull/1420#issuecomment-1254436892) above

About merging in general -- Petr, I am not blocking it. I am trying to help.  I wrote several PR feedback comments based on my experience as an engineer and as an OpenMapTiles contributor. If I didn't care about the project, I would simply stay quiet  instead. That's the whole idea of a PR, and that's why significant changes like this usually take some time, and are often discussed in an issue first -- simply because they may contain problems that should be addressed before merging.

Note that you can always release an OMT version from another branch/tag -- technically, there is no need to merge anything into the primary branch first.  Usually the state of the primary git branch has little to do with the release cycle. Big products create the release branch several months before the actual release -- the separate branch can then be tested, while the team continues the development without being blocked.

klokan:
@nyurik - thanks for your feedback. I believe "we can easily and quickly fix" any of these also later on... it is a filename and number in JSON - not a rocket science. Small things - can be easily improved later on.

Reaction to:
> - each layer has a different file name for the style file -- it would make it easier to process if they all had the same file name

OK. Let's call them `style.json` in every layer directory - as you said - the why argument is that processing with tools is easier this way.

IMHO slightly strange considering the Rename generic layer.sql to actual_layer_name.sql in #1034 - but not a problem for us.

The discussion on multiple styles is separated from this PR...

> - the "order" field is too easy to break, and will cause numerous issues going forward.

Agree. It is not ideal. We could have some "well defined groups" or ranges defined as you suggested - but this requires more effort and on the current setup with one style it does not cause any issues at all. There is not a better solution available right now - it may come with a new PR, if somebody has a pain on it.

klokan (approved):
I believe ready to be merged.

klokan (approved):

pnorman:
> I was under the impression that there was a plan to move OMT back to CC0. If that's on track, it would seem that the licensing issue would be moot.

Not really - something shouldn't be merged with problems under the assumption it will be fixed later, and changing the license on a project is generally a huge pain.

pnorman (commented):
*See review below*

pnorman (changes requested):
A re-review was requested, but I don't see that any of the issues I raised were fixed.

> This change takes OpenStreetMap Carto's CC0 cartography and relicenses it as CC BY. While CC0 allows you to do basically anything you want with the cartography, this means that any improvements you do cannot be used by upstream.

This issue remains.

> Additionally, most of the icons are OpenStreetMap Carto icons, modified only by moving the colour from MSS to the SVG. It's misleading to claim copyright on other's PD works.

This remains the issue as no changes have been made to the license file.

> LICENSE.md should be updated to add OpenStreetMap Carto contributors to the copyright line.

This remains an issue - with this PR the copyright line will be incorrect.

*Edit: Resubmitting as request changes*

klokan:
The license of the original OSM Carto for Mapnik allows everybody to "freely build upon, enhance and reuse the works for any purposes without restriction" - see https://creativecommons.org/share-your-work/public-domain/cc0/.

As such, this PR can be merged now.

daliborjanak (approved):

lazaa32:
> Hm just noticed that this merge breaks any custom layers I had before, because they do not have a style.json within them. I'm wondering what I should do... but I'm new to this project so I need to understand what to refactor here. Probably moving styles from some other styles file that the project had before.

Hi @thorsteinssonh,
It depends whether you want to use OSM OpenMapTiles style with your custom layers.
- If yes, then you have to add a `style.json` snippet as it is in other layers e.g. [poi style.json](https://github.com/openmaptiles/openmaptiles/blob/master/layers/poi/style.json). The style.json snippet has to contain `order` value as described in [README.md](https://github.com/openmaptiles/openmaptiles#work-on-layers)
- If you have your own custom layers and don't want to use OSM OpenMapTiles style then I would suggest to remove `build-style` target from `all` make target. Just remove `build-style` from Makefile at the line 236 https://github.com/openmaptiles/openmaptiles/blob/master/Makefile#L235-L236.

PR #1423, opened 2022-10-03
Add more office values

not-my-profile:
Omitted values include:

* office=yes ... too generic
* office=vacant ... not really a POI
* deprecated values (like office=administrative)
* tags that don't have a clear meaning (like office=camping, office=healthcare, office=transport, office=design)
* tags where another tag is used more often, e.g: office=religion is used more often than office=parish office=construction_company is used more often than office=construction


TomPohys:
Hi, I am very sorry for the delay! Thank you for adding `office` to the schema! Please also check the style to see if all newly added points are there. Could you also run `make generate devdoc` to update the [mapping_diagram.png](https://github.com/openmaptiles/openmaptiles/blob/master/layers/poi/mapping_diagram.png)?

TomPohys:
Thanks for the selection of "key=office".

PR #1427, opened 2022-10-15
fix: add parking to landuse as amenity

cbeddow:
Adding parking under landuse matching the amenity class


cbeddow:
@frodrigo can you check this and my other 2 pull requests adding classes to the schema to see if they look correct?

frodrigo:
I think adding parking in landuse as there are already in POI is not in the spirit of OMT.

cbeddow:
> I think adding parking in landuse as there are already in POI is not in the spirit of OMT.

It seems the parking polygon will not render however, only the parking icon. So I cannot represent a parking polygon on any tile map, its very confusing to a user I think. Parking lots are areas and not point locations.

Issue #1437, opened 2022-10-30
Parking is a way/polygon, but schema only display as a point icon


cbeddow:
Parking lots should be a polygon. These are not points in real life so the map is representing them improperly.

Can we add these to the schema under landuse? It looks very bad in current maps, see example here where the polygon is just "blank" with an icon, so it's the same color as the background of the map.

![image](https://user-images.githubusercontent.com/10554922/198889048-108b2f7a-98ce-423c-854e-d14aacc7418e.png)

See here in OSM Carto: 
![image](https://user-images.githubusercontent.com/10554922/198889061-7f9fe8b1-be6d-4db7-abb0-84b367402bc9.png)

I believe the map should show real polygons where they exist. it's not just landuse, but also represents a physically different surface and function, just like grass, sand, or water.

If "mapping for the renderer" then we otherwise would be forced to just map an area with a surface type, and have parking as a single node. No parking lot in the world is a single node, always a polygon, so I want to update the schema to reflect this. :)

PR #1429, opened 2022-10-16
Bench

cbeddow:
Bench is missing from amenities and not appearing on basemaps. I added it now in the same places were "drinking_water" appears as an amenity.

PR #1438, opened 2022-10-30
fix: add parking to landuse as amenity

cbeddow:
Discussion here: https://github.com/openmaptiles/openmaptiles/issues/1437


TomPohys:
Hi, do you have also some numbers, how much does it raises mbtiles on a bigger area (a state like France, or Germany)? (tests are using just some samples). Thanks

TomPohys:
And also (same as in https://github.com/openmaptiles/openmaptiles/pull/1429), please update PR by `make generate-devdoc` to update documentation pictures. Thanks

ZeLonewolf:
This is also a problem for OSM Americana, as well as any osm-carto-like style.  Without an area geometry, rendering looks a little silly:

image
[Location](https://zelonewolf.github.io/openstreetmap-americana/#map=17.05/41.491733/-71.310559)

TomPohys:
Thanks for bringing up this topic. Yes, having [parking areas](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparking) is a good idea. On the other hand, there are a lot of parking places that should not be visible or have a different style (e.g. [private parking](https://www.openstreetmap.org/way/210450875#map=19/49.59382/17.26508)). Also, [underground](https://www.openstreetmap.org/way/481358953#map=18/49.58735/17.25766) or [rooftop parking places](https://www.openstreetmap.org/way/287498932) are very common. Also showing the parking places until zoom level 14 (maybe 13) makes not much sense. 

If there will be more added attributes, then maybe using `transportation` layer for parking suits better.

I think that this needs more discussion.

ZeLonewolf:
In OSM Americana, we don't show parking until z17, so I'd agree that capping parking areas to z14 make sense.

There are a few places you could put it.
* You could put it in `transportation`, which gives you `subclass`, `surface`, `level`, `indoor` and `access` tags, which could all be encoded for parking areas.  One drawback is that it would add an area geometry to the `transportation` layer, and I'm not sure that we currently have areas in that layer.  Though I think this is not a real problem.
* You could put it in `landcover` which has a `subclass` and area geometries, but it's a bit of a stretch to say that it's the correct thematic layer for this.
* You could keep it here in `landuse` but add a new `subclass` column to capture the type of parking area
* You could put it in `poi`, which has `subclass`, `level`, and `indoor`, although this type of feature is not a "point"

Of all the options, I think putting it in `transportation` seems the least wrong to me, if we're OK with area features being in there.

TomPohys:
I thing too, that `transportation` layer is the least wrong. There is already some small table with the highway polygons. https://github.com/openmaptiles/openmaptiles/blob/8b5aa3273e26c88f3a6fdc1a34e2a0ee75d5c6ee/layers/transportation/transportation.sql#L814 

But it must be super careful, what parking places will be added into this layer (maybe it is not neccessary to have surface in tiles, definitely must be distinguish `private` or else inaccessible parking places etc.).

ZeLonewolf:
Perhaps we could start by excluding street-side parking areas, as these tend to be tiny and less interesting features to display as a polygon.

I would also suggest removing parking from the `poi` layer and moving it to the transportation layer so as not to duplicate the logic.  It's important that parking icons and parking areas be shown together.

ZeLonewolf:
@cbeddow wondering if you're planning to continue working on this?

cbeddow:
@ZeLonewolf yes I do! Sorry, I was unplugged for awhile in January and didn't have notifications on for this.

cbeddow:
@TomPohys when you ask for numbers: number of nodes? Square meters? Number of polygons? Data size? I can pretty quickly crunch this by querying my planet database but let me know which is best for similar proposals in the future

jdhoek:
> Perhaps we could start by excluding street-side parking areas, as these tend to be tiny and less interesting features to display as a polygon.

Street-side parking (and the shape of those areas) has a large impact on orienteering and navigation for all classes of map user. I wouldn't exclude those as areas at higher zoom levels. It would leave awkward gaps in many places.

ZeLonewolf:
@TomPohys can we run CI on this please so we can have some sense of the tile size metrics at least?

ZeLonewolf (approved):
@TomPohys 

This only adds 0.7% tile size at the highest zoom (and less at lower zooms).

Icons alone are not really sufficient for parking - this is the current Americana rendering for parking with icons alone.  IMO, having the polygons are necessary.

image

ZeLonewolf:
@cbeddow could you please run `make generate-devdoc` to complete this PR?

TomPohys:
Yes, `make generate-devdoc` will resolve the CI issue. 

I would suggest applying this change directly to the style.json to see it directly in OMT Style.

I would also suggest generating tiles of some area and at least check visually (and posting some images into the PR) if the parking space at zoom 12 and 13 makes even sense. And if not, force the parking area only for z14.

Thanks for the consideration.

TomPohys:
I am closing this in favor of #1798. Thanks for the discussion.

PR #1440, opened 2022-11-06
Show the Trans-Canada Highway at z4

### LINKED ISSUE #1439: Show the Trans-Canada Highway to zoom 4
Currently, the Trans-Canada highway, which is the most significant highway system in Canada, and is of equivalent importance to the US interstate system, is not consistently shown at zoom 4, which is the lowest zoom that highest-importance highways are shown.

This issue arose in ZeLonewolf/openstreetmap-americana#565.

The current map of the Trans-Canada highway (networks `CA:transcanada` and `CA:transcanada:namedRoute`) looks like this:
image

Because of Canada's lower population density and Arctic location, some of this network is only `highway=trunk` and not `motorway`.

I propose that trunk roads that are distinguished as part of a nation's most important highway network should get rendered at equivalent tile zoom levels to US interstates, starting with the Trans-Canada Highway.

ZeLonewolf:
Fixes #1439 

This adds support to show the Trans-Canada Highway at zoom 4.  Despite being the most important highway network in Canada, portions are `highway=trunk` due to the remoteness of the countryside.  However, it's still important to show a connected highway network at the lowest zoom without showing all trunk roads at this zoom.

This change also adds support for US Interstate Highways to be drawn at z4 when they're tagged as trunks.  There's only a few examples of this, but it removes those (tiny) gaps from the generalized road geometries. As we identify additional countries which should have their most important road network rendered at the lowest road zoom even if they're trunk, we can add them to the list of networks that get this treatment.

Tested on Prince Edward Island:
image

### DIFF
--- a/layers/transportation/update_transportation_merge.sql
+++ b/layers/transportation/update_transportation_merge.sql
@@ -349,7 +349,13 @@ BEGIN
     FROM osm_transportation_merge_linestring_gen_z5
     WHERE
         (update_id IS NULL OR id = update_id) AND
-        (highway = 'motorway' OR construction = 'motorway') AND
+        (highway = 'motorway'
+            OR construction = 'motorway'
+            -- Allow trunk roads that are part of a nation's most important route network to show at z4
+            OR highway = 'trunk' AND
+                network <> '' AND
+                network IN ('ca-transcanada','us-interstate')
+        ) AND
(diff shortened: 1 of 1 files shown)


TomPohys:
Hi @ZeLonewolf, thank you! It makes sense!

PR #1446, opened 2022-11-29
Sub-national routes in Canada and refactor national network list

### LINKED ISSUE #1442: Show Quebec Autoroutes at z4
As a followup to #1439 and ZeLonewolf/openstreetmap-americana#565 and after [community discussions](https://osmus.slack.com/archives/C01V02K52UX/p1668540341006239), it's appropriate to show the Quebec "autoroute" network at zoom 4, along with the Trans-Canada Highway and the US Interstate system.  These roads are tagged `network=CA:QC:A`.  According to [Wikipedia](https://en.wikipedia.org/wiki/Autoroutes_of_Quebec):

> The autoroute system is a network of freeways within the province of [Quebec](https://en.wikipedia.org/wiki/Quebec), Canada, operating under the same principle of controlled access as the [Interstate Highway System](https://en.wikipedia.org/wiki/Interstate_Highway_System) in the [United States](https://en.wikipedia.org/wiki/United_States) and the [400-series highways](https://en.wikipedia.org/wiki/400-series_highways) in neighbouring [Ontario](https://en.wikipedia.org/wiki/Ontario).

These roads also provide cross-border connectivity with the US interstate system, so including them at the same zoom level as the Trans-Canada network would present a visually-consistent view of the top-level highway system across this part of North America.

The Quebec Autoroute system looks like this:
image

This network has some discontinuities which will result in a few dangling roads, but this is an accurate reflection of the remoteness of Canada and it's road system.

The current road network at z4 looks like this:
image

Proposed is to bring it to this:
image

Note that the network now shows connectivity between New York and Vermont's cross-border Interstate highways.

ZeLonewolf:
Fixes #1442 

This PR defines the set of routes in Canada which are equivalent in national importance to the US Interstate Highway System, and are therefore appropriate to render at zoom 4. This creates a sane, connected highway network at this zoom level across the USA and Canada.

This draws upon the [work](https://osmus.slack.com/archives/C01V02K52UX/p1668636280056929) done by @zekefarwell to identify what the Canadian "top level" network should look like.  Unfortunately, due to differences in highway administration between the various Canadian provinces, the definition is [more complicated than we would like](https://github.com/openmaptiles/openmaptiles/compare/master...ZeLonewolf:canada-subnational?expand=1#diff-4ebc6f1fe3903e56019c0302a33df82fc4c00d62231f3df24ecf34f5fcc8f500).

This adds two additional network types for Canadian highways, ca-provincial for provincial-level roads, and ca-provincial-arterial for "highest importance" roads that are not part of the Trans-Canada highway but should be regarded as equivalent for low-zoom rendering purposes.

Additionally, this extracts out the country-specific network checks to a separate function in order to define "equivalent top-level networks" by country, providing a place to add additional national definitions as they're added by contributors.

Proposed Canada z4 network as rendered by this PR:

image

z5:

image
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -480,7 +480,7 @@ generate-changed-tiles: data/tiles.txt
 	fi
 
 .PHONY: start-tileserver
-start-tileserver: init-dirs
+start-tileserver: init-dirs build-style download-fonts
 	@echo " "
 	@echo "***********************************************************"
 	@echo "* "
--- a/layers/transportation/network_type.sql
+++ b/layers/transportation/network_type.sql
@@ -21,6 +21,17 @@ $$
     END
 $$;
 
+-- Top-level national route networks that should display at the lowest zooms
+CREATE OR REPLACE FUNCTION osm_national_network(network text) RETURNS boolean AS
+$$
+    SELECT network <> '' AND network IN (
+        -- Canada
+        'ca-transcanada', 'ca-provincial-arterial',
+        -- United States
+        'us-interstate');
+$$ LANGUAGE sql IMMUTABLE
+                PARALLEL SAFE;
+
 DO
 $$
     BEGIN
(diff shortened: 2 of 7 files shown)


TomPohys (approved):

TomPohys:
Thank you very much! It looks awesome!

PR #1454, opened 2022-12-14
Extend Natural Earth urbanized areas to z6 and rename

### LINKED ISSUE #1451: Extend Natural Earth urbanized area polygons to higher zooms
Presently, Natural Earth urbanized areas are rendered down to zoom 5.  We have implemented styling for this in openstreetmap-americana in ZeLonewolf/openstreetmap-americana#617:
image

These urbanized areas are tagged `class=residential` in the tiles.

Above zoom 5, OSM `landuse=residential` is rendered in the tiles, which is not the same thing, and does not flow at all from the z5 urbanized areas, as indicated in https://github.com/ZeLonewolf/openstreetmap-americana/pull/617#pullrequestreview-1216698007

Here is the cut from zoom 5.99 to zoom 6:

image
image

I propose the following changes:

1. Re-classify Natural Earth urbanized area polygons into a new class, `class=urbanized-area`, as these are not "residential" polygons and produce an unexpected result.
2. Render these Natural Earth polygons to the highest zoom that makes sense for urbanized area, likely zoom 9.

ZeLonewolf:
Fixes #1451 

This PR renames the Natural Earth urbanized areas from `class=residental` to `class=urban_area`, and extends it down to z6 from its present z5:

image

This allows for the continuity of urban area polygons separately from OSM `landuse=residential` polygons, which are not the same type of feature.


ZeLonewolf:
Per discussion in #1451

PR #1456, opened 2022-12-16
Add class=pond to the water layer

ZeLonewolf:
Following the discussion in ZeLonewolf/openstreetmap-americana#620, this PR proposes to add a new class, "pond" to the water layer, which allows for a style to de-emphasize minor bodies of water (tagged `water=pond,basin,wastewater`) when compared to more significant objects tagged reservoir/lake.  This PR retains the default behavior of classifying a water body as a "lake" if more specific tagging is not added.

image


TomPohys:
Thanks, for some use cases it makes sense. Could you please also add `pond`, `basin`, and `wastewater` into `mapping.yaml` to be more clear that these values will be used as `class`. All of them should already have `natural=water` (99.79%, 99.73%, 98.14%), so, most of them are already in the database. It is just to update the [mapping_diagram.png](https://github.com/openmaptiles/openmaptiles/blob/master/layers/water/mapping_diagram.png) for clearness.

Thank you!

TomPohys (approved):

TomPohys:
Great! Thank you!

PR #1457, opened 2022-12-16
Revamp water label display logic

### LINKED ISSUE #1468: Missing labels for oceans and seas
`water_name` point labels are missing for the following cases:
* `name=` is empty (but other `name:xx` tags exists), for example, [Atlantic Ocean](https://www.openstreetmap.org/node/305640306)
* The ocean or sea is mapped as an area, for example, [Gulf of Mexico](https://www.openstreetmap.org/way/808726001)

As a result, obvious ocean/sea labels that should be present are missing:

image

ZeLonewolf:
Unblocks ZeLonewolf/openstreetmap-americana#620
Fixes #1468

This PR fixes several issues that currently exist in the `water_name` layer:

1. Labels are not shown for some features that are large enough to have a label
2. The label logic does not consider latitude (features of the same area size get larger in pixels as they approach the north and south poles because of projection)
3. There is no way for a style to know the relative size of the water feature, which would allow the style author to choose an appropriate font size for the label.
4. Ocean/sea water features (e.g. Hudson / James Bay, Gulf of Mexico) don't have zoom-appropriate labels if they're mapped as areas rather than as points.
5. Bay and strait names (`natural=bay` and `natural=strait`) are not included at all. 
6. Large seas modeled as areas (e.g. Caspian Sea) render as `class=lake` rather than `class=sea`

# Approach

This PR takes the current `area` column in `osm_water_point_view` and adds additional computation to determine the absolute size of a feature at a particular zoom level, denominated as a percentage of a tile's area.  This value is then exposed in the tiles as `tile_area`, such that a value of 1.0 means "the same area as one tile" (in other words, approximately the same as 256x256 pixels at the current zoom). By exposing this value numerically, it allows styles to use size-dependent labels, with smaller labels for smaller features and vice versa.

Additionally:
* At zooms 9-13, water labels will now be rendered for water features that are at least 0.25 tiles in size.  This means that water features that are the same size as a 128x128 pixel square will get a point label.
* Labels for large water features are now shown at the lowest zoom at which it's appropriate to show a label.

This PR maintains the current behavior in that a water area name modeled as a point is only generated if there is not a corresponding lake line for that particular object.

# Samples

Here is Worden Pond, in Rhode Island, at zoom 12 ([test location](https://zelonewolf.github.io/openstreetmap-americana/#map=12/41.44593/-71.58805)):

image

This waterbody is clearly large enough that it could reasonably support a label.

Here is the same feature rendered in the tiles with this PR:

image

The `tile_area` tag in the tile data has a value of 0.43, which indicates that this feature is 43% of a tile in size at zoom 12.
Increasing to zoom 13, the tile_size value quadruples. At this zoom level, the size of this waterbody is greater than one tile:

image

Here is an example of a large water feature rendering (Hudson Bay in Canada), which renders at z3 but is dropped at z2:
image

Example of a bay rendering (Jamaica Bay in New York State):
image


ZeLonewolf:
Looking at the tile size increases, I'm thinking we don't necessarily need to expose the tile_area attribute in the output tile.

ZeLonewolf:
Marking draft in order to see if we can do a better job with `natural=bay` objects

ZeLonewolf:
After some experimentation, I updated this PR to limit rendering of `natural=bay` point labels to z13. Additional work is needed to improve `bay` display logic using some kind of rank logic, but at least this gets bays in the tiles without overly cluttering the tiles.

TomPohys:
Awesome! Thank you! Without adding `tile_area` it is much cleaner!

TomPohys (approved):

PR #1465, opened 2022-12-23
Add GB primary/secondary routes

### LINKED ISSUE #1462: Override UK pseudo-network values when highway classification differs
Based on discussion in ZeLonewolf/openstreetmap-americana#106, there are occasions where UK road classifications diverge from their underlying highway classification.  For example, an "A" road is usually trunk, but there are cases where they've been downgraded to primary because of a motorway bypass.  In this case, these sections should be encoded with a "primary" network value despite having an "A" designation.

Screen shot below shows an "A" route on a primary road:
image

CC/@trigpoint

ZeLonewolf:
Fixes #1462
Unblocks ZeLonewolf/openstreetmap-americana#106

This PR adds an additional pseudo-network for UK non-primary [routes](https://en.wikipedia.org/wiki/Great_Britain_road_numbering_scheme), which are signed with black on white signage with "A" and "B" prefixes. This scheme is described on the [OSM wiki](https://wiki.openstreetmap.org/wiki/Roads_in_the_United_Kingdom). Adding this third psuedo network is needed for generating shields for UK white/black signage for non-primary routes. 

Example of black-on-white non-primary routes:
image

This PR applies the tagging scheme described above, and adds `highway=primary` and `highway=secondary` into the pseudo-networking mapping scheme, as well as allows for "B" roads in the regex that defines which routes can be included.

Additionally, this fixes a minor bug under which single-digit refs would have been included.  Single-digit refs don't exist in the UK, because the numbering scheme is always prefixed by a letter, e.g. "A4".

I also included documentation updates and fixed the typo where a view was named "bg" instead of "gb".

Test screenshots from Wales, UK:

image

image

image


TomPohys (approved):

TomPohys:
Great! Thank you very much!

PR #1467, opened 2022-12-27
Add other river-like water area features to class=river

ZeLonewolf:
Currently, river areas (`natural=water` + `water=river`) are rendered in the tiles as `class=river`.  However, canal, stream, ditch, and drain water areas are rendered as `class=lake`.  Since these types of objects are all flowing water, they should be grouped together into `class=river`, which allows for styles that render flowing water differently to do so on a consistent basis.

As an example, the following object represents the water-covered area of a stream:
https://www.openstreetmap.org/way/940880798

It is presently rendered `class=lake`:
image

After this change, it is rendered `class=river`:
image


TomPohys (approved):

TomPohys:
Thank you! Yes, it makes sense!

layers/water/mapping.yaml
```diff
@@ -104,4 +104,9 @@ tables:
       - dock
       water:
       - river
+      - stream
+      - canal
+      - ditch
+      - drain
+      - lake
```
TomPohys, line 111: Is the `lake` intentional? It is not mentioned anywhere else. Thanks for explaining.

ZeLonewolf: You're right, this isn't actually necessary here because 'lake' is the default fallback case in water_class.  I will remove it and save a few bytes of DB space.

ZeLonewolf: All set, removed and updated devdocs

TomPohys: Thanks a lot! A few bytes in DB really do not matter :-D It was just a little bit confusing.

PR #1475, opened 2023-01-07
Render lake lines for big lakes

ZeLonewolf:
This PR reduces the minimum lake line zoom to z3, allowing the largest lakes to show a renderable label. The lakeline function is restricted by the LineLabel() function, which limits lakelines to only those that are big enough to render a text label.  Therefore, further restricting lakelines by zoom is not necessary.

Lake Baikal, z3:
image

Lake Taymyr, z4:
image


ZeLonewolf:
Sample in osm-americana (obviously we need to do a bit more work on the styling!):

image

TomPohys (approved):

TomPohys:
Great! Looks good!

PR #1479, opened 2023-01-18
Render POIs for large universities at low zoom

ZeLonewolf:
Currently, all POIs are rendered at z14, except for railway stations and ferry terminals, which are rendered at z12-13. However, for huge POIs, such as large universities, it makes sense to render them at lower zooms.

This PR adapts the code used in #1457 to allow POIs to render up to z10 for very large features. I set the threshold at a very conservative value of 10% of a tile. Thus, it will render if an area POI covers at least 10% of a tile at a given zoom.

Additionally, this consolidates a double UNION ALL on the same table with a WHERE/CASE combination, which is simpler.

## Samples

[Michigan State University](https://www.openstreetmap.org/relation/13708038)
image

[Universidad EARTH](https://www.openstreetmap.org/way/321376631), Costa Rica
image


TomPohys:
Thanks, but I would rather avoid any computation in the `layer_poi` function. Also, what other than `university` POIs will pop up from z10? Is not better to set the actual area threshold?

ZeLonewolf:
Thanks for the feedback - I've modified the code to short-circuit on `zoom_level` and restricted the PR to university/college, which brings performance to negligible differences from the master branch (I'm seeing around 1% differences in runtimes for a render of Costa Rica).  We can revisit other POI types as we identify them.  I've not done a deeper investigation of what POIs might show up at z13 with this algorithm so that may be premature.

> Is not better to set the actual area threshold than to calculate the percentage of area in one tile?

I'm investigating this question and will report back.

ZeLonewolf:
The reason I used a cosine function is because WebMercator projection distorts the sizes of objects the closer you get to the poles.  In Milan, objects are 41% larger, double in Stockholm, and 177% larger in Murmansk.

For example, below is a zoom 17 image tile taken from Bing aerial imagery of the port in Longyearbyen (380% distortion). The long rectangles are 40-foot (2 TEU) shipping containers, while the short rectangles are 20-foot shipping containers:
image

For comparison, I grabbed a Bing tile from the port in Bata, Equatorial Guinea, on the equator, also showing 20- and 40-foot shipping containers:
image

This is why it's not possible to just set a fixed value for area comparisons, because of the distortion that happens at various latitudes.

If the cosine computation is too expensive, we could probably find some kind of simplified formula, but hopefully the restriction by POI type and short-circuit logic by zoom level puts this within acceptable performance windows.

ZeLonewolf:
Another fun comparison:

Soccer field in Murmansk (z16): https://www.openstreetmap.org/#map=17/68.96661/33.07903
Soccer field in Equatorial Guinea (z16): https://www.openstreetmap.org/#map=16/1.8181/9.7884

ZeLonewolf:
Profiled - Costa Rica to z14:

master
```
real	17m16.982s
user	2m17.479s
sys	0m33.893s
```

PR
```
real	17m23.782s
user	2m18.566s
sys	0m32.925s
```

TomPohys:
Thanks for the examples! It looks good!

TomPohys (approved):

PR #1483, opened 2023-01-20
Consider boundary_type=maritime to be maritime=1

### LINKED ISSUE #1482: Boundary maritime field misses a common OSM encoding
Some admin4-level boundaries are missing the `maritime = 1` value, even though they are marked as maritime within OSM. It appears that some boundaries are marked within OSM with the tag `maritime=yes` and others with the tag `boundary_type=maritime`.

The issue is clearly visible [at this link](https://www.maptiler.com/maps/#style=basic-v2-light&mode=3d&position=7.06/31.87/-82.01), where the maritime borders of Florida and South Carolina appear while Georgia's do not.

maptiler-light

- [Example of `boundary_type=maritime`](https://www.openstreetmap.org/way/80794145#map=11/30.7114/-81.4032)
- [Example of `maritime=yes` (and `boundary_type=maritime`)](https://www.openstreetmap.org/way/226290304#map=11/30.8518/-81.3524)

I am unsure if this should be addressed in OSM or within this project, but I have developed a fix for the issue within this project by considering both types of tags to be `maritime = 1`. Will open a PR shortly.

Thanks!

kevinschaul:
Some admin4-level boundaries are missing the maritime = 1 value, even though they are marked as maritime within OSM. It appears that some boundaries are marked within OSM with the tag maritime=yes and others with the tag boundary_type=maritime.

This PR considers both tags to be valid techniques of marking maritime borders.

Fixes #1482

I've created a [test style](https://gist.github.com/kevinschaul/ad8c9502c42ccd999323c9c854905e11) that can be used with Maputnik to view the behavior. Screenshots follow.

Teal lines are `maritime = 1`, purple lines are not.

**Current behavior**
maritime-current

**Proposed behavior**
maritime-proposed

### DIFF
--- a/layers/boundary/boundary.sql
+++ b/layers/boundary/boundary.sql
@@ -25,7 +25,8 @@ FROM (
                  OR BOOL_OR(disputed_by <> '') AS disputed,
              NULLIF(name, '') AS name,
              NULLIF(claimed_by, '') AS claimed_by,
-             BOOL_OR(maritime) AS maritime
+             BOOL_OR(maritime)
+                OR BOOL_OR(boundary_type = 'maritime') AS maritime
       FROM osm_border_linestring
       WHERE admin_level BETWEEN 3 AND 10
             AND type = 1 -- ways only


ZeLonewolf:
What would be a good extract to test this behavior on?  South Carolina?

kevinschaul:
Florida should show it. South Carolina appears to have `maritime=yes` in OSM now, but Florida does not.

ZeLonewolf:
I'm looking at [taginfo](https://taginfo.openstreetmap.org/tags/boundary_type=maritime), and it seems that there's only 205 global usages of `boundary_type=maritime` versus 14,600 `maritime=yes`.  Perhaps it would be more sensible to add `maritime=yes` to those 205 objects instead?

kevinschaul:
That seems like a better approach. I will work on that submission now and close this PR. Thanks!

PR #1485, opened 2023-01-24
[Bugfix] Set paved attribute for paved tracks

ZeLonewolf:
This PR sets `surface=paved` for `highway=track` + `tracktype=grade1` if no other value of `surface` is set.  Roads tagged in this way are [generally paved](https://wiki.openstreetmap.org/wiki/Key:tracktype) and therefore encoding them as a paved road is appropriate.

Render samples for a road marked as a [paved track road](https://www.openstreetmap.org/way/1126554999):

image

Compare to [current baseline](https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/v3/#15.84/41.618992/-71.419593):

image


TomPohys (approved):

TomPohys:
Looks good! Thanks a lot!

PR #1486, opened 2023-01-27
Render long-distance ferries at longer zooms

ZeLonewolf:
Unblocks ZeLonewolf/openstreetmap-americana#720

In some areas of the world, ship transportation is the main means of long-distance travel.  For example, in Alaska, the [Alaska Marine Highway System](https://dot.alaska.gov/amhs/route.shtml) is the principal means of transportation through the Alaska Panhandle and down the Aleutian island chain.  The AMHS carries the same importance that the Interstate Highway System has on land, and these ferry routes are exceptionally long distance.

**Alaska Marine Highway System**
image

Currently, ferries do not begin to render until zoom 11, which is [inadequate](https://zelonewolf.github.io/openstreetmap-americana/#map=11/54.8289/-131.4887) for such long routes.

**Lowest-zoom render of AMHS routes**
image

### Solution

This PR adds rendering for ferries down to zoom 4, which is the zoom at which the most important highways are rendered. Lower zoom (<z11) ferries are filtered so that only ferries that are of a significant length on the screen are shown.

Since ferry lines are drawn with low fidelity (few nodes), they tend to add very little size to the tiles compared to other features.

### Data Render Samples

Rendered tile server (Alaska): https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/alaska-low-zoom/#4/56.59/-138.6

AMHS at zoom 4:
image

Detail added at zoom 5:
image

Detail at z10:
image

Styled tiles (OSM Americana). Clearly we could use a bit of work on the cartography but you get the idea:
image


TomPohys:
Thanks for the PR. Could you please also adjust the [style](https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/style.json#L8)? Would be worth it to merge the lines e.g. by [`network`](https://www.openstreetmap.org/way/232411464)? In the e.g. [Mediterranean sea](https://www.openstreetmap.org/way/999670971), there is no network but ferries are also long.

ZeLonewolf:
There is a nice implementation of styling ferries at low zoom in ZeLonewolf/openstreetmap-americana#775 that I'm planning to pull into the style here.

ZeLonewolf:
Style update in 872cd71. I changed the cartography to better match how it looks in the standard tile layer:

Middle zoom:
image

Low zoom:
image

ZeLonewolf:
I'm experimenting with merging the parallel linestrings in https://github.com/ZeLonewolf/openmaptiles/tree/zlw-coalesce-ferries, but it's really complicated and I'd like to tackle that as a separate PR so we can examine the performance/size/updates tradeoffs separately.

TomPohys (approved):

TomPohys:
Thanks! Looks good.

PR #1488, opened 2023-01-29
Add housename

rinigus:
Adds import of addr:housename as a separate layer. It is reusing the code available for housenumber with just number replaced by name.

### DIFF
--- a/layers/housename/README.md
+++ b/layers/housename/README.md
@@ -1,0 +1,10 @@
+## housename
+
+### Docs
+Read the layer documentation at **http://openmaptiles.org/schema#housename**
+
+### Mapping Diagram
+![Mapping diagram for housename](mapping_diagram.png?raw=true)
+
+### ETL diagram
+![ETL diagram for housename](etl_diagram.png?raw=true)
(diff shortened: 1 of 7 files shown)


TomPohys:
Thanks for the PR. I appreciate your effort. But, from [osm wiki ](https://wiki.openstreetmap.org/wiki/Key:addr:*) the `housename` is used, when there is no `housenumber` for the address. So basically, this should be part of `housenumber` layer. 

There should be also logic, when there is just `housename` and no `housenumber` then the `housename` should be in `housenumber` attribute, e.g. in https://www.openstreetmap.org/way/350463782. 

On many occasions the `addr:housename` is mistagged, because it is the same as the `name` or it is the name of the restaurant which is in the `poi` layer (e.g. https://www.openstreetmap.org/node/822499939). So `addr:housename` should be only for the buildings and where it will be part of the post address (local address) for a mail carrier. This will be tricky to unify everywhere.

rinigus:
@TomPohys, thanks for looking into it and the feedback. I think I will close PR as starting to implement the logic of selection housename layer instead of housenumber when empty is a bit too much for me right now.

Issue #1296, opened 2021-11-09
Adjust `protection_title` tag in `park` layer


TomPohys:
With PR #773 was added a lot of parks. The main issue is that there are now [2264 classes](https://taginfo.openstreetmap.org/keys/protection_title#values) filled with `protection_title` value https://github.com/openmaptiles/openmaptiles/blob/ebeafc65fad2352662b9de2a07cfafcebca23082/layers/park/park.sql#L30 

Then there are polygons in ocean e.g. with `class` 
 - `área_marinha_protegida_para_a_gestão_de_habitats_ou_espécies`
 -  `área_marinha_protegida_para_a_gestão_de_recursos`
 - `natura2000`
 - `natura_2000-gebied` 
 - ...
 
![image](https://user-images.githubusercontent.com/5182210/140954896-6442ac9e-9e44-4dc5-9ff6-f090eb70df01.png)

there should introduce some filters for parks for example
 - explicit list of `protection_title` areas
 - unification of `class` value for correct styling
 - use `protect_class` for filtering (https://wiki.openstreetmap.org/wiki/Key:protect_class)

More suggestions are welcome.

ZeLonewolf:
Thanks for bringing this up!  Parks and protected areas is something I think about a lot in OSM.

I would oppose using `protect_class` for any sort of primary class logic.  The numeric classes are poorly defined ("loosely based on IUCN categories") and for the stranger (and rarely-used numbers), completely made up.  There are only a rare handful of exceptions for objects tagged `boundary=protected_area` where you wouldn't want to treat them as any other nature reserve for styling.  I think that openstreetmap-carto makes a mistake in limiting protected area rendering to specific `protect_class` values.

The `class` value should ultimately be linked to some rationale for styling.  The only categories I think that renderers really care about for styling differences are national parks, nature reserves (protected areas), indigenous lands, ordinary parks, and recreation grounds.  I would favor something vastly simpler, like this:
`class=national_park`
* Any `boundary=national_park`
`class=protected_area`
* Any `boundary=protected_area`
* Any `leisure=nature_reserve`
`class=aboriginal_lands`
* Any `boundary=aboriginal_lands`
`class=park`
* Any `leisure=park`
* Any `landuse=village_green`
`class=recreation`
* Any `landuse=recreation_ground`

[...]

PR #1489, opened 2023-01-30
Add indigenous land boundaries

ZeLonewolf:
This PR adds support for `boundary=aboriginal_lands` by adding it to the `park` layer. While these lands are certainly not "parks", they have similar treatment from a technology perspective and can benefit from the existing processing chain established in that layer. I set all of these objects to `class=aboriginal_lands` in the tiles, including at the lowest zoom in order to separate it from the protected area merging implemented in #1160. In order to distinguish these from general parks, I expose the `class` attribute for these objects at z4 and also ensure that the z4 generalization is performed separately for protected areas versus aboriginal lands.

This unblocks #ZeLonewolf/openstreetmap-americana#105, which describes why having indigenous land boundaries is an important general feature on the map. This is also consistent with my suggested implementation in https://github.com/openmaptiles/openmaptiles/issues/1296#issuecomment-967749403.

Sample data-view renderings from Arizona, USA:

image

image


ZeLonewolf:
Note that another option, would be to put this in the `landuse` layer, but extend that layer to include lower zooms. For example, `landuse=military` can be similarly large and could benefit from being brought up to lower zooms.

TomPohys:
Thanks! It makes sense.

TomPohys (approved):

PR #1491, opened 2023-02-07
Simplify park classes

### LINKED ISSUE #1296: Adjust `protection_title` tag in `park` layer

ZeLonewolf:
Fixes #1296
~~Blocked by #1604~~

This PR simplifies the assignment of `park` layer  classes from 2,000+ to a shorter list.  With this PR, the possible classes are:
- `national_park`
- `marine`
- `nature_reserve`
- `recreation`
- `historic`
- `conservation`
- `wildlife_refuge`
- `wilderness_preserve`
- `sustainable`
- `community`
- `protected_area`
- the raw value in the [`protected_area`](https://taginfo.openstreetmap.org/keys/protected_area) tag

If all else fails, we still fall back to `protection_title` as a fallback, if there is no other protected area tagging, but that should be a rare case.

The reason for the additional `recreation` and `historic` classes are that areas are occasionally tagged a `boundary=protected_area` but also tagged with historic or recreation_ground tagging, or are explicitly tagged with `protected_area=recreation`. In those cases, it's more appropriate to classify those areas as a sep
arate type of object. For completeness, I also explicitly pulled in 3 types of historic that are sometimes conflated with `boundary=protected_area`: archaeological sites, historic districts, and historic battlefields. This adds a small number of objects to this layer. An alternate option would be to explicitly remove objects in the two latter classes from this layer and leave only national parks and nature reserves as the two classes. However, I think these four classes logically go together.

image

image


dschep:
This feels like an oversimplification to me. For example, in [OpenTrailStash](https://gitlab.com/trailstash/openstyle) I rely on `class=wilderness`, `class=national_wildlife_refuge`, and `class=wildlife_management_area` to style wilderness areas,  Wildlife Refuges, & WMAs differently from parks:

image

ZeLonewolf:
@dschep can you describe the category of lands that you would want to render differently? That scheme would miss quite a few examples in the US based on [US Public Lands](https://wiki.openstreetmap.org/wiki/United_States/Public_lands). It's not hard to create a class to cover wilderness-type areas but the challenge is in the definition.  Perhaps we could pull in `protect_class=1a,1b,4` to cover these. Since the current `class` values come from a free-text field, we'd want something that works globally.

dschep:
I currently use the three values I mentioned, but I could see wanting more. What about retaining the original values as `subclass`?

ZeLonewolf:
Retaining the original values in `subclass` would of course not be difficult, though I would look to @TomPohys for what the original intent was for simplifying the values. It does not seem that this simplification has any substantive impact on tile sizes.  The goal is that the attributes are more _useful_ to style authors.

The problem with the original `class` values of `wilderness`, `national_wildlife_refuge`, `wildlife_management_area` is that they basically only work in the United States, or at least definitely only in English-speaking places. So what I'm hoping to gain out of this discussion is a more holistic understanding of the "type of area" that you see as valuable to render differently, and then we can figure out what sets of tags people are generally using to describe them such that you (or any other style author) isn't hunting down what a wildlife area is named in hundreds of different countries.

Now, if it's just that you're making a US-specific map and you're specifically interested in those categories, that's a somewhat different use case and I could see why a `subclass` scheme would work.

dschep:
Yeah, in general it seems like something derived from `protect_class` would be better international use. I think it goes without saying that I consider the presence of the original values to be... _useful_ :wink: (seeing as I make use of it). This PR's logic is trivial (unlike some other cartographicdecisions) to implement in a client-side style.

But that may fall generally into the bucket of "make cartographic decisions in tiling" that OMT may follow (not too sure about the specific goals of OMT). I'm also making a fair bit of other cartographic decisions in the tiles that would probably best be made in the tiling (like a more nuanced implementation using protect class). I should just setup my own tiling, but that takes time&infrastructure I don't have.

Totally understand if this PR is in the spirit of OMT's goals and my use of OMT isn't (`surface` simplification already is more than I personally want. another reason I should do my own tiling)

TomPohys:
Hi, the original thought was to reduce number of classes and if it possible some type of grouping logic (e.g. as it is in `poi` layer). Now, there is almost no chance to style park globaly based on e.g. `national_park`. It could be styled for local perspective (e.g. `natura_2000` vs `natura_2000-gebied` or https://www.openstreetmap.org/relation/5421660) or style single park.

The simplification could reduce number of parks, but it should not reduce it too much (e.g. parks in ocean - https://www.openstreetmap.org/relation/7512408; be in `marine` group? ). Maybe, for better grouping, I could brings the logic from `transportation` or `poi` layer 
https://github.com/openmaptiles/openmaptiles/blob/f3dbb0a83959c05b6a8a2757f170b7f2696501f0/layers/transportation/class.sql#L14-L20
https://github.com/openmaptiles/openmaptiles/blob/f3dbb0a83959c05b6a8a2757f170b7f2696501f0/layers/transportation/transportation.yaml#L30-L38

I think, that add additional field e.g. `protection_class` or `subclass` will not cause massive tile size raise.

ZeLonewolf:
Moving to draft to merge changes with #1489 and make modifications consistent with feedback above.

TomPohys (approved):
Awesome, thanks a lot, and sorry for such a delay!

Issue #1422, opened 2022-09-30
The "make" command returns "docker-compose" not found even when it is installed


douglasdcm:
Hi, how are you? It is the first time I'm using the tool. I'm facing a problem while running the "make" command. It seems the docker-compose is not found. However, I installed it and the current version is  v.2.10.2. Notice the command has been changed in docker documentation from "docker-compose" to "docker compose" (without dash). I'd say it is breaking the "make" process. Check the log bellow where I print the "make" and the "docker compose" results. The ./quickstart.sh is suffering of the same problem.

```[root@localhost openmaptiles]# make
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
/bin/bash: .env: No such file or directory
make: Warning: File `Makefile' has modification time 22940680 s in the future
docker-compose run --rm --user=0:0 openmaptiles-tools bash -c \
	'generate-tm2source openmaptiles.yaml > build/openmaptiles.tm2source/data.yml'
/bin/bash: docker-compose: command not found
make: *** [build/openmaptiles.tm2source/data.yml] Error 127
[root@localhost openmaptiles]# docker compose version
Docker Compose version v2.10.2
[root@localhost openmaptiles]# 
```

https://docs.docker.com/compose/install/linux/

StephenAtty:
I think that the docker docs have changed and now compose is installed as a plugin rather than stand alone.

https://docs.docker.com/compose/install/other/#install-compose-standalone

But make is still using the "dash compose syntax"

On my linux box

 docker-compose  -v
docker-compose version 1.29.2, build 5becea4c

docker compose version
Docker Compose version v2.10.2

TomPohys:
Hi, thanks for pointing this out. PR for adjusting documentation is very welcome! Thanks

sukamare:
For those who are using the docker compose plugin this link might help:

[...]

PR #1497, opened 2023-02-20
Support docker compose V2 syntax `docker compose`

wcedmisten:
Currently the Makefile and `quickstart.sh` expect the [deprecated docker compose V1](https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command) syntax, `docker-compose`.
The new V2 syntax uses `docker compose` instead.

When using the V2 version, an error `docker-compose: command not found` occurs.

This is mentioned here: https://github.com/openmaptiles/openmaptiles/issues/1422 with suggestion to update the docs.

However, I think it would be preferable to support docker compose V2 syntax directly in the Makefile and `quickstart.sh`.

Let me know if you have any questions! I'm happy to explore a different approach to fixing this, I'm no make/bash expert, so if there's a better way, I'd be happy to adopt it.


wcedmisten:
@TomPohys @ZeLonewolf do you all have any thoughts on this? Happy to create an Issue if discussion there is preferred.

ZeLonewolf:
I'm a complete amateur around docker, so I don't have an opinion.  I would defer more to @nyurik.

nyurik:
i don't see any reasons not to update to v2.  Overall, docker compose is a pain to use in many cases, and hard to scale it to use in k8s, but it does offer a quick start.

TomPohys (approved):

TomPohys:
Thanks! It looks good!

Issue #1498, opened 2023-02-23
New postgres index causing failures in imposm - only master branch


darren1713:
## Your Environment
Ubuntu 22
imposm version 0.11.1
Postgresql 15.2-1.pgdg22.04+1
Postgis 3.3.2

## Description
There is a new index created in #1361 , specifically:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS osm_route_member_network_ref_idx ON osm_route_member ("member", "network", "ref");
```

Which is causing imposm to fail with:

```
[2023-02-20T00:15:20Z] 9:55:51 [info] Importing #3616 including changes till 2023-02-19 21:21:39 +0000 UTC (2h53m41s behind)
[2023-02-20T00:15:20Z] 9:55:51 [step] Starting: Importing #3616
[2023-02-20T00:15:20Z] 9:55:51 [step] Starting: Processing /import/000/003/616.osc.gz
[2023-02-20T00:15:20Z] 9:55:51 [step] Starting: Parsing changes, updating cache and removing elements
[2023-02-20T00:15:22Z] 9:55:53 [step] Finished: Parsing changes, updating cache and removing elements in 2.469450234s
[2023-02-20T00:15:22Z] 9:55:53 [step] Starting: Importing added/modified elements
[2023-02-20T00:15:22Z] 9:55:53 [progress]     2s C:       0/s (28316) N:       0/s (944) W:       0/s (5409) R:      0/s (53)
[...]

PR #1501, opened 2023-02-28
Convert osm_route_member unique index to trigger

### LINKED ISSUE #1496: Integrity CI fails with indecipherable error message
The integrity CI currently fails in all case.
The error messages in the console are not helpful in troubleshooting the problem.

I've also opened #1495 in order to provide a script matching what this CI does in order to help with present and future troubleshooting.

ZeLonewolf:
Fixes #1496, #1498

See also omniscale/imposm3#288

This PR changes the three-column unique index to an on-insert trigger. This should fix the issues we're having with CI failures and still achieve the behavior of coalescing duplicate routes.

Test area (on current release, still showing the duplicate route issue):
[image](https://zelonewolf.github.io/openstreetmap-americana/#map=11/39.2478/-82.9378)
[image](https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/v3/#11/39.2478/-82.9378)

Fix with this PR (localhost links):
[image](http://localhost:1776/#map=11/39.2478/-82.9378)
[image](http://localhost:8080/data/openmaptiles/#11/39.2478/-82.9378)


pnorman:
If you import a diff which results in two conflicting rows from different relations, what finally ends up in the table will be determined by the order the changes are processed in, and may not be deterministic. To me, this indicates you are probably doing something wrong by expecting uniqueness where there's no guarantee.

ZeLonewolf:
I am not expecting uniqueness, I'm attempting to create it. This condition occurs when there are two route relations in opposite directions with the same network and ref.

ZeLonewolf:
Currently working on a more robust fix...

layers/transportation/update_route_member.sql
```diff
@@ -122,73 +122,59 @@ BEGIN
...
+  network,
+  ref,
+  DENSE_RANK() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index,
+  rank
+FROM (
+  SELECT DISTINCT
+    member,
+    network_type,
+    network,
+    ref,
```
ZeLonewolf, line 155: I moved the concurrency_index calculation into an intermediate materialized view to to separate the de-duplication capability from both DENSE_RANK() and from imposm updates.

layers/transportation/update_route_member.sql
```diff
@@ -122,73 +122,59 @@ BEGIN
...
+    REFRESH MATERIALIZED VIEW transportation_route_member_coalesced;
 END;
 $$ LANGUAGE plpgsql;
 
-CREATE INDEX IF NOT EXISTS osm_route_member_network_idx ON osm_route_member ("network", "ref");
-CREATE INDEX IF NOT EXISTS osm_route_member_member_idx ON osm_route_member ("member");
-CREATE INDEX IF NOT EXISTS osm_route_member_name_idx ON osm_route_member ("name");
-CREATE INDEX IF NOT EXISTS osm_route_member_ref_idx ON osm_route_member ("ref");
-
-CREATE INDEX IF NOT EXISTS osm_route_member_network_type_idx ON osm_route_member ("network_type");
```
ZeLonewolf, line 150: Investigation of the pg_stat tables indicates that these indexes aren't actually used in this scheme.

ZeLonewolf:
Functionality still working:
[image](http://localhost:8080/data/openmaptiles/#14/39.23203/-82.98333)

TomPohys (approved):

TomPohys:
Thank you very much! Looks good!

PR #1444, opened 2022-11-22
Bugfix: z4/z5 transportation layer progression

ZeLonewolf:
This PR improves the behavior introduced in #1440, which inadvertently included all motorways at z4, which was unintended.  Instead, this PR provides an appropriate progression from z4 to z5 with the highest national-importance highways being rendered at z4 (including where they degrade to trunk, as in Canada), and all motorways being added to the mix at z5.

|    | Current                                                             | This PR                                                             |
|----|---------------------------------------------------------------------|---------------------------------------------------------------------|
| z4 | All `highway=motorway` and "National importance" trunk + motorway   | "National importance" trunk + motorway                              |
| Z5 | All  `highway=motorway`  and "National importance" trunk + motorway | All  `highway=motorway`  and "National importance" trunk + motorway |

Zoom 4 North America:

image

Zoom 5 North America:

image

PR #1505, opened 2023-03-19
Local Test Fixes

benedikt-brandtner-bikemap:
This PR fixes local tests affected by the changes introduced in #1440 #1444 #1361 #1501

### DIFF
--- a/tests/import/500_import-highway.osm
+++ b/tests/import/500_import-highway.osm
@@ -165,4 +165,14 @@
     <tag k="ref" v="95"/>
     <tag k="description" v="I-95 Southbound"/>
   </relation>
+
+  <!-- Zoom-Level 5 Trunk National Network -->
+  <relation id="517" visible="true" timestamp="2019-01-01T00:00:00Z" version="1" changeset="1" user="u" uid="1">
+    <member type="way" ref="5001" role=""/>
+    <tag k="type" v="route"/>
+    <tag k="route" v="road"/>
+    <tag k="network" v="US:I"/>
+    <tag k="ref" v="94"/>
+    <tag k="description" v="I-94"/>
+  </relation>
 </osm>
--- a/tests/test-post-import.sql
+++ b/tests/test-post-import.sql
@@ -72,19 +72,24 @@ BEGIN
   -- Test 500
 
   -- Verify that road classifications show up at the right zoom level
-  SELECT COUNT(*) INTO cnt FROM osm_transportation_merge_linestring_gen_z4 WHERE highway='motorway';
+  SELECT COUNT(*) INTO cnt FROM osm_transportation_merge_linestring_gen_z4 WHERE osm_national_network(network);
   IF cnt < 1 THEN
-    INSERT INTO omt_test_failures VALUES(500, 'import', 'osm_transportation_linestring z4 motorway count expected >=1, got ' || cnt);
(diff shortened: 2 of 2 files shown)


layers/transportation/update_transportation_merge.sql
```diff
@@ -483,7 +483,7 @@ BEGIN
         WHERE
             (highway = 'motorway'
             OR construction = 'motorway'
-            -- Allow trunk roads that are part of a nation's most important route network to show at z4
+            -- Allow trunk roads that are part of a nation's most important route network to show at z5
             OR (highway = 'trunk' AND osm_national_network(network))
```
ZeLonewolf, line 487: Is this `highway = 'trunk'` necessary given the change at line 525?

ZeLonewolf:
This PR reverses #1444 which corrected a _bug_ introduced in #1440. It was never anyone's intention to show all motorways at z4, only the most important ones. So far "most important" has only been defined for the US and Canada, but the intent is that similar top-level networks in other countries would be similarly defined for visibility at z4. Unless there is a specific client demand for z4 motorways in general, I don't think it makes sense to render them to the tiles as a universal rule. Of course the client can always filter them out at style time.

benedikt-brandtner-bikemap:
Hey, oh sorry so the intention of #1444 was to remove all motorways from `osm_transportation_merge_linestring_gen_z4`?

Because in the currently released version 3.14 this table contains all motorways independent of their network tags.
Then #1440 added trunks with the mentioned network tags and #1444 removed those trunks again but also removed all motorways and i thought that was not intentional.
But if it was then of course i'll revert https://github.com/openmaptiles/openmaptiles/pull/1505/commits/83589c25f587143e98e69fa91f86646a667eea6e

layers/transportation/update_transportation_merge.sql
benedikt-brandtner-bikemap: That is correct if we add all important networks independent of their highway/construction tags to z4 anyways we should do it here as well.

ZeLonewolf:
The intent of the transportation layer at zoom 4 is to render an unlabeled transportation network that shows the overall shape of the road network without excessive detail. Including "any motorway" at that zoom would include many motorway spurs and islands that aren't desirable to see at that zoom.

Some of the discussion in ZeLonewolf/openstreetmap-americana#565 might be helpful to understand what this lowest-zoom rendering is trying to achieve.  Also #1444 has some of the metrics, which include a 15% tile size difference at z5 (!!).

Adding test cases of course is a good idea (and would have prevented me from making the mistake in the first class!)

benedikt-brandtner-bikemap:
Alright, have removed the commit and ammended the tests

benedikt-brandtner-bikemap:
Have also removed this branch as a base for all the other PRs and rebased them ontop of master.

ZeLonewolf:
Good work.  I dropped a note in Slack asking for a maintainer to run workflows.

benedikt-brandtner-bikemap:
thanks a lot :)

TomPohys:
Thanks a lot, @benedikt-brandtner-bikemap! Could you please also edit the "explanation text" for this PR? Thanks!

benedikt-brandtner-bikemap:
Hey, have updated the title and description

TomPohys:
Thanks a lot!

Issue #1433, opened 2022-10-25
Performance Upgrades for Diff-Updates


benedikt-brandtner-bikemap:
When applying diff-updates to a planet-import once a week by merging daily OSM-Change-Files from the past week I encountered very long processing times of up to 10 days.

After going through all refresh functions I identified a few key-issues which I believe should address the general performance of diff-updates across all refresh functions with only minimal cost to import time:
1. Refactor diff-updates to tables containing merged LineStrings to be ID-Based instead of Tag-Based or even full
2. Refactor Buildings update from full to partial-diff-update
3. Refactor Poi-Point updates from full to partial-diff-updates
4. Refactor Water-Point/Lakeline updates from full-incremental to partial-diff-updates
5. Refactor Park-Polygon update from full-incremental to partial-diff-update 
6. Adding partial Indices on tables affected by the update matching the update query
7. Restrict entries in changes-tables to be unique by adding Primary Keys
8. Restrict Diff-Updates to `INSERT` and `UPDATE` operations where applicable
9. Refactor `IN` queries to `EXISTS` queries
10. Refactor CTEs in refresh-functions to temporary tables with indexes
11. Analyzing tables before and during diff-updates

I have implemented solutions to this issues except for point 2 in [this branch at our fork](https://github.com/openmaptiles/openmaptiles/compare/master...bikemap:openmaptiles:update-performance) and would like to ask if you would be interested in merging them back into the main repository.

Environment:

AWS m5.2xlarge EC2-Instance (EC2 not RDS)
- 8 Cores
- 32GB RAM

AWS GP3 Volume
- 1000GB
- 6000 IOPS
- 125 mb/s max-throughput
- EXT4 Filesystem

Operating System and Libraries:
- Amazon-Linux 2 with Linux-Kernel 5.x
- PostgreSQL 14
- PostGIS 3.3.1
- Proj 4.9.3
- GEOS 3.9.3
- GDAL 2.4.4
- Protobuf 3.14.0
- Protobuf-C 1.3.3
- LibKakasi 2.3.6
- UTF8-Proc 2.5.0
- Mapnik-German 2.5.9.1
- PGSQL-gzip 1.0.0

PostgresSQL-Config:
- jit = 'OFF'
- shared_buffers = '7424MB'
- work_mem = '23756kB'
- maintenance_work_mem = '2GB'
- effective_io_concurrency = 200
- max_worker_processes = 8
- max_parallel_maintenance_workers = 4
[...]

PR #1510, opened 2023-03-19
Update Performance poi Layer

benedikt-brandtner-bikemap:
Improved update performance of poi layer
- Refactored update_poi_point.sql to partial diff update
- Refactored IDs to be unique in poi_polygon.osm_ids
- Restricted updates to INSERT and UPDATE operations during poi_polygon.refresh
- Added analyze statements before update queries during poi_polygon.refresh

See #1433 for more detailed discussion.


TomPohys:
Thank you! Could you please rebase it?

TomPohys:
Thanks a lot!

Issue #1516, opened 2023-03-19
Remove unused indexes


ZeLonewolf:
There are a significant number of indexes in OpenMapTiles that appear to be unused.  If these are truly unused, we could save on import and update performance by removing them.  Each index makes inserting or updating data slower because the index has to be updated in addition to the table data. After running the integrity script (which performs import and update action), I found that 274 indices appear to be unused.

The SQL query I used is below and a listing of the 274 apparently unused indices is listed below.

```sql
select indexrelname from pg_stat_all_indexes where schemaname = 'public' and idx_scan = 0;
```

```
 osm_ocean_polygon_gen_z10_idx
 osm_ocean_polygon_gen_z9_idx
 osm_ocean_polygon_gen_z6_idx
 osm_ocean_polygon_gen_z8_idx
 ne_10m_ocean_gen_z5_idx
 osm_ocean_polygon_gen_z7_idx
 ne_50m_ocean_gen_z4_idx
 ne_50m_ocean_gen_z3_idx
 ne_50m_ocean_gen_z2_idx
 ne_110m_ocean_gen_z1_idx
 ne_110m_ocean_gen_z0_idx
 ne_110m_lakes_gen_z1_idx
 ne_110m_lakes_gen_z0_idx
 osm_important_waterway_linestring_gen_z11_geometry_idx
 osm_important_waterway_linestring_gen_z10_geometry_idx
 osm_important_waterway_linestring_gen_z9_geometry_idx
 osm_important_waterway_linestring_pkey
 osm_important_waterway_linestring_geometry_idx
 ne_110m_rivers_lake_centerlines_gen_z3_idx
 waterway_relation_geometry_idx
 waterway_relation_gen_z8_geometry_idx
 waterway_relation_gen_z7_geometry_idx
 waterway_relation_gen_z6_geometry_idx
 ne_10m_admin_1_states_provinces_geometry_geom_idx
 ne_10m_admin_0_boundary_lines_land_geometry_geom_idx
 ne_10m_admin_1_states_provinces_lines_geometry_geom_idx
 ne_10m_antarctic_ice_shelves_polys_geometry_geom_idx
 ne_10m_geography_marine_polys_geometry_geom_idx
 ne_10m_glaciated_areas_geometry_geom_idx
 ne_10m_lakes_geometry_geom_idx
 ne_10m_ocean_geometry_geom_idx
 ne_50m_admin_0_boundary_lines_land_geometry_geom_idx
[...]

PR #1518, opened 2023-03-20
Remove unused indexes from osm_transportation_* tables

ZeLonewolf:
This is an initial attempt to get at removing unnecessary indexes as identified in #1516.

This PR removes apparently unused indexes from the osm_transportation_* tables.

I would appreciate feedback as to why these indexes appear to be unused and whether or not they are safe to remove for performance improvement.


pnorman (changes requested):
You can't look at index usage data from only loading an extract and say the indexes aren't used when the full planet is loaded, and based on the linked issue it sounds like that's what you did.

The best you can do is identify candidates and investigate those.

In particular I'm very skeptical about removing the geom indexes from the generalized tables unless the tables are completely unused.

ZeLonewolf:
Starting over

PR #1520, opened 2023-03-23
Remove unused rank indexes in place layer

ZeLonewolf:
Based on analysis from planet builds for this layer as well as running updates via the integrity script as discussed in #1516, these three indexes are not used and can be removed.


ZeLonewolf:
My understanding of the update scripts are that they primarily use id fields in doing updates.

ZeLonewolf:
Also - conceptually - an index only on the rank column is pretty pointless - we're not pulling out data by queries on rank, we're pulling them out by geometry or updating by ID.

ZeLonewolf:
Reviewing the code, I'm confident that these rank indexes aren't solving a useful purpose.  All of the queries that use rank are filtering by geometry, which would be the index used.

benedikt-brandtner-bikemap:
Sorry was not available the past few days.

Yeah after looking at the query in more detail you are correct the index will never be since the primary-key (id) will take precedence.

Am currently also running a test-import and updates (#1516)  so maybe we could still wait until i post results from `pg_stat_all_indexes` there?

ZeLonewolf:
Thanks, additional data would be great to ensure that I'm not missing anything.

benedikt-brandtner-bikemap:
Have posted my results of the import and update run of a small region in the mentioned issue and you were correct these indexes have not been accessed during the update.

TomPohys:
Thanks a lot!

PR #1522, opened 2023-03-28
Fix copy/paste error on boundary z1 index

ZeLonewolf:
This PR fixes a copy/paste error which caused the z2 boundary layer index to be created twice and the z1 index to not be created at all.  I discovered this while investigating #1516.


TomPohys (approved):

TomPohys:
Thanks! Good catch!

Issue #1424, opened 2022-10-05
mbtiles not being created


sukamare:
I tried to run the quickstart.sh on two different server setups. Both are debian based, one using latest docker-ce and compose plugin

```
➜  openmaptiles git:(master) docker --version
Docker version 20.10.18, build b40c2f6
➜  openmaptiles git:(master) docker compose version
Docker Compose version v2.10.2
```

the other one using the current versions in debian buster-backports:

```
root@machine ~# docker --version
Docker version 20.10.17, build 100c701
root@dmachine ~# docker-compose --version
docker-compose version 1.29.2, build 5becea4c
```

In both cases, after downloading the repository and immediately running `./configure.sh` the script will run and finish without visible error message, but no albania.mbtiles file is being created. Here's the quickstart.log from the first machine:

[quickstart.log](https://github.com/openmaptiles/openmaptiles/files/9717033/quickstart.log)

Any help is much appreciated!

TomPohys:
Related to https://github.com/openmaptiles/openmaptiles/issues/1422

sukamare:
Second machine does use the standalone version of docker-compose so there should be no relation to #1422 in that case. Here's the logfile.
[quickstart2.log](https://github.com/openmaptiles/openmaptiles/files/9717180/quickstart2.log)

sukamare:
Am 5. Oktober 2022 18:29:05 MESZ schrieb Tomas Pohanka ***@***.***>:
>Sorry, I reacted to the first setup with docker compose as plugin.
>
>The second setup correctly generate 'tiles.mbtiles' as configured in '.env' file. 
>

Ok, but shouldn't the output be named after the chosen region (e.g. albania.mbtiles)?

TomPohys:
Yes, you are correct. The `quickstart.sh` should create `area.mbtiles` and in the log file, it looks like that. But after https://github.com/openmaptiles/openmaptiles/pull/1364 (@zstadler) it looks like that `area` does not propagate to https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/generate-tiles#L42 and it is preferred the value using environment variables or value in `.env`

A workaround is, that you use an environmental variable such as `MBTILES_FILE=my.mbtiles ./quickstart.sh monaco`

Maybe we can bring this behavior to `./quickstart.sh`
https://github.com/openmaptiles/openmaptiles/blob/cd0ffacc7242c9c932544c61847f9c2e65390c32/quickstart.sh#L282

[...]

PR #1523, opened 2023-03-30
Use area name for mbtiles in quickstart.sh

wcedmisten:
# Motivation

Some parts of quickstart.sh reference `${area}.mbtiles`, but the file that gets created with the default `.env` is `tiles.mbtiles`. We should either reference `$MBTILES_FILE` instead, or override the filename with `${area}.mbtiles` (which is the approach taken by this PR).

# Issue

https://github.com/openmaptiles/openmaptiles/issues/1424

# Solution

This PR implements the fix suggested by @TomPohys [here](https://github.com/openmaptiles/openmaptiles/issues/1424#issuecomment-1269604848), but I'm wondering if we should actually take this approach.

I also noticed that other parts of the repo reference a hard-coded `tiles.mbtiles`, merging this PR as is might break those examples. In this case, it may be better to just use `$MBTILES_FILE` in quickstart.sh


wcedmisten:
Sure thing! How does this look?

TomPohys (approved):

TomPohys:
Awesome! Thank you!

PR #1527, opened 2023-04-08
Add cusine subclass for food POIs

ZeLonewolf:
This PR adds select `cuisine` values as subclass values for `fast_food` and `restaurant`. This allows for custom icons for common cuisine types such as "pizza", "burger", "donut", etc.

### DIFF
--- a/layers/poi/mapping.yaml
+++ b/layers/poi/mapping.yaml
@@ -1,4 +1,3 @@
-
 # imposm3 mapping file for https://github.com/osm2vectortiles/imposm3
 # Warning: this is not the official imposm3
 
@@ -74,6 +73,15 @@ def_poi_mapping_barrier: &poi_mapping_barrier
   - sally_port
   - stile
   - toll_booth
+
+# cuisine values , see http://taginfo.openstreetmap.org/keys/cuisine#values
+def_poi_mapping_cuisine: &poi_mapping_cuisine
+  - burger
+  - coffee_shop
+  - donut;coffee_shop
+  - ice_cream
+  - pizza
+  - pizza;italian
 
 # building values , see http://taginfo.openstreetmap.org/keys/building#values
 def_poi_mapping_building: &poi_mapping_building
@@ -383,11 +391,15 @@ def_poi_fields: &poi_fields
   - name: brand
     key: brand
     type: string
+  - name: cuisine
+    key: cuisine
+    type: string
 
 def_poi_mapping: &poi_mapping
   aerialway: *poi_mapping_aerialway
   amenity: *poi_mapping_amenity
   barrier: *poi_mapping_barrier
+  cuisine: *poi_mapping_cuisine
   building: *poi_mapping_building
   highway: *poi_mapping_highway
   historic: *poi_mapping_historic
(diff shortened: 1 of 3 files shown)


layers/poi/mapping.yaml
```diff
@@ -75,6 +74,15 @@ def_poi_mapping_barrier: &poi_mapping_barrier
...
   - toll_booth
 
+# cuisine values , see http://taginfo.openstreetmap.org/keys/cuisine#values
+def_poi_mapping_cuisine: &poi_mapping_cuisine
+  - burger
+  - coffee_shop
+  - donut;coffee_shop
+  - ice_cream
+  - pizza
+  - pizza;italian
```
1ec5, line 84: The `cuisine` key is currently a mismash of orthogonal ideas about food type and ethnic/national cuisine. There seems to be some interest in [splitting it into two keys](https://wiki.openstreetmap.org/wiki/Talk:Key:cuisine#Separate_cuisine_from_food_type), but probably not enough to be relevant to OpenMapTiles yet. In the meantime, perhaps OMT could bucket the values into different feature properties. This will dramatically reduce the number of multiple-value situations, particularly in a country like Italy that’s associated with its own `cuisine` value.

layers/poi/mapping.yaml
```diff
@@ -75,6 +74,15 @@ def_poi_mapping_barrier: &poi_mapping_barrier
   - stile
   - toll_booth
 
+# cuisine values , see http://taginfo.openstreetmap.org/keys/cuisine#values
+def_poi_mapping_cuisine: &poi_mapping_cuisine
+  - burger
+  - coffee_shop
+  - donut;coffee_shop
```
1ec5, line 81: When there are multiple values, should the feature have multiple separate properties such as `cuisine_1` and `cuisine_2` (following the `route_*` precedent in #1152)? If not and the `cuisine` property ends up being set to multiple values, then 

* Maybe OMT should sort the values alphabetically, because semicolon lists are [supposedly unordered](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2500812) when it comes to `cuisine`. This way the style won’t need to check for each permutation of values in a multi-value list.
* Downstream, MapLibre will need to add support for splitting and joining expression operators, or better yet, an array filtering operator: maplibre/maplibre-gl-js#2064.

layers/poi/mapping.yaml
ZeLonewolf: These particular semi-colon-separated values seem to be the direct result of iD presets for chain restaurants, and they come in high numbers, so I'm not terribly worried about treating them as a singular tag.  Though I agree it would be a bit nicer in the output to bucket them.

1ec5: Chain restaurants are more homogeneous thanks to name-suggestion-index, but those popular values obscure the very significant value lists that come in different orders.

https://github.com/SomeoneElseOSM/SomeoneElse-style/blob/ede27ed78bd35a7e89cffd077079156346e667bf/taginfo.json#L804-L942 is an example of the kinds of gymnastics that a stylesheet has to pull if it consumes the entire list as a monolithic value in order to pick an icon based on just one of the values.

PR #1535, opened 2023-04-24
Remove LEAKPROOF attribute from functions

benedikt-brandtner-bikemap:
LEAKPROOF requires the executing user to have superuser privileges when creating the functions which would be the only statement required to be executed as a superuser.

My question would be if the performance gain is worth having to run the import as a superuser or if we could remove this again?


TomPohys:
These are the only two functions in the OMT project that are `LEAKPROOF`. As the function is also marked as `IMMUTABLE` I think, that there should not be a performance issue if the `LEAKPROOF` will be missing (and the tests are OK).

benedikt-brandtner-bikemap:
Hey, sorry have been busy but thx for the review :) 

have rebased the branch ontop of current master!

TomPohys (approved):

PR #1544, opened 2023-05-24
feat: add charging station to car POIs

cbeddow:
Icon exists already, now adding this under POIs section for `car` next to the `fuel` amenity


TomPohys:
Hi, thanks for the PR, and sorry for such delay. 

As `charging_station` could be part of the `fuel` class, do you think, that you could modify this PR and modify `poi.yaml` which will include 
```
fuel:
          subclass: ['fuel', 'charging_station`]
```
Also, and this is just a suggestion, there is a `brand` (or `operator`) tag, which could be useful for distinguishing the companies.

Please, could you look at https://github.com/openmaptiles/openmaptiles/pull/1390 as a template?

Thanks

TomPohys (approved):

PR #1545, opened 2023-05-24
add tree near garden and park, as POI leisure

cbeddow:
Tree can be shown as a single point, similar to a garden or park point POI.

Adding tree would add 1,819,499 new nodes in France to the tiles.


TomPohys:
Hi, you adding `leisure=tree` into `POI layer`, where based on [taginfo](https://taginfo.openstreetmap.org/tags/leisure=tree) there are only 5 and are probably `typos`

21M trees are in [natural=tree](https://taginfo.openstreetmap.org/tags/natural=tree) (e.g. https://www.openstreetmap.org/node/10586718270#map=19/47.37933/8.53014)

Anyway, the trees should not be part of the POI layer. The design of this schema is to be general-topographic. It would be better to add city-level and zoom 15 stuff (bench, trees) as a different layer with the behavior of https://github.com/openmaptiles/openmaptiles/pull/1003 turned off by default.

TomPohys:
We can discuss the `natural=tree` tag in a separate issue (open one if needed). Thanks!

PR #1562, opened 2023-07-07
Convert semi-colon separated house numbers to a range

### LINKED ISSUE #1558: Collapse multi-value housenumber values
For house numbers which have many values, common practice is to map them as semi-colon separated ([location](https://vtile.ourmap.us/inspect.html#19/41.7279201/-72.2071153)):

image

I propose that these be combined to a more readable string with an en dash in between such as 58-66.

ZeLonewolf:
Fixes #1558 

This PR collapses housenumber values into the form min(housenumber)-max(housenumber) for cases where housenumber is a semi-colon separated list. If the list is all numbers, the bounds are the smallest and largest numbers.  If the list includes non-numeric characters, it falls back to the first and last values in the list.

image

image

Location (localhost link):
http://localhost:8080/data/v3/#19/41.7278712/-72.2072315


layers/housenumber/housenumber_display.sql
```diff
@@ -0,0 +1,37 @@
...
+$$ LANGUAGE plpgsql;
+
+
+CREATE OR REPLACE FUNCTION display_housenumber(raw_housenumber text)
+RETURNS text AS $$
+DECLARE
+  min_number int;
+  max_number int;
+BEGIN
+  -- Check if the input string contains a semi-colon separator
```
1ec5, line 21: It’s very common to tag `addr:housenumber` with a range, as an alternative to enumerating each number in the range. The most common range separator is a hyphen, although some features use ` to ` to avoid the ambiguity in https://github.com/openmaptiles/openmaptiles/issues/1558#issuecomment-1624649822. It would make sense for the raw range separator to become an en dash too. If that sounds too easy, consider that a tag value might contain a range followed by a semicolon followed by another value.

ZeLonewolf: There is probably not a perfect processing that will work in all cases - is there logic you would suggest to handle these cases?

1ec5: If you’re comfortable with ignoring Queens-style addresses, you could replace `-` with `;` upfront before the rest of the routine.

ZeLonewolf: That would produce very wrong results for Hawaiian addresses.

1ec5: Oh, does Hawaii use Queens-style addressing too?

ZeLonewolf: I'm not familiar with Queens-style addressing, but in Hawaii (with the exception of the city of Honolulu), all addresses are prefixed with a 2-digit code indicating what sector of the island you're at.  For example, in ʻAiea the code is 99, so you would have addresses in the form `99-12 <street name>`. Conceptually, if anyone's tagged semi-colon separated addresses, you could have housenumbers that might look like `99-12;99-14`.  Treating the hyphen/dash as a delimiter would produce either `99-14` or `12-99`, both of which would be horribly wrong.  Further, on just a standard address of `99-12 Whatever Street`, if we treated the dash as a delimiter _and sorted it_, you'd end up with `12-99` which would indicate a location on a different sector of the island.  So hence my concern.

ZeLonewolf: From what I've read it sounds like yes, this is the same situation as Queens-style addressing, and I don't think it would be prudent to cause bad address numbers for an entire US state.

1ec5: What if this routine avoids sorting if something looking like a range appears anywhere in the raw data? Either you could look for a hyphen and call it a day, or you could check whether the hyphen separates values in ascending order.

It’s probably worth starting a broader tagging discussion about resolving this ambiguity. I could see a case for eliminating ranges from the tagging scheme and always enumerating the middle values, as long as more renderers and geocoders introduce behavior like what you’ve implemented. However, the well-established [Karlsruhe schema](https://wiki.openstreetmap.org/w/index.php?oldid=2323041#Using_interpolation_to_mark_many_houses_along_a_way) specifically allows for ranges. There are probably also edge cases to consider: what happens if a building is signposted with its start and end numbers but nothing in between? Ironically, you’d have to tag it as not a range in order for it to look like a range in OpenMapTiles.

ZeLonewolf: At the risk of inserting yet more en_US localization, perhaps the logic can be "if the string contains a dash, convert all semi-colons to commas", which would better serve the Queens/Hawaii problem.

Example: https://www.openstreetmap.org/node/2882804955

layers/housenumber/housenumber_display.sql
```diff
@@ -0,0 +1,37 @@
...
+DECLARE
+  arr text[];
+BEGIN
+  -- Convert the input string into an array of values
+  arr := string_to_array(raw_housenumber, ';')::text[];
+
+  -- Return the first and last value in the array
+  RETURN arr[1] || '–' || arr[array_length(arr, 1)];
+END;
+$$ LANGUAGE plpgsql;
```
pnorman, line 12: This function could be SQL, not plpgsql, for better performance.

layers/housenumber/housenumber_display.sql
```diff
@@ -0,0 +1,37 @@
...
+  END IF;
+
+  -- Find the minimum and maximum numbers in the list
+  SELECT MIN(value), MAX(value) INTO min_number, max_number
+  FROM unnest(string_to_array(raw_housenumber, ';')::int[]) AS value;
+
+  -- Return the consolidated range string
+  RETURN min_number::text || '–' || max_number::text;
+END;
+$$ LANGUAGE plpgsql;
```
pnorman, line 37: Same as other function. This could be a fairly simple CASE statement

pnorman (changes requested):
plpgsql functions take significantly longer to be called than SQL, particularly if the SQL gets inlined then there is no cost to calling the function. Both functions should be declared as `IMMUTABLE`

pnorman (commented):
SQL looks good - no opinion on the overall effect.

TomPohys:
Thanks a lot! It looks good. Is it ready to be merged?

ZeLonewolf:
Yes, ready here👍

TomPohys (approved):

PR #1564, opened 2023-07-10
Fix for classification of CA:ON:primary sub-national routes in Canada

### LINKED ISSUE #1563: suspicious classification of CA:ON:primary sub-national routes in Canada
Sub-national routes in Canada are among other things classifying network `CA:ON:primary` using a `CASE` statement. But all branches end up with value `ca-provincial-arterial`, see:

https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/update_route_member.sql#L59,L61

I guess the proper value for `ELSE` is `ca-provincial` otherwise the `CASE` statement does not make sense.

Locations affected should be in Canada and changes visible (slightly?) at Z4, as described PR https://github.com/openmaptiles/openmaptiles/pull/1446 which introduced that code.

phanecak-maptiler:
This PR changes classification of `CA:ON:primary`: "all else" should be `ca-provincial`.

This is a follow-up on https://github.com/openmaptiles/openmaptiles/pull/1446 and should fix #1563 .

### DIFF
--- a/layers/transportation/update_route_member.sql
+++ b/layers/transportation/update_route_member.sql
@@ -58,7 +58,7 @@ SELECT CASE
                CASE
                    WHEN ref LIKE '4__' THEN 'ca-provincial-arterial'::route_network_type
                    WHEN ref = 'QEW' THEN 'ca-provincial-arterial'::route_network_type
-                   ELSE 'ca-provincial-arterial'::route_network_type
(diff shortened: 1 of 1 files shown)


TomPohys (approved):

TomPohys:
Thanks, good catch!

PR #1568, opened 2023-08-06
Convert PL/PGSQL to SQL

ZeLonewolf:
This PR converts three plpgsql functions in the `transportation_name` layer to sql functions, which perform better.


TomPohys:
Thanks! Looks good!

TomPohys (approved):

PR #1574, opened 2023-09-06
Remove buffering from point-only layers

ZeLonewolf:
Based on discussion on Slack and initial experimentation, it appears that we can potentially eliminate buffers on point-only layers, at least as far as maplibre-gl-js rendering is concerned.  It appears that labels and icons still render as expected across tile boundaries. This would save significant duplication in POI or place-heavy tiles.

Opening as draft pending further tests and screenshots to validate whether this can be done safely.

CC/@msbarry


ZeLonewolf:
Closing this PR in favor of treating this as a render-time option in planetiler.

PR #1576, opened 2023-09-09
Add name fallback for cases where a route has no ref

ZeLonewolf:
This PR adds a route's `short_name` to the route_X attributes in transportation_name for cases where a route does not have a route number, but rather is named, for example in a case such as https://github.com/ZeLonewolf/openstreetmap-americana/pull/912.

This prevents OSM mappers from manufacturing route refs solely for the purpose of drawing route shields.


layers/transportation/update_transportation_merge.sql
```diff
@@ -84,12 +84,12 @@ FROM (
...
-        NULLIF(rm3.network, '') || '=' || COALESCE(rm3.ref, '') AS route_3,
-        NULLIF(rm4.network, '') || '=' || COALESCE(rm4.ref, '') AS route_4,
-        NULLIF(rm5.network, '') || '=' || COALESCE(rm5.ref, '') AS route_5,
-        NULLIF(rm6.network, '') || '=' || COALESCE(rm6.ref, '') AS route_6,
+        NULLIF(rm1.network, '') || '=' || COALESCE(rm1.ref, rm1.short_name, rm1.name, '') AS route_1,
+        NULLIF(rm2.network, '') || '=' || COALESCE(rm2.ref, rm2.short_name, rm2.name, '') AS route_2,
+        NULLIF(rm3.network, '') || '=' || COALESCE(rm3.ref, rm3.short_name, rm3.name, '') AS route_3,
+        NULLIF(rm4.network, '') || '=' || COALESCE(rm4.ref, rm4.short_name, rm4.name, '') AS route_4,
+        NULLIF(rm5.network, '') || '=' || COALESCE(rm5.ref, rm5.short_name, rm5.name, '') AS route_5,
+        NULLIF(rm6.network, '') || '=' || COALESCE(rm6.ref, rm6.short_name, rm6.name, '') AS route_6,
```
1ec5, line 92: Nothing about this syntax explicitly tells the client that the right-hand side of the equals sign is a ref versus a short name versus a name. Clients have to assume based on the `network` value. The nested key=value mini-language also doesn’t accommodate other attributes that identify a route relation, such as `ref:colour` for Spanish A-roads: ZeLonewolf/openstreetmap-americana#654.

It might be too late to replace this mini-language with something more structured, but for additional pieces of information like the name, I think separate `route_#:name` and `route_#:color` properties would be more usable. Can the schema support that?

ZeLonewolf: I think it would be a bad idea to add name as a general case when it's not needed as it would be unnecessary tile inflation. Hence the coalesce to ensure that routes have enough data to render shields without incorporating data that's unused.  Of course we could add attributes, but it would take this 6-way table join to a 12- or 18-way table join. That seems like a tall order simply to add data to support a few corner case networks. And adding 6 color attributes to support a single route network seems poor. At least if we extend the mini-languages, it doesn't expand the already excessive table join more.

1ec5: Do the columns automatically translate directly into properties, or is there an opportunity to intervene when writing out the properties? Instead of separate columns, there could be a single hstore column. Moreover, if you can assume that the client is only using these `route_*` properties in code (as opposed to expressions), then the value of `route:#` can simply be the hstore object converted to a JSON object string.

If we’re looking for more general use cases, one would be the route relation’s `wikidata` tag, so mappers no longer need to make up similar initialisms as `network`s for one-off routes. This will matter more for recreational routes.

1ec5: At least this mini-language could be laid out more like a key-value table:

```json
{
  "route_1": "network=US:KY\nshort_name=MP",
}
```

ZeLonewolf: Why is it important to encode whether the value comes from a ref, short_name, or name?  As a practical matter.

1ec5: Names aren’t necessarily suitable for direct display inside a shield. They’re longer, and if the route relation is part of a super relation structure, it might contain some disambiguation. The style might choose to map specific names to specific images, but that’s also less robust than (eventually) exposing something more stable like a Wikidata QID to more reliably identify an individual route.

Overall, my feedback is less about determining the source of the right hand side and more about extensibility. It is not the case that every route can be uniquely mapped 1:1 to a shield or other presentation based on its network and ref/short_name/name combination alone. This network=ref syntax is very easy to outgrow.

ZeLonewolf: > Names aren’t necessarily suitable for direct display inside a shield
The goal isn't to provide a pseudo-ref for displaying inside a shield.  The goal is to provide a unique identifier for a route so a client can make a decision about what to display.

I think a more robust extension, where we can add arbitrary attributes, is a better fit for OMT 4.0 aka a planetiler configuration language implementation that isn't so tightly coupled to the specific attribute list. My resistance to adding new attributes is tied to the complexity of the SQL approach, the need to explicitly declare tile attributes, and the code proliferation associated with attribute expansion. Substituting another OSM tag when `ref` would otherwise be present seems like an easy win to distinguish routes that would not normally be distinguishable from `network` and `ref` alone. The only other information I've heard might be missing is `colour`, which is a valid need for a few route networks but not incompatible with what I'm proposing to do for `ref` substitution.

layers/transportation/update_transportation_merge.sql
```diff
@@ -84,12 +84,12 @@ FROM (
         CASE WHEN highway IN ('footway', 'steps') THEN layer END AS layer,
         CASE WHEN highway IN ('footway', 'steps') THEN level END AS level,
         CASE WHEN highway IN ('footway', 'steps') THEN indoor END AS indoor,
-        NULLIF(rm1.network, '') || '=' || COALESCE(rm1.ref, '') AS route_1,
-        NULLIF(rm2.network, '') || '=' || COALESCE(rm2.ref, '') AS route_2,
-        NULLIF(rm3.network, '') || '=' || COALESCE(rm3.ref, '') AS route_3,
-        NULLIF(rm4.network, '') || '=' || COALESCE(rm4.ref, '') AS route_4,
-        NULLIF(rm5.network, '') || '=' || COALESCE(rm5.ref, '') AS route_5,
-        NULLIF(rm6.network, '') || '=' || COALESCE(rm6.ref, '') AS route_6,
+        NULLIF(rm1.network, '') || '=' || COALESCE(rm1.ref, rm1.short_name, rm1.name, '') AS route_1,
```
1ec5, line 87: If [a route relation](https://www.openstreetmap.org/relation/2555099) is tagged `unsigned_ref` but not `ref`, should it fall back to `name`? OSM Americana has been relying on a blank right-hand side to avoid rendering a shield. Would it now try to stuff “Farm to Market Road 1431 Business (J - Shaffer Bend)” into a county route shield?

If [a route relation](https://www.openstreetmap.org/relation/3075583) belongs to a network that is a mix of numbered and unnumbered routes, where the numbered routes have a slightly different shield than the unnumbered ones, would it end up showing the shield for a numbered route, where the number is actually a name with a cardinal direction as disambiguation?

If [a route relation](https://www.openstreetmap.org/relation/14287121) belongs to a network that the style [does not yet fully support with a custom shield](https://github.com/ZeLonewolf/openstreetmap-americana/issues/789), will it instead try to render the `name` as a very long plain-text shield, crowding out the road name label?

ZeLonewolf: So there's a few categories of problems described here.  It seems like the main issue is perhaps with `name`, which is more freeform and prone to descriptive naming vice `short_name`, which should have more deliberate use and less variability, and it would resolve the cases listed here.

I should still note that "a long name rendered in a shield" shouldn't be a problem since the renderer can simply restrict the length of possible string that can be drawn in the shield (the approach we take in Americana). Also, I note that the decision to render a default shield in the absence of style support is something we chose to do for data QA reasons, and I wouldn't blame the tile schema for awkward rendering because we're blindly drawing database content.

Would a fallback just to `short_name` solve our problem here?

1ec5: > I should still note that "a long name rendered in a shield" shouldn't be a problem since the renderer can simply restrict the length of possible string that can be drawn in the shield (the approach we take in Americana).

OSM Americana currently requires the `ref` to be [no longer than six characters](https://github.com/ZeLonewolf/openstreetmap-americana/blob/0a0c9fa53cb6eea478529c985520ccd6ffe3e1ee/shieldlib/src/shield.js#L29). This is a good criterion, because [some `ref`s](https://www.openstreetmap.org/relation/14548660) really do get that long.

[LA 1208-5](https://commons.wikimedia.org/wiki/File:Louisiana_1208-5_%282008%29.svg)

On the other hand, six characters could really be just a name in [hiragana](https://www.openstreetmap.org/relation/16237053), [kanji](https://www.openstreetmap.org/relation/11833981), or even [Latin](https://www.openstreetmap.org/relation/9204543).

> I wouldn't blame the tile schema for awkward rendering because we're blindly drawing database content.

I would _never_ blame a tile schema for awkward rendering – only for overconflation. There’s a reason OSM has maintained `name`, `short_name`, and `ref` as separate keys; any conflation of these keys necessarily results in a reduction in versatility even if there’s some gain in coverage.

> Would a fallback just to `short_name` solve our problem here?

It would certainly be an improvement over what’s here now, though it probably won’t be the end of the story. Abbreviations are somewhat language-specific, so inevitably there will be a route (probably a recreational route) that needs to be identified by a full name or, better yet, a Wikidata QID.

Note that `short_name` is only guaranteed to have fewer characters than `name`. I think most uses of it on route relations are similar in nature to a `ref`, but there’s nothing stopping someone from tagging, say, “Pennsylvania Turnpike” as `short_name=Penna. Turnpike` or “NS Vítězslava Nováka” as [`short_name=NS V. Nováka`](https://www.openstreetmap.org/relation/9043480). The [Palisades Interstate Parkway](https://www.openstreetmap.org/relation/412008) is tagged `short_name=Palisades Parkway` based on its colloquial name, though arguably “PIP” should be right there alongside it.

I wonder if there still needs to be some consideration of `name:signed` for denoting an unsigned named route, just as `unsigned_ref` denotes an unsigned numbered route. OSM Americana is right to suppress the shield if the relation is tagged `unsigned_ref`, but for instance [routes like these](https://www.openstreetmap.org/relation/6366232) lead me to think that unsigned short names aren’t that unlikely.

ZeLonewolf: > It would certainly be an improvement over what’s here now

Would it be an improvement over _what's currently in the tiles_?

> OSM Americana currently requires the ref to be [no longer than six characters](https://github.com/ZeLonewolf/openstreetmap-americana/blob/0a0c9fa53cb6eea478529c985520ccd6ffe3e1ee/shieldlib/src/shield.js#L29). This is a good criterion, because [some refs](https://www.openstreetmap.org/relation/14548660) really do get that long.

It seems reasonable that on the style side we could choose to override that on a case by case basis if we really do want to draw something longer.  And, if we're only using it as a key for choosing what shield to draw, it can effectively be arbitrary length.

1ec5: > Would it be an improvement over what's currently in the tiles?

For [the handful of networks](https://github.com/ZeLonewolf/openstreetmap-americana/blob/0a0c9fa53cb6eea478529c985520ccd6ffe3e1ee/src/layer/highway_shield.js#L3-L10) where OSM Americana has had to hard-code _road_ names in the shield selection logic, this would be a marginal improvement. For the larger set of scenic route networks that Americana plans to support, this would remove an existing obstacle. On the other hand, it might complicate future work on recreational route shields. (Proper recreational route shields are blocked by the lack of tagging consensus among mappers, let alone changes to OpenMapTiles.)

> And, if we're only using it as a key for choosing what shield to draw, it can effectively be arbitrary length.

This is what OSM Americana intends to use the right-hand side for certain values on the left-hand side. That’s a very specific presentational detail. By contrast, the style that’s in this repository really should be rendering concurrencies if it wants to mimic osm-carto and its generic route badges, but it won’t be able to tailor the logic by network because it’s a conventional GL style rather than a Web application that provides its own image fallback code.

Just to confirm: the only reason for this `{network}={ref}` format is that you’re stuck with what was chosen in #1152, correct? Americana is already [faking a `{network}={ref}\n{way_name}` format](https://github.com/ZeLonewolf/openstreetmap-americana/blob/0a0c9fa53cb6eea478529c985520ccd6ffe3e1ee/src/js/shield_format.ts#L17) by concatenating the road name onto a second line. If OpenMapTiles adds a `short_name` from the relation to that second line, Americana would be automatically compatible. But since Americana isn’t a code-frozen product, it could just as easily adapt to something more flexible like `network={network}\nref={ref}\nwikidata={wikidata}`.

ZeLonewolf: > Just to confirm: the only reason for this {network}={ref} format is that you’re stuck with what was chosen in https://github.com/openmaptiles/openmaptiles/pull/1152, correct? Americana is already [faking a {network}={ref}\n{way_name} format](https://github.com/ZeLonewolf/openstreetmap-americana/blob/0a0c9fa53cb6eea478529c985520ccd6ffe3e1ee/src/js/shield_format.ts#L17) by concatenating the road name onto a second line. If OpenMapTiles adds a short_name from the relation to that second line, Americana would be automatically compatible. But since Americana isn’t a code-frozen product, it could just as easily adapt to something more flexible like network={network}\nref={ref}\nwikidata={wikidata}.

I wouldn't say we're "stuck with it" but certainly `{network}={ref}` was the implementation I arbitrarily invented in #1152.

If we just added `short_name` on a second line to that attribute, it seems to be that would add lots of `short_name` values in places where it's not actually needed, and unnecessarily inflate tile size as a result.  The reason I'm suggesting it as a fallback for `ref` is to have a higher success in only expanding the size of attribute value when it's necessary for disambiguation.

1ec5: What if the second line with an explicit `short_name` only appears when `ref` is missing?

ZeLonewolf: Yes, that's what I'm suggesting we should do.

ZeLonewolf:
Superceded by #1620

PR #1579, opened 2023-09-14
Restore missing low-zoom transportation_name features

ZeLonewolf:
Unblocks #1576

This PR restores missing transportation_name features that are missing from zoom 11 and improves the mechanism by which transportation_name features are merged through lower zooms. This prevents route features from being dropped due to bridge interleaving (when a bridge interrupts an otherwise continuous route).

This fix moves the length check to a later stage of processing and merges bridges/tunnels if they're too small.  This is similar to the work done in the transportation layer to merge bridges that are too small to render lines.  In this layer, bridges are merged when they're too small to justify a separate label.

Current Release:
image

Before this PR:
image

After this PR:
image

### DIFF
--- a/layers/transportation_name/transportation_name.sql
+++ b/layers/transportation_name/transportation_name.sql
@@ -94,7 +94,7 @@ FROM (
                 NULL::int AS level,
                 NULL::boolean AS indoor
          FROM osm_transportation_name_linestring_gen3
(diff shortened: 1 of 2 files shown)


TomPohys (approved):

TomPohys:
Awesome! Thanks!

PR #1581, opened 2023-09-20
Update README.md - added link to workshop at FOSS4G 2021

michalgwo:
This workshop is extremely helpful in understanding how OpenMapTiles works, especially in creating new layers, which is not properly documented (or I just couldn't find any documentation).


TomPohys:
Thanks!

TomPohys (approved):

Issue #251, opened 2017-05-16
Import wikidata translations


klokan:
The translations available in OSM database has a lower coverage then Wikidata.

Wikidata can be used in the import to extend the number of translations for different places.

This ticket is about implementing wikidata import and binding - and filling the translations for more complete multilingual coverage - in #211

klokan:
@nyurik what process and data dump would you recommend for regularly importing the wikidata IDs + various translations of placenames into Postgres?

klokan:
New repo:
https://github.com/openmaptiles/import-wikidata

Official dumps:
https://www.wikidata.org/wiki/Wikidata:Database_download#JSON_dumps_.28recommended.29
https://dumps.wikimedia.org/wikidatawiki/entities/

Relevant info:
http://wiki.openstreetmap.org/wiki/Wikidata

nyurik:
@klokan I'm currently experimenting with importing Wikidata & OSM data into a local Blazegraph  (rdf db) instance.  The Wikidata import is relatively simple - you download the turtle file, and run two scripts to import it.  There is even an autoupdater script - running it keeps local wikidata always up to date. But this is not very convenient if you are trying to generate vector tiles using postgis. So for your usecase, it might be better to have a different script that downloads JSON dump and imports it directly into postgres, where you can do joins.  Once a ..., you can re-download/re-import/re-gen the whole thing.

stirringhalo:
@nyurik That sounds like a good idea, having an updater script (like the "update-osm") that runs continuously. Vector tiles can be generated off the join. Might be better to do a refresh of a materialized view to avoid too many big joins.

klokan:
Related comment: https://github.com/OSMNames/OSMNames/issues/58#issuecomment-304661743

klokan:
https://www.mapbox.com/blog/streets-arabic-portuguese/
(greetings to @lukasmartinelli at Mapbox)

jirik:
Pull request opened: https://github.com/openmaptiles/openmaptiles/pull/358

jirik:
Closed by https://github.com/openmaptiles/openmaptiles/commit/741be6d92109191a082c28959bc86bffe1568ba6

PR #1584, opened 2023-09-26
fix unexpected error brought by typo: tigger -> trigger & fix `trigger_flag` was created on a wrong target

plainheart:
Hi, I'm just working on OpenMapTiles, it's very cool! There are two issues that may be a bug:

1) The trigger name has a typo of `trigger`, which is incorrectly spelled as `tigger` and doesn't match the following create command: `CREATE TRIGGER trigger_flag`. This may result in an error when invoking `make import-sql`.

```
ERROR:  trigger "trigger_flag" for relation "osm_park_polygon_gen_z4" already exists
xargs: sh: exited with status 255; aborting
```

2) The `trigger_flag` may be created on the wrong target in line [#251](https://github.com/plainheart/openmaptiles/blob/master/layers/park/update_park_polygon.sql#L251). I think it should be an unintentional mistake. I'm sorry and will revert this change if it's not so.


TomPohys (approved):

TomPohys:
Thanks! Good catch! Looks good.

PR #1589, opened 2023-10-21
Add natural=valley

### LINKED ISSUE #274: Missing natural data

zstadler:
https://github.com/openmaptiles/openmaptiles/issues/274 requested the addition of several `natural` features.
https://github.com/openmaptiles/openmaptiles/pull/1202 addressed most of them.
However, `natural=valley` ways were excluded from the PR.
Following the [discussion about this exclusion](https://github.com/openmaptiles/openmaptiles/pull/1202#issuecomment-1773743352), it looks like it would be fine to add them and resolve https://github.com/openmaptiles/openmaptiles/issues/274


TomPohys:
Thanks! Maybe one improvement for `valley` should be `LineLabel` function (example in `water_name` layer). Some of the valleys could be very short even for zoom 13 e.g. https://www.openstreetmap.org/way/451566241.

https://github.com/openmaptiles/openmaptiles/blob/6c31841f4674f15e15afde346a060cf7c22e6cdd/layers/water_name/water_name.sql#L34

zstadler:
> Thanks! Maybe one improvement for `valley` should be `LineLabel` function (example in `water_name` layer). Some of the valleys could be very short even for zoom 13 e.g. https://www.openstreetmap.org/way/451566241.
> 
> https://github.com/openmaptiles/openmaptiles/blob/6c31841f4674f15e15afde346a060cf7c22e6cdd/layers/water_name/water_name.sql#L34

LGTM

TomPohys:
Adding the `LineLabel` will be beneficial. Also please consider excluding the `valley` without `name` (or how you would like to style just the line?). Could you please update the style to show the names of the valleys?

Could you please update the PR? 

Thanks!

zstadler:
@TomPohys,

On second thought, I prefer not to expand this PR. The intention here is for a minimal addition to the layer where `natural=valley` would follow the existing handling of other linestring elements, such as `natural=arete`.

Each of the two enhancement recommendations requires a non-trivial enhancement of the layer:
1. If the `LineLabel` function is used, it should be applied to all linestring features, probably except `natural=cliff`.
   - How would/should the `LineLabel` function interact with the `LabelGrid` function
      https://github.com/openmaptiles/openmaptiles/blob/e61442c49dd32974d4952046b4149202c7072d59/layers/mountain_peak/mountain_peak.sql#L104
2. The style does not seem to handle linestring elements appropriately. They are handled as if they were a point element. As a result:
   - Cliffs are not shown at all
   - A `peak` icon is used for all other elements, even for a `natural=arete`
   - The icon and the name are positioned at the edge of the line, rather than at the center or along the line itself

I prefer that these enhancements will be put in a separate issue/s that would be handled by a different person.

ZeLonewolf:
Cliffs, ridges, and arêtes are all commonly rendered on maps as a line feature without labelling (though I would love to have labelled ridges in OpenMapTiles).  Valleys on the other hand are rarely rendered on maps as a line feature only - they are usually rendered only as a label feature.  So the primary cartographic reason for the linestring for valleys are as placement for a label.  So I'm not sure what benefit we really gain from adding valleys if we don't also include sufficient attributes (`name` and `name:xx` tags) as @TomPohys suggests to draw a label.

TomPohys:
Hi @zstadler, thanks again for opening this PR!
I realize it has been inactive for quite some time, so I just wanted to check if you’re still planning to continue working on it.

If so, could you please take a look at the suggested changes (show only valleys with a name and use LineLabel)?

Happy to help if needed. Thanks!

zstadler:
Hi @TomPohys ,

Please feel free to apply the change you proposed, as well as any other change you see fit, to this PR.

PR #1591, opened 2023-10-24
Remove oneway attributes from ferry lines and railway

ZeLonewolf:
This PR removes the `oneway` attribute from ferry lines and rail lines.  These attributes don't really make sense for these feature classes and they are not typically rendered in general-purpose maps.

CC / ZeLonewolf/openstreetmap-americana#972


TomPohys (approved):

TomPohys:
Thanks! Make sense.

PR #1602, opened 2023-12-23
Remove expressway attribute from ramps

ZeLonewolf:
#1313 added `expressway` tagging to the transportation layer. However, this also added the `expressway` attributes to ramps where it doesn't make sense (ramps by definition are controlled-access roads in all cases).  This PR corrects this by dropping the `expressway` attribute from ramps where it doesn't make sense to do so.

See also ZeLonewolf/openstreetmap-americana#1017

Tested on Great Britain location:
image

Regression check:
image


TomPohys:
Awesome, thanks!

TomPohys (approved):

PR #1526, opened 2023-04-04
Update park layer description to include indigenous lands addition

ZeLonewolf:
This PR updates the park layer description to reflect #1489 and clarify why we chose to place them in this layer (technology reasons).

PR #1604, opened 2023-12-24
Move aboriginal lands boundaries to boundary layer

ZeLonewolf:
Unblocks ZeLonewolf/openstreetmap-americana#105

#1489 added indigenous land boundaries to OpenMapTiles. For technical reasons, I added this to the `park` layer in order to take advantage of the existing processing chain in that layer, which handled large polygons very well.  While we got indigenous lands added to the tiles and took advantage of this processing, organizing them under a layer named parks was poor from a logical organization and historical context perspective.  Thus, in #1526 we expanded the description of that layer to clarify why these objects were placed there.

However, while we used the park layer to take advantage of the technical processing chain, this caused the follow problems:
1. Styles that use undifferentiated rendering of all objects in the `park` layer now found themselves rendering indigenous lands as parks, which should be regarded as a breaking schema change. This is fixable on the style side but requires styles to make a change as we did in OSM Americana in ZeLonewolf/openstreetmap-americana#1019.
2. Indigenous lands were dissolved along with other area features in the `park` layer at z4, which made it impossible to differentiate indigenous lands from parks at that zoom. This is a bug. This means that any style which does not desire to render indigenous lands as parks must disable the z4 park layer, as we did in ZeLonewolf/openstreetmap-americana#1020.

After a Slack [discussion](https://osmus.slack.com/archives/CH6GFAFRC/p1703361153711949) in the #openmaptiles channel, there was support for moving indigenous lands into the `boundary` layer. While `boundary` is the best fit from a schema perspective (grouping like objects together), there are presently neither area features nor processing currently in place in that layer.

This PR adds indigenous lands to the `boundary` layer by adding new area features for these objects, along with generalization up to z4. It borrows heavily from the technical processing chain of the `park` layer.

This limited rendering of Oklahoma, USA shows both park and boundary area features rendered together:
image

I added new `class`, `name`, and `name:xx` attributes to the boundary layer, in order to support the new area features. Should we (or someone extending this schema) come up with new objects to place in the boundary layer in the future, `class` would support additional categories.

Indigenous lands will have the attribute `class=aboriginal_lands`. Additionally, I removed #1489 and #1526 as this work supplants that.

Since we have not yet released OpenMapTiles 3.15, now is the time to get this right.


1ec5:
> I added new `class`, `name`, and `name:xx` attributes to the boundary layer, in order to support the new area features. Should we (or someone extending this schema) come up with new objects to place in the boundary layer in the future, class would support additional categories.

Actual name properties will be a welcome improvement to the `boundary` layer, as opposed to the existing alpha-3 country codes that aren’t very flexible or interoperable. Hopefully the new name properties won’t preclude later differentiating between left-hand and right-hand names on the linestrings in this layer.

ZeLonewolf:
> Actual name properties will be a welcome improvement to the `boundary` layer, as opposed to the existing alpha-3 country codes that aren’t very flexible or interoperable. Hopefully the new name properties won’t preclude later differentiating between left-hand and right-hand names on the linestrings in this layer.

Just to be clear, I'm only adding these attributes to the aboriginal lands polygons that I've added in this PR to keep this narrow in scope.

In the absence of proper design work, I can't say what impact this will have on future improvements to this layer. But, I will sketch out my thoughts here.

What would make the most sense for a proper boundary implementation is to have both polygons and linestrings for administrative boundaries, so that the polygons can be used for naming and the linestrings can be used for styling boundaries without overlapping dash patterns.  We might also similarly add aboriginal land boundary linestrings for the same reason, and then we could populate `class` on linestrings to differentiate them. Then, a data consumer that desires to show inner-boundary-edge labels could simple draw them on the inside of the polygons, with the style handling overlapping boundaries of multiple admin_level values.

With that implementation, you could fully-deprecate the 3-letter codes and the awful complicated SQL that generates them.

So in a nutshell I'd advocate a complete implementation that looks like:
* Polygons for administrative areas with `class=administrative`
* **Polygons for aboriginal lands with `class=aboriginal_lands`**
* **Boundary linestrings for administrative areas** with `class=administrative`
* Boundary linestrings for aboriginal lands with `class=aboriginal_lands`

After this PR, we will have implemented just the parts in bold above.

IMO, completing this implementation is required in order to accomplish this effect seen here on the US/Canada boundary:
image

The next question to consider is, _would completing this implementation be a breaking change to the schema?_

I think the answer is no.  First of all, boundary polygons don't exist prior to this PR, so no breaking change there, just an addition.  The only potential collision is in adding aboriginal lands linestrings.  However, unless a style is using an incredibly primitive implementation, they are already using `admin_level` to differentiate different boundary linestrings.  So, adding `class=` should not risk rendering indigenous land boundaries where administrative boundaries are expected due to the lack of `admin_level`.  And, in the case where an indigenous land boundary _does_ have a properly-tagged `admin_level` (these cases exist), it would argue that rendering them like other boundaries of the same `admin_level` would be reasonable and appropriate and thus nonbreaking.

In this scheme, the `name` and `name:xx` would be populated on the polygons and omitted from the linestrings, since it's the polygons that would determine inner name labels rather than a set of left/right attributes on the linestrings.  The current 3-letter codes would become effective (if not outright) deprecated.

So this is a long-winded way of saying that I believe the new attributes I've added here support a proper, full implementation of boundary linestring and label processing.

TomPohys:
Thanks, @ZeLonewolf, for a detailed comment. It looks awesome. 

I have only one concern about showing the land's name. Do you think adding `boundary=aboriginal_lands` into the `place` layer as a point feature will be beneficial for styling? Or adding directly into the `boundary` layer? As it is now, there is no nice way how to style a polygon name in its centroid.

ZeLonewolf:
@TomPohys I've added additional point objects to the place layer (and in doing so, made place depend on boundary so we could generate just one set of boundary polygons in the database.

Example from Oklahoma, USA:

image

TomPohys:
Awesome! Thanks!

TomPohys (approved):

PR #1605, opened 2023-12-24
Document `name:xx` and future removal of `name_de` and `name_en`. 

ZeLonewolf:
This PR is a **documentation change only**. It documents the fact that `name:xx` values are present in the tiles and documents that `name_de` and `name_en` will be removed in a future release in favor of `name:de` and `name:en` respectively.


TomPohys (approved):

TomPohys:
Thanks! Looks good.

PR #1607, opened 2023-12-26
Fix university office/amenity collision

### LINKED ISSUE #1606: POI for `amenity=university` is incorrectly conflated with `office=university`
University POIs are currently coded as `class=office` when it should be `class=college` (with subclass values of `college` and `university` within that).  This is a bug:

image

This issue occurs because of the subclass collision (the same subclass `university` exists within two different `class` values with different meaning).

ZeLonewolf:
Fixes #1606
Unblocks ZeLonewolf/openstreetmap-americana#1022

This PR is a bugfix which ensures that `amenity=university` is encoded as `class=college` rather than `class=office`.  It's required because of a subclass collision for `university`, which exists under both class values.


TomPohys (approved):

TomPohys:
Thanks, it makes sense.

Issue #1613, opened 2024-01-15
Unsigned routes should not be included in the transportation_name layer


Blijbol:
The `transportation_name` layer is intended for labelling highways. However, routes are currently included in this layer even if they are explicitly tagged with `unsigned=yes`.

For example, [I 375 Business Spur (Detroit, MI)](https://www.openstreetmap.org/relation/14460380) is currently rendered by OpenMapTiles:

![Schermafdruk van 2024-01-15 23-11-25](https://github.com/openmaptiles/openmaptiles/assets/489845/fd51223f-fd4d-41c6-b07d-faeb0b2fb763)

For the vast majority of non-expert use-cases, maps will be better if they do not show unsigned routes. I therefore suggest to exclude routes with `unsigned=yes` from the `transportation_name` layer.

reventlov123:
This is based (IMO) on the erroneous tagging of unsigned routes as "unsigned=yes". They should be tagged with "unsigned_ref". Tagging as "unsigned" in the given example would imply that there is not only no signage for the highway designation, but no signage for the "street name" (which presumably is not the case).

Blijbol:
> Tagging as "unsigned" in the given example would imply that there is not only no signage for the highway designation, but no signage for the "street name" (which presumably is not the case).

I am talking about the tagging of the relation, not the tagging of the way. Tagging `unsigned=yes` on the route relation for the I 375 Business Spur simply indicates that the I 375 Business Spur is unsigned. The street name (East Jefferson Avenue) comes from the highway way and has nothing to do with the route relation.

reventlov123:
Apparently I misunderstood the example you gave... after "driving" it on street view, I now get that there are actually no signs for the route whatsoever... it's not a "virtual congruency" but a "virtual highway". The "highway" signs along it are all directional ("to" whatever).

PR #1614, opened 2024-01-15
Transportation: add filter to reject unsigned routes

Blijbol:
This is my suggested fix for issue #1613. It discards transportation routes tagged with `unsigned=yes`.

The code is inspired by the already-existing filtering of water polygons with `covered=yes` in the water layer.

Disclaimer: I have no experience with this code repository and have not tested this.

### DIFF
--- a/layers/transportation/mapping.yaml
+++ b/layers/transportation/mapping.yaml
@@ -491,6 +491,9 @@ tables:
     - name: colour
       key: colour
       type: string
+    filters:
+      reject:
+        unsigned: ["yes"]
     mapping:
       route:
       - road


ZeLonewolf:
In principle, this change is fine, but `ref` + `unsigned=yes` is an obscure combination that, according to taginfo, only occurs in 189 (now 188) cases.  The more typical method of signing unsigned routes is to use `unsigned_ref`, which has over 258,000 usages.

Source:
https://taginfo.openstreetmap.org/keys/unsigned#combinations
https://taginfo.openstreetmap.org/keys/unsigned_ref#overview

The case that you found in Michigan, I would regard as a tagging error and I've corrected it to use `unsigned_ref`, as is the norm in the United States.  I would suggest changing the tagging on the 188 instances of `ref` + `unsigned=yes` would be a better course of action rather than encouraging a duplicate tagging system to arise via schema usage.

1ec5:
For context, a route can be signposted independently of its number (ref). For example, [this route](https://www.openstreetmap.org/relation/317862) is conspicuously signposted with signs that indicate the route’s name but not its number. Its number exists only for internal inventory purposes, but that number is tagged as `unsigned_ref` so that someone won’t come along, think it’s missing, and add it as `ref`. Conversely, there are routes marked by signs that only mention their numbers rather than their names.

[`unsigned=no`](https://wiki.openstreetmap.org/wiki/Key:unsigned) is documented as applying to the feature’s name, not its number. If you encounter a relation that uses this tag to indicate that only the number is unsigned, `unsigned_ref` or `ref:unsigned=no` would be better alternatives, depending on how certain you are that the route has no signs anywhere.

reventlov123:
I've edited https://wiki.openstreetmap.org/wiki/Key:unsigned to attempt to make the situation more clear. "unsigned" would imply that a feature has no on-the-ground signage **at all**. "unsigned_ref" implies that the feature does actually have on-the-ground signage, but that the signs don't display the particular legal designation.

1ec5:
Even so, the fact that this key originally only pertained to the name (in the context of bus stops) means that data consumers need to be careful interpreting it to mean anything other than an unsigned name, depending on how lenient or strict the use case needs to be. OpenMapTiles may get away with hiding these features out of an abundance of caution, but redefining the key (even in a very sensible manner) probably deserves at least a brief discussion in the OSM community beyond this relatively obscure PR.

ZeLonewolf:
> For example, replacing `ref=*` with `unsigned_ref=*` on the route relation is problematic if the `ref=*` is actually important for identifying the unsigned route (e.g. unsigned E-roads). It e.g. breaks the labelling of the relation based on `ref=*` on the OpenStreetMap website and in the JOSM editor.

This does not seem to be a problem in practice.  In the UK, where E-roads are unsigned, `unsigned_ref` doesn't seem to break things:

https://www.openstreetmap.org/relation/6246893

image

1ec5:
> It e.g. breaks the labelling of the relation based on `ref=*` on the OpenStreetMap website and in the JOSM editor.

I hadn’t thought of that. We should probably make sure that all the editors and the website fall back to `unsigned_ref` for the purpose of relation labels in non-map UI elements, regardless of what OMT does about `unsigned`.

pnorman (changes requested):
Quoting from [trolltag](https://wiki.openstreetmap.org/wiki/Trolltag),

> It is not OK to use one tag and add second tag that negates or massively change its meaning. Additional tags should clarify meaning of main tags rather than negate it.
> 
> In general, any tag that must be processed to avoid serious misinterpretation of the feature is a trolltag.

This is an excellent example of such a tag. `unsigned=yes` goes against the common understanding of ref, and we already have a standard tag for unsigned refs. Quoting from the ref page, 
> [unsigned_ref](https://wiki.openstreetmap.org/wiki/Key:unsigned_ref)=* - reference number or code that is generally not signed on the route

TomPohys:
Thanks a lot for the discussion here and in #1613. I'm going to close this PR since the main cause was fixed directly in OSM. OMT doesn't currently filter out `unsigned` or `unsigned_ref` tags at all, and even though these routes lack physical signs, the legal or administrative designations still exist, so there is value in at least having them in the map data.

1ec5:
> We should probably make sure that all the editors and the website fall back to unsigned_ref for the purpose of relation labels in non-map UI elements, regardless of what OMT does about unsigned.

Since this discussion, openstreetmap/iD#9588 was finally merged, so that the label does indeed fall back to `unsigned_ref=*`.

PR #1617, opened 2024-01-19
Keep the filter for motorways on z4 same as z5.

TomPohys:
During development of https://github.com/openmaptiles/openmaptiles/pull/1440 and https://github.com/openmaptiles/openmaptiles/pull/1446 were prefered national-importance motorways included in https://github.com/openmaptiles/openmaptiles/blob/eb7f6be45545cdb2db0afbe180864c9b0afb18c1/layers/transportation/network_type.sql#L26-L34

It works for the US and the CA, where these attributes are filled. But the rest of the world on a zoom 4 is empty. 

This PR returns highways for a zoom 4 worldwide.

@ZeLonewolf as these two PRs are from you, is it OK to keep it like this?

Visuals of the Czech Republic (over zoom with 200% to see more CZ)

Current state ->
Screenshot from 2024-01-19 17-45-33

PR proposal ->
Screenshot from 2024-01-19 17-45-37

### DIFF
--- a/layers/transportation/update_transportation_merge.sql
+++ b/layers/transportation/update_transportation_merge.sql
@@ -807,8 +807,7 @@ BEGIN
             WHERE transportation.changes_z4_z5_z6_z7.is_old IS FALSE AND
                   transportation.changes_z4_z5_z6_z7.id = osm_transportation_merge_linestring_gen_z5.id
         )) AND
-        osm_national_network(network) AND
-        -- Current view: national-importance motorways and trunks
+        -- Current view: same as z5
         ST_Length(geometry) > 1000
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
I think allowing all motorways at z4 is too much motorway density at that level, which is why I took steps to only show the most important roads.  I think it would be a better OpenMapTiles product to add e-Roads (European E-route) and A roads (Asian route network) to the z4 filter, plus a smattering of national route systems.  If we show "more" at z4, then we would definitely want to have the output of `osm_national_network(..)` somehow reflected in the tiles so I could filter out the unwanted roads at z4.

TomPohys:
Thanks for the suggestion. I will close this PR and open another one for adding `e-road` and `a-road`.

PR #1619, opened 2024-01-21
Add e-road and a-road for transportation z4.

TomPohys:
Adding [e-road](https://wiki.openstreetmap.org/wiki/WikiProject_Europe/E-road_network) and [AsianHighway](https://wiki.openstreetmap.org/wiki/Asia/Asian_Highway_Network) (as `a-road`) as part of the network to show up from zoom 4.

example of `a-road`
image

example of `e-road`
image

Follow up of https://github.com/openmaptiles/openmaptiles/pull/1617


layers/transportation/update_route_member.sql
```diff
@@ -70,6 +70,10 @@ SELECT CASE
            WHEN network = 'omt-ie-motorway' THEN 'ie-motorway'::route_network_type
            WHEN network = 'omt-ie-national' THEN 'ie-national'::route_network_type
            WHEN network = 'omt-ie-regional' THEN 'ie-regional'::route_network_type
+           -- https://wiki.openstreetmap.org/wiki/WikiProject_Europe/E-road_network
+           WHEN network = 'e-road' THEN 'e-road'::route_network_type
+           -- https://wiki.openstreetmap.org/wiki/Asia/Asian_Highway_Network
+           WHEN network = 'AsianHighway' THEN 'a-road'::route_network_type
```
ZeLonewolf, line 76: I would recommend moving the international routes above the national routes.  If a road is both an international route and a national route, the international one is more "important".

TomPohys: Thanks, fixed in https://github.com/openmaptiles/openmaptiles/pull/1619/commits/93f8872e948b544a8a23d5c78c5b6b421e3a8378

phanecak-maptiler (approved):

PR #1620, opened 2024-01-23
[WIP] Expanded road route attributes

ZeLonewolf:
Alternative approach to #1576

This PR provides expanded route attributes that allow for a complete implementation of highway shields which includes a few categories of edge case that aren't currently supported. Networks that share a common `network` value but different shields, such as [Vermont's Scenic Byways](https://vermontvacation.com/~/media/Files/PDFs/Byways/Vermont%20Byways/2019%20Byways%20Brochure%20-%20WEB), can now be rendered with this PR.

Upon discussions in Slack, there is a concern that the current "mini-language" for expressing route information is brittle and should have a decomposed implementation to better express route information, avoid manufactured `ref` tagging for the renderer, and simplify style library (e.g. maplibre) implementations of route rendering.

The current approach combines `network` and `ref` information into a mini-language by packing a `route_N` attribute as a key-value pair.  For example, [California State Route 99](https://en.wikipedia.org/wiki/California_State_Route_99) would be represented in the tiles like this:

`route_1`=`US:CA=99`

This works for _most_ route networks, but fails to support two less-common cases of route:
1. Routes which have a name but no route number or ref.
2. Routes which have a color-coding system.

There are many named routes that lack a route number.  Some examples in OSM include:
* The [Spaulding Turnpike](https://www.openstreetmap.org/relation/13048921) in New Hampshire, USA
* The [Forty-Nine Mile Scenic Drive](https://www.openstreetmap.org/relation/109668) in San Francisco

Color-coded routes include [highways in Spain](https://en.wikipedia.org/wiki/Highways_in_Spain) (blue/red color codes) as well as [Yukon territorial highways](https://en.wikipedia.org/wiki/List_of_Yukon_territorial_highways) in Canada, which use four colors. These routes are color coded with the OSM `colour` tag. A description of the colour issue can be found in ZeLonewolf/openstreetmap-americana#654.

Example: [Autovía de Guadix a Límite de Región de Murcia](https://www.openstreetmap.org/relation/9465880)

## Approach

This PR replaces the `route_N` attribute scheme with individual attributes for name and colour on routes.  Thus you will have:
* `route_N_network` to hold the route `network` value
* `route_N_ref` to hold the route `ref` value
* `route_N_name` to hold the route `name` value
* `route_N_colour` to hold the route `colour` or `ref:colour` value

...for N values 1-6, which is the present behavior.

Behind the scenes, the `route_N` columns used throughout the SQL code have been converted to an hstore object without rename, which simplifies the changes required.

This has an efficiency advantage in that it minimizes the number of unique text strings that need to be stored - we can take advantage of repeated `network` and `ref` values rather than tons of unique strings like `US:CA=99`. Although, we are also storing the name now which adds some tile size in cases where the name is represented in the attributes.

A live preview of this change being handled by OpenStreetMap Americana can be found [here](https://preview.ourmap.us/pr/1032/#map=6/43.157/-74.062).

## Examples

The examples below are rendered on my [tileserver](https://tile.ourmap.us/inspect-massachusetts.html) from this PR.

1. Interstate highway in Connecticut, USA ([sample location](https://tile.ourmap.us/inspect-massachusetts.html#9/41.7518/-72.3516)):
image

2. Concurrent named route and state highway in Connecticut, USA ([sample location](https://tile.ourmap.us/inspect-massachusetts.html#11/41.0628/-73.3812)):
image

3. Spain color-coded routes ([sample location](https://tile.ourmap.us/inspect-massachusetts.html#9/42.8697/-7.4435)):
image

## Impact on Existing Users

Route information is recent addition to OpenMapTiles, so the schema upgrade should have minimal impact. Known users of route information include OpenStreetMap Americana and the AARoads style.  Americana has already developed a PR for consuming expanded route data in ZeLonewolf/openstreetmap-americana#1032 and AARoads (tbd). Since this PR changes the attributes names (from `route_N` to `route_N_keyword`), there should be no collision from changed meanings.

## Performance

In this PR, I investigated whether route name should be kept in all cases where it's present in the data or only in cases where there's no ref.  I found that there was very little difference in tile size when allowing the route names to be included in the data in all cases.  This allows for renderings where a renderer chooses to render the names of multiple route concurrencies, possibly alongside shields.

Tile size increase taken from CI for `ref` + `name` attributes (`name` attributes included in all cases, even when a `ref` is present):
image

Same graph for suppressing `name` when `ref` is present:
image


TomPohys:
Thank you. It looks good.

TomPohys (approved):

Issue #1595, opened 2023-11-24
problem with "James Bay" in water_name (water label)


phanecak-maptiler:
There is a problem with https://github.com/openmaptiles/openmaptiles/pull/1457 in regards to "James Bay" (possibly others) as following changes:

* https://github.com/openmaptiles/openmaptiles/pull/1457 , e.g. `natural: bay` and `... straight` added into mappings
* https://github.com/openmaptiles/openmaptiles/pull/1457/files#diff-201daa1c61c99073fe3280d440c9feca5ed2236b251ad454caa14cc203f952d1R76 , e.g. `WHEN zoom_level >= "rank" AND "rank" IS NOT NULL THEN TRUE`

clash with:

* https://github.com/openmaptiles/openmaptiles/blob/master/layers/water_name/update_marine_point.sql#L21,L22 , e.g. `LEFT JOIN ne_10m_geography_marine_polys AS ne ON lower(trim(regexp_replace(ne.name, '\\s+', ' ', 'g'))) IN (lower(osm.name), ...`

In OpenMapTiles 3.14 that name matching works OK (since only `place: ocean` and `... sea` are in mappings), but in current `master` it means that thanks to https://www.openstreetmap.org/relation/13360255 being listed in Natural Earth with "rank: 3", several other much smaller "James Bay" points/nodes get "rank: 3" too, for example:

* https://www.openstreetmap.org/node/1385157299

Which means that we get on Z3 more than we should/want to:

![Screenshot from 2023-11-23 22-10-03](https://github.com/openmaptiles/openmaptiles/assets/115141505/00ca8290-d8dd-4f23-8da8-a34e4a9ec068)

I'm not sure about how to tackle this properly, so I'd would like some help with that, say from @ZeLonewolf or maybe @TomPohys or others.

pnorman:
> * LEFT JOIN ne_10m_geography_marine_polys AS ne ON lower(trim(regexp_replace(ne.name, '\\s+', ' ', 'g'))) IN (lower(osm.name), ...

[...]

PR #1621, opened 2024-01-24
fix for problem with "James Bay" in water_name (water label) #1595

phanecak-maptiler:
This should fix the problem described in #1595 .

Fix is similar to what is used in `update_city_point.sql` as suggested by @pnorman, see https://github.com/openmaptiles/openmaptiles/issues/1595#issuecomment-1877961486 .

Screenshot from the area mentioned in the issue: 
Screenshot from 2024-01-24 16-43-48

See also "Java version" in `planetiler-openmaptiles`: https://github.com/phanecak-maptiler/planetiler-openmaptiles/pull/35


TomPohys (approved):

TomPohys:
Thanks! Looks good!

PR #1623, opened 2024-01-26
Make sure areas above 160000000 get rank 2 or 1

### LINKED ISSUE #1622: Wrong rank for "Seminole Nation": 3 instead of 1
PR https://github.com/openmaptiles/openmaptiles/pull/1604 introduces (or moves) `aboriginal_lands` into `boundary` and `place` layers. @ZeLonewolf used "Seminole Nation" (https://www.openstreetmap.org/relation/9405576) as an example of point in `place` layer, see https://github.com/openmaptiles/openmaptiles/pull/1604#issuecomment-1902197768 .

While re-implementing that into `planetiler-openmaptiles` I'm getting rank 1 for that land, while ZeLonewolf's screenshot shows 3. I believe that is caused by the typo in the `area_rank()` function adjusted in the PR:

```
SELECT CASE
           WHEN area < 10000000 THEN 6
           WHEN area BETWEEN 1000000 AND 15000000 THEN 5
           WHEN area BETWEEN 15000000 AND 40000000 THEN 4
           WHEN area > 40000000 THEN 3
           WHEN area > 160000000 THEN 2
           WHEN area > 640000000 THEN 1
           ELSE 7
           END;
```

With statement as such, values greater than 160000000 or 640000000 will trigger `area > 40000000` part and get rank `3` instead or `2` or `1`.

phanecak-maptiler:
This is a follow-up on https://github.com/openmaptiles/openmaptiles/pull/1604 . Seminole Nation is shown in the screenshot there with rank 3 while based on area it should be 1. This PR fixes `area_rank()` function to handle areas greater than 160000000, e.g. this fixes #1622 .


ZeLonewolf:
Thank you for fixing this!  I assumed that the CASE - WHEN sequence would get evaluated in order, and I'm surprised that the BETWEEN was requred.

ZeLonewolf:
Oh, wait, I see what I did now 😅

ZeLonewolf (approved):

pnorman (commented):
Looks good, but a way that this could be written without having to duplicate the values between lines is

```sql
SELECT CASE
WHEN area > 640000000 THEN 1
WHEN area > 160000000 THEN 2
WHEN area > 40000000 THEN 3
...
WHEN area > 1000000 THEN 5
WHEN area > 0 THEN 6
ELSE 7
END

pnorman (approved):

TomPohys:
Thanks!

PR #1625, opened 2024-01-29
Fix empty brunnel attribute on transportation_name

ZeLonewolf:
I noticed in some cases that empty aggregated `brunnel` attributes was coming through in the tiles in the `transportation_name` layer:
image

This PR adds the missing NULLIF to correct this.


TomPohys (approved):

TomPohys:
Thanks! Good catch!

PR #1626, opened 2024-01-30
🐛 remove duplicate "requires" key

makeeno:
We're reading and parsing the Yaml files within Node.Js and noticed that the parser won't handle the scenario where an object has the same keys twice.

I also assume that the first one is being ignored in favor of the second one.


TomPohys (approved):

TomPohys:
Thanks! Good catch.

PR #1627, opened 2024-01-31
Moved confectionery from ice_cream class to shop

michalgwo:
Correct me if I'm wrong, but ice_cream doesn't sound like a good class for confectionery. It also confuses renderers and creates something like that (photo from MapTiler OpenStreetMap style):

[OSM POI location](https://www.openstreetmap.org/node/6357619085)
image

This happened because of the classification of confectionery to ice_cream class instead of shop.

### DIFF
--- a/layers/poi/poi.yaml
+++ b/layers/poi/poi.yaml
@@ -18,7 +18,7 @@ layer:
       values:
         shop:
           subclass: ['accessories', 'antiques', 'beauty', 'bed', 'boutique', 'camera', 'carpet', 'charity', 'chemist',
-                     'coffee', 'computer', 'convenience', 'copyshop', 'cosmetics', 'garden_centre', 'doityourself',
+                     'coffee', 'computer', 'convenience', 'confectionery', 'copyshop', 'cosmetics', 'garden_centre', 'doityourself',
                      'erotic', 'electronics', 'fabric', 'florist', 'frozen_food', 'furniture', 'video_games', 'video',
                      'general', 'gift', 'hardware', 'hearing_aids', 'hifi', 'ice_cream', 'interior_decoration',
                      'jewelry', 'kiosk', 'locksmith', 'lamps', 'mall', 'massage', 'motorcycle', 'mobile_phone',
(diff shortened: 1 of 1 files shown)


TomPohys:
Hi, no problem move the confectionery to `class=shop`. The icons for OpenMapTiles style (default in this repo) is taken from `subclass` which mean it takes icon -> [confectionery.svg ](https://github.com/openmaptiles/openmaptiles/blob/master/style/icons/confectionery.svg). 
This is also the correct icon for [OpenStreetMap Carto](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dconfectionery).

michalgwo:
> Hi, no problem move the confectionery to `class=shop`. The icons for OpenMapTiles style (default in this repo) is taken from `subclass` which mean it takes icon -> [confectionery.svg ](https://github.com/openmaptiles/openmaptiles/blob/master/style/icons/confectionery.svg). This is also the correct icon for [OpenStreetMap Carto](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dconfectionery).
> 
> OpenMapTiles style image

I was talking about MapTiler's style, treat it like a side note, I actually shouldn't even mention this here as it's a third-party style.

TomPohys:
> This happened because of the classification of confectionery to ice_cream class instead of shop.

I am sorry, but I may not understand. All three styles (OSM Carto, MapTiler OpenStreetMap, and OpenMapTiles Style) show the same icon. The icon is not taken from the class attribute (now the `ice_cream` but let's move it to `shop`) but from the subclass attribute (`subclass=confectionery`).

michalgwo:
> > This happened because of the classification of confectionery to ice_cream class instead of shop.
> 
> I am sorry, but I may not understand. All three styles (OSM Carto, MapTiler OpenStreetMap, and OpenMapTiles Style) show the same icon. The icon is not taken from the class attribute (now the `ice_cream` but let's move it to `shop`) but from the subclass attribute (`subclass=confectionery`).

Thank you for merging. About this MapTiler style, sorry for the confusion, I'm terrible at explaining things xD 
So, the style and place I was talking about is [here](https://www.maptiler.com/maps/#style=openstreetmap&mode=2d&position=20/50.0370623/22.0227204). The icon is fine, what I meant was a different color of text and icon, which looks ugly. This happened in this style because confectionery fell into the Food layer (together with ice cream). Now, after the change (assuming MapTiler will update their schema), confectionery will fall into the Shop layer, which will change the text color, making it the same as the icon's color.

PR #1628, opened 2024-02-03
Changed classification of ice_cream and chocolate

michalgwo:
A few days ago I made a [pull request](https://github.com/openmaptiles/openmaptiles/pull/1627) about wrongly classified confectionery under the ice_cream class. Now I noticed the problem with the ice_cream class is a little deeper.

Firstly there is a chocolate subclass under the ice_cream class, which I think doesn't belong there (as OSM tag is [shop=chocolate](https://wiki.openstreetmap.org/wiki/Pl:Tag:shop%3Dchocolate)) and should be moved to the shop class, just like confectionery.

The second, more complicated thing is ice_cream, which appears in 2 places - as a subclass of the shop class and a separate class.

I guess it was added because we have 2 different OSM tags for ice_cream, which is [amenity=ice_cream](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dice_cream) and [shop=ice_cream](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dice_cream), but the problem is, that OpenMapTiles doesn't distinguish between amenity=ice_cream and shop=ice_cream, in both cases, it returns class=shop + subclass=ice_cream.

I think it makes sense to have just one ice_cream, so we can either delete the ice_cream class or the ice_cream subclass from the shop class.

According to [OpenMapTiles Schema](https://openmaptiles.org/schema/) ice_cream should be a class, not a subclass of the shop, so I think it makes more sense to remove ice_cream from the shop class. I tested this, and after the change, both amenity=ice_cream and shop=ice_cream OSM tags get class=ice_cream + subclass=ice_cream, which is fine, because both OSM tags shop=ice_cream and amenity=ice_cream represents the same thing

The next thing I did was to change the color of the ice_cream icon. The reason was that after the above changes, in the default OpenMapTiles style, ice_cream falls under the poi_food layer, which has a different text color from the shop, so to make it consistent I needed to change the color of the icon, which by the way matches now with OpenStreetMap Carto.

The last thing was changing the chocolate icon in the default layer to confectionery (this icon is suggested by [OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dchocolate)). One reason to do that was that it didn't have any icon before, the second reason was that after the above changes, chocolate fell into (in default style) poi_shop layer, which again changes the color of text, so it's better to make it consistent with the icon.


TomPohys (approved):

TomPohys:
Awesome! Thank you for digging deeper into this!

PR #1637, opened 2024-02-26
Fix housenumber_display.sql function

lazaa32:
Fixes `housenumber_display.sql` function and further improves https://github.com/openmaptiles/openmaptiles/pull/1632

Parsing text values that contain `;`, the string was split into array of values which were casted to `integer`. However, that fails for housenumber that exceed Postgres `int` range `-2147483648` to `+2147483647`.
There are such housenumber, e.g.: https://www.openstreetmap.org/way/1091331906#map=19/-3.82721/-38.59269


TomPohys (approved):

TomPohys:
Thanks!

PR #1642, opened 2024-03-12
More roads for z4 and z5.

TomPohys:
This PR adds more roads to the zoom 4 for the United Kingdom and Ireland. Adding more roads for the USA at zoom 5.

The United Kingdom and Ireland zoom 4 now

Screenshot from 2024-03-12 17-07-32

after
image

The USA zoom 5 now

image

after

image

### DIFF
--- a/layers/transportation/network_type.sql
+++ b/layers/transportation/network_type.sql
@@ -31,7 +31,11 @@ $$
         -- Canada
         'ca-transcanada', 'ca-provincial-arterial',
         -- United States
-        'us-interstate',
+        'us-interstate', 'us-highway',
+        -- UK
+        'gb-motorway', 'gb-trunk',
+        -- Ireland
+        'ie-motorway', 'ie-national',
         -- Europe
         'e-road',
         -- Asia
--- a/layers/transportation/update_transportation_merge.sql
+++ b/layers/transportation/update_transportation_merge.sql
@@ -807,7 +807,8 @@ BEGIN
             WHERE transportation.changes_z4_z5_z6_z7.is_old IS FALSE AND
                   transportation.changes_z4_z5_z6_z7.id = osm_transportation_merge_linestring_gen_z5.id
         )) AND
-        osm_national_network(network) AND
+        (highway = 'motorway' AND osm_national_network(network)
+        ) AND
         -- Current view: national-importance motorways and trunks
         ST_Length(geometry) > 1000
     ON CONFLICT (id) DO UPDATE SET osm_id = excluded.osm_id, highway = excluded.highway, network = excluded.network,


adamfranco:
Hi Tom, thanks for adding to low-zoom roads. I love seeing a broad networks at a continental scale!

In the US, we've finally gotten the majority of [the country's road network reclassified](https://wiki.openstreetmap.org/wiki/United_States/2021_Highway_Classification_Guidance) so that the combination of `highway=motorway`+`highway=trunk` is a good representation of the long-distance routes. There's certainly still work to be done, but most states are completed. 

While the `us-highway` network was historically the main route network in the US, today it is often superseded by newer State highways, expressways, and bypasses. Much of the `us-highway` network is rightfully tagged as `highway=primary` with some `highway=secondary` as well.

I think you will get a more connected national result by including `highway=motorway`+`highway=trunk` in the US rather than selecting for the `us-interstate` + `us-highway` networks at low zooms. The one downside of this approach is that this does include a higher density of non-`us-interstate` motorways in urban areas.

The [Americana rendering at z4](https://zelonewolf.github.io/openstreetmap-americana/#map=4/37.79/-97.46) which includes just the `us-interstate` network looks like your before shot:
Screenshot 2024-03-12 at 5 32 37 PM

The current [Americana rendering at z5](https://zelonewolf.github.io/openstreetmap-americana/#map=5/39.16/-96.91) includes `highway=motorway`+`highway=trunk` and shows the connected road network after the mapping updates of the past few years: 
Screenshot 2024-03-12 at 5 19 34 PM

I've found by zooming out in my browser to 50% and viewing Americana's `trunk`+`motorway` networks at z5 at continental scale that selecting `highway=motorway`+`highway=trunk` does well for most of the the world, including the rest of the Americas, and most of Africa, Asia, Australia, and much of Europe (UK, Ireland, Sweden, Norway, Finland, Spain, Romania, etc). 
Screenshot 2024-03-12 at 6 03 10 PM

The most significant places with disjoint `trunk`+`motorway` networks are the central European countries like France, Germany, Austria, etc. Also Algeria.
Screenshot 2024-03-12 at 6 09 39 PM

lazaa32 (approved):

TomPohys:
Thanks, @adamfranco for the analysis. The disjoined roads are not ideal, but they could be enhanced by adding more national networks that will fill up the segments (e.g. in France - https://www.openstreetmap.org/relation/14713871). But for now, it is OK, and `network=e-road` and `network=a-road` will do most of the visuals for z4 and z5. For the US, there is a condition for z4 that only motorways from `us-interstate` or `us-highway` could be there, and z5 is extended by `trunk` roads.

PR #1644, opened 2024-03-18
Fix maritime low zooms

lazaa32:
This PR fixes boundary layer at low zoom levels:

-z4 - remove maritime boundary (Peru, northern and southern end of boundary between Canada and Alaska).

Screenshot from 2024-03-18 09-17-45

-z1-z4 - do not mix disputed boundary from NE and OSM - linestrings got duplicated. Use only NE data (z0-z4), then OSM data (z5+)

image


TomPohys (approved):

TomPohys:
Thanks!

PR #1647, opened 2024-03-27
Fix borders z0-z4

lazaa32:
This PR fixes two things:
- remove disputed name like `ne_*m_ogc_fid` from lines
- use South Sudan disputed boundary from `ne_10m_admin_0_boundary_lines_land_disputed` so there is no gap.

Before: 
image

After:
image


TomPohys:
Thanks!

PR #1650, opened 2024-03-29
Unifying network names.

TomPohys:
This PR is for discussion, if the network names should be unified between attributes `network` and `route_X_network` based on function `osm_route_member_network_type(network, ref)`

Kansas (8/38.087/-97.756) (this PR)
image

vs (current)
image

### DIFF
--- a/layers/transportation_name/transportation_name.sql
+++ b/layers/transportation_name/transportation_name.sql
@@ -58,32 +58,32 @@ SELECT geometry,
            WHEN length(coalesce(ref, '')) > 0
                THEN 'road'
            END AS network,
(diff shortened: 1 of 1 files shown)


zekefarwell:
The[ OSM heirarchical network code format](https://wiki.openstreetmap.org/wiki/Key:network#Hierarchical_format) is [integral to the Americana style](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/src/js/shield_defs.js) so it would be prefereable to keep the original OSM `network` value in the `route_X_network` attributes.  If consistency between the attributes is desireable, perhaps the `network` attribute could also keep the original OSM `network` value rather than modifying it.

PR #1651, opened 2024-03-29
Better merging roads on upper zooms.

TomPohys:
For upper zoom levels (4-6) the roads are very segmented and sometimes there are two lines simplified on the same place but reversed (so still two lines).

This extra line merging with fewer attributes reduces the number of lines and stitches more touching segments into longer roads.

original table 
image

merged table

image


TomPohys:
Let's close this PR as it does not implement `updates` for zoom 4-6.

PR #1656, opened 2024-04-17
Fix z7-z4 roads

TomPohys:
Some US roads were not included in tiles, because the `ST_AsMVTGeom` did not transform geometry (result was NULL). This PR is trying to fix the geometry for `ST_AsMVTGeom`, even if the geometry is marked as valid (`ST_IsValid`). The odd geometry was caused by simplification when it could create `LINESTRING(1 1, 2 2, 1 1)` linestring, which in some cases was not converted into MVT. 

Also, some more complex lines were not transformed into MVT, in this case, the `ST_SnapToGrid` helps a lot without losing the precision up to zoom 7.

It is similar to https://github.com/openmaptiles/openmaptiles/pull/1651 but with working `DIFF`

Also, for the US (on zoom 4) keeps just the network `us-interstate` and the `highway=motorway` or `osm_national_network(network)` for the rest of the world.

Current
current

Fixed
new


lazaa32 (approved):
Great improvement, thanks.

Issue #1663, opened 2024-06-02
No Styles section in the tileserver landing page


iyan31:
Hello,

I tried this quickstart step `./quickstart.sh singapore`.  Next, I started the tileserver by `make start-tileserver`, in the landing page (or localhost:8080) I can only see Data section and I cannot see the Styles as shown below
<img width="1012" alt="Screenshot 2024-06-02 at 07 59 06" src="https://github.com/openmaptiles/openmaptiles/assets/701674/06ade9ad-0dd0-4842-bcc8-0ff4857002c8">

but when I tried to run the tileserver with this command `docker run -it -v $(pwd):/data -p 8000:8080 maptiler/tileserver-gl data/tiles.mbtiles`, I can get the Styles to show up on the landing page as shown below. 
<img width="835" alt="Screenshot 2024-06-02 at 08 12 13" src="https://github.com/openmaptiles/openmaptiles/assets/701674/8d2849b8-91c9-42ab-9167-8674ae9f6812">

What do I need to change in the make file or config so that when I start the tileserver by `make start-tileserver`, the Styles section can be visible too?

Thanks!

PR #1672, opened 2024-06-30
update transportation style filter format

redheli:
to fix https://github.com/openmaptiles/openmaptiles/issues/1663

the issue is transportation layer filter format not match [maplibre style spec](https://maplibre.org/maplibre-style-spec/deprecations/#combining-filters)

it cause tileserver-gl errors when parse the style.json:
```
The file "style.json" is not a valid style file:
undefined: layers[94].filter[2][1][0]: expected one of [==, !=, >, >=, <, <=, in, !in, all, any, none, has, !has], "match" found
undefined: layers[94].filter[2][2][0]: expected one of [==, !=, >, >=, <, <=, in, !in, all, any, none, has, !has], "match" found
undefined: layers[95].filter[2][1][0]: expected one of [==, !=, >, >=, <, <=, in, !in, all, any, none, has, !has], "match" found
undefined: layers[95].filter[2][2][0]: expected one of [==, !=, >, >=, <, <=, in, !in, all, any, none, has, !has], "match" found
```

note: to reproduce the errors, you need start the docker with entrypoitn bash and run the command `docker-entrypoint.sh` inside the docker

after fix:
image

### DIFF
--- a/layers/transportation/style.json
+++ b/layers/transportation/style.json
@@ -4535,29 +4535,15 @@
         [
           "all",
           [
-            "match",
-            [
-              "get",
-              "brunnel"
-            ],
-            [
-              "bridge",
-              "tunnel"
-            ],
-            false,
-            true
+            "in",
+            "brunnel",
+            "bridge",
+            "tunnel"
           ],
           [
-            "match",
-            [
-              "get",
-              "class"
-            ],
-            [
-              "minor"
-            ],
-            true,
-            false
+            "in",
+            "class",
+            "minor"
           ]
         ]
       ],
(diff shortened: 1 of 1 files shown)


layers/transportation/style.json
```diff
@@ -4637,29 +4623,15 @@
...
-              "class"
-            ],
-            [
-              "minor_construction"
-            ],
-            true,
-            false
+            "in",
+            "class",
+            "minor"
```
petraduriancikova, line 4634: Thanks for the fixes! One small thing - here should be "minor_construction" instead of "minor".

redheli: nice catch, updated the code to use `minor_construction`

petraduriancikova (approved):

PR #1661, opened 2024-05-17
Add landuse=flowerbed to landcover

nickswalker:
This adds `landuse=flowerbed` to landcover.

Closes #1660

PR #1680, opened 2024-08-19
Update CI to use docker compose v2.

TomPohys:
GitHub Actions and self-hosted server using `docker compose` v2 (without hyphen).


TomPohys:
As `HEAD is now at 10d1f75 Add landuse=flowerbed to landcover (#1661)` the Performance CI will fail.

Issue #1328, opened 2021-12-10
Include peaks without elevation


zstadler:
Currently, the `mountain_peak` layer excludes peak points with elevation.

https://github.com/openmaptiles/openmaptiles/blob/7d08e5b97e560efd783282d0e09fae4041fb84d7/layers/mountain_peak/mountain_peak.sql#L74-L76

This limits the value of the maps in areas where mapping data is not detailed as in other parts of the world. 
Unlike the default OSM maps, OMT-based maps cannot show peaks if their elevation is unknown.
I also believe that if OSM mappers have added a peak it is likely that this mapping is of value even if the elevation is unknown.

I've already implemented this in my fork and I'll be happy to send a PR.

TomPohys:
Hi, I understand that this is an old issue (and another even older https://github.com/openmaptiles/openmaptiles/issues/761). We should include peak points with a name and without elevation, e.g. as it is now including peak points with elevation without a name.

PR #1682, opened 2024-08-23
include mountain peaks without elevation, but with name tag

### LINKED ISSUE #761: Include all peaks in the `mountain_peak` layer
Only 65% of the peaks have an `ele` tag and therefore are included in the `mountain_peak` layer.
Some mappers, like the OSM default map, would like to show peaks even if the have no name or elevation. Peaks with names, for example, are valuable even of their elevation is unknown.

If all peaks are included in the `mountain_peak` layer, it would be up to the style creator to decide which peaks to show.

I will be happy to submit a PR.

Data source: [Overpass-Turbo query](http://overpass-turbo.eu/s/PvT) (click "Run" and be patient):
* 619k peaks
* 401k peaks with `ele` (65%)
* 142k peaks with `name` and without `ele` (23%)
* 76k peaks without `name` or `ele` (12%)

michalgwo:
Fixes #761 and #1328

According to the https://github.com/openmaptiles/openmaptiles/issues/1328#issuecomment-1463800079 I included peaks without elevation, but with a name.

### DIFF
--- a/layers/mountain_peak/mountain_peak.sql
+++ b/layers/mountain_peak/mountain_peak.sql
@@ -72,8 +72,10 @@ FROM (
                     )::int AS "rank"
          FROM peak_point
          WHERE geometry && bbox
-           AND ele IS NOT NULL
-           AND ele ~ E'^-?\\d{1,4}(\\D|$)'
+           AND (
+            (ele IS NOT NULL AND ele ~ E'^-?\\d{1,4}(\\D|$)')
+            OR name <> ''
+           )
      ) AS ranked_peaks
 WHERE zoom_level >= 7
   AND (rank <= 5 OR zoom_level >= 14)


TomPohys:
Hi, this adds peaks with names and without elevation, but makes them more prominent as their rank is `NULL`.

I would suggest replacing (just for nice preview, without yours `OR name <> ''`)

https://github.com/openmaptiles/openmaptiles/blob/7cf98a62938148703449bc7374da5af532f2b470/layers/mountain_peak/mountain_peak.sql#L56-L77

with

```SQL
SELECT osm_id,
                geometry,
                NULLIF(name, '') as name,
                COALESCE(NULLIF(name_en, ''), NULLIF(name, '')) AS name_en,
                COALESCE(NULLIF(name_de, ''), NULLIF(name, ''), NULLIF(name_en, '')) AS name_de,
                tags,
                substring(ele FROM E'^(-?\\d+)(\\D|$)')::int AS ele,
                round(substring(ele FROM E'^(-?\\d+)(\\D|$)')::int * 3.2808399)::int AS ele_ft,
                CASE WHEN iso_a2 = 'US' THEN 1 END AS customary_ft,
                row_number() OVER (
                    PARTITION BY LabelGrid(geometry, 100 * pixel_width)
                    ORDER BY (
                            (CASE WHEN ele <> '' THEN substring(ele FROM E'^(-?\\d+)(\\D|$)')::int ELSE 0 END) +
                            (CASE WHEN wikipedia <> '' THEN 10000 ELSE 0 END) +
                            (CASE WHEN name <> '' THEN 10000 ELSE 0 END)
                        ) DESC
                    )::int AS "rank"
         FROM peak_point
         WHERE geometry && bbox
           AND (
            (ele <> '' AND ele ~ E'^-?\\d{1,4}(\\D|$)')
            OR name <> ''
           )
     ) AS ranked_peaks
```

michalgwo:
Thank you for the review, I just committed the proposed changes ;)

TomPohys (approved):
Thanks!

PR #1689, opened 2024-09-29
Add Toki Pona to languages

tbodt:
Would like to add this language not because it is widely spoken (there are not a super large number of labels in this language in OSM (though many more on Wikidata)) but because I want to make a website with a map in it.


TomPohys (approved):
Thanks!

PR #1690, opened 2024-10-15
Undo inverted brunnel filter for minor roads

mhernz:
Re PR #1672: accidentally hides all minor roads except when brunnels


petraduriancikova:
Many thanks for the catch, Matt! I accidentally overlooked this.

TomPohys:
Thanks!

PR #1698, opened 2024-12-03
vacuum-db:  use parenthesis for  vacuum options

PatrikSylve:
Updated `vacuum-db` target to have the vacuum options within parenthesis to avoid syntax error. 

```sql
syntax error at or near "VERBOSE"
LINE 1: VACUUM ANALYZE VERBOSE;
```

When not using parenthesis, the options had to be ordered as listed in https://www.postgresql.org/docs/14/sql-vacuum.html.


TomPohys (approved):
Thanks! Good catch!

PR #1704, opened 2025-02-07
Improve getmvt performance on lower zooms (transportaion & poi)

PatrikSylve:
I noticed slow performance in `getmvt` when generating tiles at zoom levels ~ 0-8.

The issue was due to CASE blocks in the `transportation` and `poi` layers, which were processing unneccessary rows at lower zoom levels. To optimize this I added a pre-filter on `zoom_level` to reduce the workload. 

In my tests, I timed the query `select getmvt(0,0,0);`. 

`area=europe/united-kingdom`: 

* Before: ~17 seconds
* With this fix:  ~80 ms

 `area=planet`:
 * Before: have not yet timed it, but i believe it was around 15min if i remember correctly
 * With this fix: ~10 seconds

### DIFF
--- a/layers/poi/poi.sql
+++ b/layers/poi/poi.sql
@@ -76,7 +76,7 @@ FROM (
                     ELSE osm_id * 10 + 1
                     END AS osm_id_hash
          FROM osm_poi_polygon
-         WHERE geometry && bbox AND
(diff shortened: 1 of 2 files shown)


TomPohys (approved):
Awesome! Thanks!

Issue #1618, opened 2024-01-19
Add information necessary to implement Trails Working Group rendering guidelines


dschep:
It's currently not possible to implement the [Trail Working Group's rendering suggestions](https://wiki.openstreetmap.org/wiki/United_States/Trail_Access_Project#Suggested_rendering_and_recommendations_to_data_consumers) with OpenMapTiles because it doesn't expose the `informal` or `operator` tags.

The easiest way to support this would of course be to just pass these through to tiles directly (probably only for `class=path`).

But I could see creating a combined field sort of like `brunnel`, in part because the specific operator probably doesn't matter for renderers, maybe just `informal` and set `informal=no` when an `operator` tag is set or just the value of `informal` if that's present. And omit the field when there exists no `informal` or `operator` tag on the way.

I chatted with @ZeLonewolf about this a little, and he suggested maybe something in `class` or `subclass` but I don't know how OMT feels about breaking changes, given those are already populated for these sorts of ways.

TomPohys:
Hi, I think that the easiest way could be just to extend the `access` attribute to "if `access=no` or `access=private` or `informal=yes`" then the value will be `access=no`

dschep:
IMO that's too little information. You certainly couldn't implement the example rendering tree linked.

pnorman:
I was going to write something about the tree, but then I noticed the top of the page linked

> The scope of this project is US public access lands administered by the US Government. However, the principles in this project plan may be applicable to other public lands at the state and local level

OpenMapTiles needs to work everywhere. Their recommendations are built around paths with Land Managers, while most paths don't have a Land Manager in the sense they are talking about. I suspect most footpaths are far more similar to a sidewalk than a footpath on public lands, but would welcome an analysis.

I think two attributes are the best method of doing this. One for access with three possible values - yes (including plain highway=footway with no other tags), sort of, and no access. The other would be surface, also with three possible values - paved, unpaved, or unknown. The cartographer making a style could then choose if they wanted emphasis on access or surface, depending on their needs.

ZeLonewolf:
I am presently hosting a trails vector layer for OSM US's trail project (the working group and the OpenTrailMap.us map).  It's accessible here for anyone that wants to play with it: https://d1zqyi8v6vm8p9.cloudfront.net/trails.json

That's outside of OpenMapTiles of course, but it's fairly easy to generate custom layers for this kind of thing in planetiler.

[...]

PR #1707, opened 2025-02-18
Add 'official' boolean property to foot paths

### LINKED ISSUE #1706: Add 'official' boolean to foot paths in transportation layer
The [OSM US Trails Working Group](https://wiki.openstreetmap.org/wiki/United_States/Trails_Stewardship_Initiative) has been working to add missing attributes to hiking trails (and other recreational trails) in OpenStreetMap, including adding information about whether a trail is official or unofficial. An official trail is one that is built/maintained by the land owner or land management agency; unofficial trails are things like social trails (desire paths) that arise organically, or trails that were built without the land owner's permission. The goal of this effort is to help map makers symbolize this distiction on their maps, so that people recreating in the outdoors can make more informed decisions about where to go, and hopefully decrease the safety and environmental risks associated with misuse of unofficial trails.

There are two tags in OSM that tell a data consumer whether a trail is official or not:
- `operator=*`, which states the name of the entity (usually a government agency, land trust, etc) that oversees and maintains the trail. If a trail has an `operator`, then it is an official trail.
- `informal=yes`, which states that the trail has no operator and is therefore an unofficial trail. (In rare cases where a trail is known to be official but the operator's name is not known, `informal=no` may also be used).

It would be great if OpenMapTiles included an `official` boolean property on foot paths in the `transportation` layer which was derived from these two tags. This would help users of OpenMapTiles make maps that follow the Trails Working Group's [recommendations](https://wiki.openstreetmap.org/wiki/United_States/Trails_Stewardship_Initiative#Suggested_rendering_and_recommendations_to_data_consumers) for renderers, which suggest visually emphasizing official trails and de-emphasizing unofficial trails.

I am happy to make a PR for this feature if it's desired.

jake-low:
Closes #1706.

This adds a field called `official` to `highway={path,footway,cycleway,bridleway}` features in the `transportation` layer. The field's value is:
- `0` if the OSM element is tagged `informal=yes`
- `1` if the feature has an `operator` in OSM (or is tagged `informal=no`)
- unset otherwise (this is the large majority of cases in the data today, especially when considering sidewalks, bike lanes, etc that aren't typically tagged with an `operator`).

The field is present only at 

Questions I have about whether this PR is correct:
1. The `official` field is conceptually an optional boolean. I chose to make it an `int` because I saw that existing boolean fields in the tiles (e.g. `intermittent` on waterways, `toll` on roads) have been implemented as 0/1 integers. Is this the right type for this field?
2. I added `description` and `values` fields to the transportation layer's `mapping.yaml` file. Is there anything else I should do to document this new field? Not sure how the schema docs on opentrailmap.org are built, but I am assuming they're generated from these mapping files somehow.

---

Here's a screenshot of a demo map style I whipped up (by mild tweaking of OSM Liberty) as an example of how this field helps make better trail maps.

On the left is a trail map where no distinction is made between official an unofficial trails. In the area shown, there is a dense braided network of social trails which are causing damage to the alpine meadow environment.

On the right is the same map but with official trails emphasized and unofficial ones de-emphasized. This can hopefully help hikers who want to stick to the official trails navigate the area responsibly.

Screenshot 2025-02-17 at 10 31 56 PM
Screenshot 2025-02-17 at 10 32 01 PM

### DIFF
--- a/layers/transportation/mapping.yaml
+++ b/layers/transportation/mapping.yaml
@@ -193,6 +193,14 @@ sac_scale_field: &sac_scale
   key: sac_scale
   name: sac_scale
   type: string
+operator_field: &operator
+  key: operator
+  name: operator
+  type: string
+informal_field: &informal
+  key: informal
+  name: informal
+  type: string
 surface_field: &surface
   key: surface
   name: surface
@@ -249,6 +257,8 @@ tables:
     - *horse
     - *mtb_scale
     - *sac_scale
+    - *operator
+    - *informal
     - *surface
     - *expressway
     mapping:
--- a/layers/transportation/transportation.sql
+++ b/layers/transportation/transportation.sql
@@ -30,6 +30,7 @@ CREATE OR REPLACE FUNCTION layer_transportation(bbox geometry, zoom_level int)
                 foot       text,
                 horse      text,
                 mtb_scale  text,
+                official   int,
                 surface    text
             )
 AS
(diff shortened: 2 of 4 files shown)


jake-low:
Made some changes that should hopefully fix the `import-diff` tests that were failing

dschep:
this also closes #1618

jake-low:
Done. Thanks for fixing this.

PR #1708, opened 2025-02-18
Use pg_trigger_depth to avoid re-trigger in update

PatrikSylve:
Some of the `update_osm_${LAYER}`-functions, which are executed by triggers on updates, execute an UPDATE statement on the same tables that have these triggers. This causes the trigger functions `flag` and `store` to run multiple times for one record. 

For instance, if I add log in the trigger functions and run an INSERT on `osm_housenumber_point`, this output is generated:  

```
NOTICE:  Store
NOTICE:  Flag
NOTICE:  Refresh housenumber
NOTICE:  Flag
NOTICE:  Store
NOTICE:  Flag
INSERT 0 1
```

If we limit the triggers from executing if they are called from another trigger using `pg_trigger_depth() < 1`, the triggers will only be triggered once per record:

```
NOTICE:  Store
NOTICE:  Flag
NOTICE:  Refresh housenumber
INSERT 0 1
```

This will prevent redunant executions and might improve update performance.

### DIFF
--- a/layers/aerodrome_label/update_aerodrome_label_point.sql
+++ b/layers/aerodrome_label/update_aerodrome_label_point.sql
@@ -93,12 +93,14 @@ CREATE TRIGGER trigger_store
     AFTER INSERT OR UPDATE
     ON osm_aerodrome_label_point
     FOR EACH ROW
+    WHEN (pg_trigger_depth() < 1)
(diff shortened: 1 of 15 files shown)


TomPohys (approved):
Awesome! Thank you very much!

PR #1710, opened 2025-02-26
Set columns to null instead of dropping to avoid pg_deadlock during update

PatrikSylve:
This PR addresses a pg_deadlock error that would cause an update to fail if a `select getmvt` was queried during update. The deadlock occurred due to an ALTER TABLE DROP COLUMN operation on temporary columns  ( `new_source_ids` and `old_source_ids`) in some linestring tables.

**Changes:**
*  instead of adding/dropping the columns during update, they are now updated and set to NULL. 
* Added new indexes on the source_ids columns  to speed up the UPDATE. 
* Will refresh `osm_park_polygon_dissolve_z4` concurrently in `park_polygon.refresh()` to avoid blocking.

### DIFF
--- a/layers/park/update_park_polygon.sql
+++ b/layers/park/update_park_polygon.sql
@@ -149,7 +149,7 @@ BEGIN
 
     -- Analyze tracking and source tables before performing update
     ANALYZE osm_park_polygon_gen_z4;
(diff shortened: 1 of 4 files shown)


TomPohys (approved):
Looks good! Thanks!

PR #1718, opened 2025-06-18
Fix QUIET var handling in Makefile

### LINKED ISSUE #1717: --quiet flag used by default for docker-compose pull
First of all, thank you for this nice tool well documented and easy to setup for a newcomer!

The [QUICKSTART.md](https://github.com/openmaptiles/openmaptiles/blob/f70ae783b286e362431efc53a07dfa35472be2bd/QUICKSTART.md#L18) mentions that docker-compose version 1.7.1 or newer is needed.

But `./quickstart.sh` makes the Makefile set the `--quiet` flag to `docker-compose pull` by default here:

https://github.com/openmaptiles/openmaptiles/blob/f70ae783b286e362431efc53a07dfa35472be2bd/Makefile#L47-L52

I have tried to run it on an old server with docker-compose version 1.10.0 and it fails.
It seems that the `--quiet` flag has been introduced in version 1.19: https://github.com/docker/compose/commit/e9b6cc23fcf01d4768c7e082b7bc91b43ff84e7e

romainreignier:
The brackets were unbalanced and the condition always evaluated to false so the --quiet flag was always added.

Fix #1717

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ else
 endif
 
 # Make some operations quieter (e.g. inside the test script)
-ifeq ($(or $(QUIET),$(shell (. .env; echo $${QUIET})))),)
+ifeq ($(or $(QUIET),$(shell (. .env; echo $${QUIET}))),)
   QUIET_FLAG :=
 else
   QUIET_FLAG := --quiet


TomPohys (approved):
Awesome! Thanks!

PR #1728, opened 2025-07-28
Update style

lazaa32:
Update style snippets with latest changes.
The style should now be the same as at the https://www.openstreetmap.org#layers=V


TomPohys (approved):
Awesome! Thank you!

PR #1738, opened 2025-08-16
Update font file downloaded by "make download-fonts"

nhirokinet:
Hi, thank you for nice OSS product.

I think I can make download-fonts appropriate by this change, could you take a look and merge if in my understanding is correct?

There are noto-sans.zip and v2.0.zip available in font repo tag v2.0. https://github.com/openmaptiles/fonts/releases/tag/v2.0

Currently this code downloads noto-sans.zip, but v2.0.zip looks like more comprehensive one which includes Noto Sans.

In my understanding noto-sans.zip is not enough and v2.0.zip is appropriate for openmaptiles style.

# Pull request guidelines

Before working on a pull request, create an issue explaining what you want to contribute. This ensures that your pull request won't go unnoticed, and that you are not contributing something that is not suitable for the project. Once a core developer has set the `help wanted` label on the issue, you can submit a pull request. The pull request description should reference the original issue.

Your pull request must:

 * Address a single issue or add a single item of functionality.

 * Contain a clean history of small, incremental, logically separate commits,
   with no merge commits.

 * Use clear commit messages.

 * Be possible to merge automatically.

When you are making a PR that adds new spatial features to OpenMapTiles schema, please make also PR for at least one of our GL styles to show it on the map. Visual check is crucial.


ZeLonewolf:
What does this PR intend to fix?

nhirokinet:
Currently, `make download-fonts` looks like downloads Noto Sans fonts only.
But actually, openmaptiles required other fonts as well.

Thus, by following current `Makefile`, openmaptiles starts with lack of fonts like Open Sans, which lead to issue of displaying.

This patch is intended to fix the problem that displaying issue happens if started just by `Makefile`, by downloading all fonts including Noto Sans or Open Sans, not only Noto Sans.

TomPohys:
Hi, sorry for the delay. I will modify the openmaptiles/font v2.0 release and add a new zip file with only the fonts used in the OpenMapTiles style. So when you build an OMT style, you will get all the fonts, but you won't download the ones you don't need.

nhirokinet:
I see, thanks. Then shall I close this pull request?

TomPohys:
I will add another zip file with a different name to the release, so if you change the name of the new zip file in the makefile, this PR can stay. Thank you.

I will let you know when it's ready.

TomPohys:
Hi, I uploaded the fonts into the release with the name `noto-open-sans`.

https://github.com/openmaptiles/fonts/releases/download/v2.0/noto-open-sans.zip

Makefile
```diff
@@ -308,9 +308,9 @@ build-style: init-dirs
 .PHONY: download-fonts
 download-fonts:
 	$(DOCKER_COMPOSE) run $(DC_OPTS) openmaptiles-tools bash -c '[ ! -d "/export/fonts" ] && mkdir /export/fonts && \
-		echo "Downloading fonts..." && wget -qO /export/noto-sans.zip --show-progress \
-		https://github.com/openmaptiles/fonts/releases/download/v2.0/noto-sans.zip && \
-		echo "Unzipping fonts..." && unzip -q /export/noto-sans.zip -d /export/fonts && rm /export/noto-sans.zip || \
+		echo "Downloading fonts..." && wget -qO /export/v2.0.zip --show-progress \
```
TomPohys:
```suggestion
		echo "Downloading fonts..." && wget -qO /export/noto-open-sans.zip --show-progress \
```

Makefile
```diff
@@ -308,9 +308,9 @@ build-style: init-dirs
...
+		https://github.com/openmaptiles/fonts/releases/download/v2.0/v2.0.zip && \
```
TomPohys:
```suggestion
		https://github.com/openmaptiles/fonts/releases/download/v2.0/noto-open-sans.zip && \
```

Makefile
```diff
@@ -308,9 +308,9 @@ build-style: init-dirs
...
+		echo "Unzipping fonts..." && unzip -q /export/v2.0.zip -d /export/fonts && rm /export/v2.0.zip || \
```
TomPohys:
```suggestion
		echo "Unzipping fonts..." && unzip -q /export/noto-open-sans.zip -d /export/fonts && rm /export/noto-open-sans.zip || \
```

TomPohys (commented):
This should now download all necessary fonts for OMT style!

nhirokinet:
Thank you! I imported your suggestions.

TomPohys (approved):
Thanks for pointing this out. The style is now better :-)

PR #1746, opened 2025-09-04
Parking Areas in Transportation

GA-Kevin-Codes:
Discussion from #1438 and #1437 adding parking areas to the transportation layer (similar to plazas).


GA-Kevin-Codes:
1438 adds to landuse not transportation.

ZeLonewolf:
Yes, past me analyzed that poorly - I think we can just push on #1438

Issue #1750, opened 2025-09-08
Orientation of single letter in road name label does not match orientation of label itself


nc011:
At certain zoom levels (18-21), in one specific road label, the last letter of a word is not being rotated to match the orientation of the label itself.

**Link:** 
https://www.openstreetmap.org/#map=19/52.448460/-1.819997&layers=V

**Image:**
<img width="842" height="733" alt="Image" src="https://github.com/user-attachments/assets/23d9d5f9-ab16-4f33-adb3-3979882d6304" />

The "d" at the end of "Oxford" in that one specific label has not been rotated to match the orientation of the road and the rest of the label. I thought this might just be a random artifact but I'm seeing it across different devices and browsers.

Viewing through the OSM homepage but can also see the same artifact here: https://www.maptiler.com/maps/#style=openstreetmap&lang=auto&mode=2d&position=17.66/52.448581/-1.819795

The road is one long continuous way, so it's not that. Though there is a speed bump mapped just after where that label appears.

PR #1751, opened 2025-09-11
Fix road labels letter orientation

petraduriancikova:
Fixes the bad orientation of a letter reported in https://github.com/openmaptiles/openmaptiles/issues/1750. Fixes any other potential weird letter rendering from zoom level 17 to 20 by making the symbol-spacing bigger at these zooms.


TomPohys (approved):
Awesome! Thanks!

PR #1767, opened 2025-10-25
[BUGFIX] Fix dropped road segments at z4-5

ZeLonewolf:
This PR fixes an issue which is seen in the zoom 4 and 5 road network, where roads that switch between `trunk` and `motorway` cause intermediate segments to drop, causing a dashed line effect:

image

The reason this is happening, is that we have logic for dropping small segments.

This PR fixes this by first upgrading any small `trunk` segments to motorway at z5, then performing a merge, and then dropping small segments.  This will cause legitimate small segments to get dropped while blending small `trunk` segments into the larger motorway network, which is appropriate at these small zooms.

This is my test rendering of Iowa, USA, showing that the dashed-line effect no longer occurs:

image

The OpenMapTiles style does not show roads at these zooms, so there is no impact on the built-in style.

### DIFF
--- a/.github/workflows/integrity.yml
+++ b/.github/workflows/integrity.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
 
       - name: Checkout the changes
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Run quickstart for a small area
         env:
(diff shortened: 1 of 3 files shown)


ZeLonewolf:
@TomPohys, PR submitted to planetiler-openmaptiles for this feature!
https://github.com/openmaptiles/planetiler-openmaptiles/pull/289

PR #1798, opened 2026-05-09
added parking area to landuse

### LINKED ISSUE #1437: Parking is a way/polygon, but schema only display as a point icon

michalgwo:
Fixes #1437 

As #1438 has been inactive for a while, I thought I would make a replacement PR, solving all the issues.

This PR adds parking polygons to landuse with parking (as a subclass) and access tags.

I think capping it to a specific zoom is not needed, as landuse seems to have some logic to filter out small polygons. I didn't really found this in code, but smaller polygons seems to disappear. For example, [this parking](https://www.openstreetmap.org/way/85511474) is present only on zoom >= 12, but [this parking](https://www.openstreetmap.org/way/317269927) is present only on zoom >= 14

This is how the first parking looks like on zoom 12:
image
This is how it looks like when the parking is hidden:
image

The second screenshot looks pretty empty to me, I think it still makes sense to show the parking in this case. Anyway, most of the parkings are smaller, so they will probably be capped to zooms 13 or 14.


layers/landuse/landuse.sql
```diff
@@ -369,11 +425,13 @@ FROM (
...
+                parking,
+                access
          FROM osm_landuse_polygon
          WHERE zoom_level >= 14
      ) AS zoom_levels
 WHERE geometry && bbox;
 $$ LANGUAGE SQL STABLE
                 -- STRICT
-                PARALLEL SAFE;
+                PARALLEL SAFE;
```
TomPohys:
```suggestion
                PARALLEL SAFE;

```

layers/landuse/landuse.yaml
```diff
@@ -45,13 +45,28 @@ layer:
...
       - dam
       - quarry
+      - parking
+    subclass:
+      description: |
+          Original value of the
+          [`parking`](http://wiki.openstreetmap.org/wiki/Key:parking) tag.
+          Use this to do more precise styling.
+      values:
+      - parking
```
TomPohys:
```suggestion

```

layers/landuse/landuse.yaml
```diff
@@ -45,13 +45,28 @@ layer:
...
-    query: (SELECT geometry, class FROM layer_landuse(!bbox!, z(!scale_denominator!))) AS t
+    query: (SELECT geometry, class, subclass, access FROM layer_landuse(!bbox!, z(!scale_denominator!))) AS t
 schema:
   - ./class.sql
   - ./prep_landuse.sql
   - ./landuse.sql
 datasources:
   - type: imposm3
-    mapping_file: ./mapping.yaml
+    mapping_file: ./mapping.yaml
```
TomPohys:
```suggestion
    mapping_file: ./mapping.yaml

```

layers/landuse/mapping.yaml
```diff
@@ -103,4 +110,4 @@ tables:
       - quarter
       - neighbourhood
       waterway:
-      - dam
+      - dam
```
TomPohys:
```suggestion
      - dam

```

TomPohys (changes requested):
Hi @michalgwo, Thanks for the PR, and sorry for such a delay. 

Just a small tune-up:

1) I would keep the last empty line, just for the sake of consistency

2) in landuse.yaml, I would not include `values=parking` as these values will be like `surface`, `underground`, `multi-storey`, `rooftop`, ... so it is not necessary to include `values` or extend the list from https://taginfo.openstreetmap.org/keys/parking#values, which will be converted into  `possible values:` in e.g., https://openmaptiles.org/schema/#landuse

Thanks

TomPohys (approved):
Awesome! Thanks!

PR #1807, opened 2026-07-09
Add rock, clay, laterite to surface_value() unpaved classification

### LINKED ISSUE #1806: surface_value() missing rock, laterite, clay from paved/unpaved classification
[\`surface_value()\`](https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/class.sql#L52-L54) classifies `surface` into `paved`/`unpaved`, three fairly common values fall through unclassified (NULL) in both this SQL and the matching [SURFACE_UNPAVED_VALUES/SURFACE_PAVED_VALUES sets in planetiler-openmaptiles's Transportation.java](https://github.com/openmaptiles/planetiler-openmaptiles/blob/main/src/main/java/org/openmaptiles/layers/Transportation.java):

- `rock`: 34,201 uses (taginfo), unpaved
- `clay`: 42,844 uses, unpaved
- `laterite`: 2,250 uses, unpaved. Documented via OSM's approved [Proposal:Surface=laterite](https://wiki.openstreetmap.org/wiki/Proposal:Surface%3Dlaterite) (29-0-0)

All three are higher usage than several values already in the unpaved list (`salt` at 511, `snow` at 508), so this isn't a usage-threshold question, just values that were missed when the lists were last put together.

Came up via [osm-americana/openstreetmap-americana#1434](https://github.com/osm-americana/openstreetmap-americana/issues/1434), where a consuming style's `isUnpaved` check turned out to be correct, the actual gap traces back to this classification.

julcnx:
## Summary
`surface_value()` in `layers/transportation/class.sql` classifies `surface` into `paved`/`unpaved`. Three fairly common values fall through unclassified (NULL) today:

- `rock`: 34,201 uses (taginfo)
- `clay`: 42,844 uses
- `laterite`: 2,250 uses, documented via OSM's approved [Proposal:Surface=laterite](https://wiki.openstreetmap.org/wiki/Proposal:Surface%3Dlaterite) (29-0-0)

All three exceed the usage of several values already in the unpaved list (`salt` at 511, `snow` at 508), so this isn't a usage-threshold question, they were simply missed when the lists were last assembled. Also updated the field description in `transportation.yaml` to match, per CONTRIBUTING.md.

Fixes #1806

## Test plan
- [ ] Recommended per CONTRIBUTING.md to add a SQL unit test for this, but that requires the full Postgres/osm2pgsql import harness (see TESTING.md), which I couldn't stand up in this environment. The change itself just extends an existing literal-value `IN (...)` list with the same pattern as the 17 values already there, relying on CI/review here.

### DIFF
--- a/layers/transportation/class.sql
+++ b/layers/transportation/class.sql
@@ -50,7 +50,8 @@ SELECT CASE
            WHEN surface IN ('paved', 'asphalt', 'cobblestone', 'concrete', 'concrete:lanes', 'concrete:plates', 'metal',
                             'paving_stones', 'sett', 'unhewn_cobblestone', 'wood', 'grade1') THEN 'paved'
            WHEN surface IN ('unpaved', 'compacted', 'dirt', 'earth', 'fine_gravel', 'grass', 'grass_paver', 'gravel',
-                            'gravel_turf', 'ground', 'ice', 'mud', 'pebblestone', 'salt', 'sand', 'snow', 'woodchips')
+                            'gravel_turf', 'ground', 'ice', 'mud', 'pebblestone', 'salt', 'sand', 'snow', 'woodchips',
+                            'rock', 'clay', 'laterite')
                THEN 'unpaved'
            END;
 $$ LANGUAGE SQL IMMUTABLE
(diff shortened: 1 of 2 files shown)


TomPohys (approved):
Thanks @julcnx for the PR! It looks good!

Issue #1587, opened 2023-10-12
New ids are missing encoded OSM type


zbycz:
Hi @TomPohys, 👋 

Since yesterday i am seeing that large number of layers with OSM IDs are not encoding the OSM type properly.
It used to be `$OSMID+'0'` = node, `$OSMID+'1'` = way, `$OSMID+'4'` = relation on most of the layers (i can tell you which), but now the vector feature `id` is just `$OSMID`.

Is this some known change? or a mistake? Can you suggest some alternative approach perhaps?

It broke the clickable poi map on [osmapp.org](https://osmapp.org/#15.72/40.5880/-73.6605), it tries to remove the last number in vector feature id and use it as osm type, so almost none pois are clickable anymore. City names still works the old way, but almost all pois are wrong. (I suspect cache is making it stilll work in CZ)

TomPohys:
Thanks, I have to check it!

zbycz:
Hi @TomPohys - could I prehaps help you somehow? :) I looked at the OMT repo, but there are no related changes. I guess that has to be bronek at the comerecial maptiler stack, right?

TomPohys:
Hi, sorry for the delay! The glitch in MapTiler infrastructure was identified and fixed. POI with OSM ID should be OK now. This bug has no relation to the OMT schema. Sorry for the inconvenience and thank you for pointing this out.

zbycz:
Hi @TomPohys, no worries. Thank you very much for resolving it. 🙂  I will close the issue once I see the correct data again (some cache is probably in place).

TomPohys:
Thanks! Yes, there is a cache, but the [point in US](https://osmapp.org/node/1089053983#15.72/40.5880/-73.6606) works.

zbycz:
Oh-la-la! Somewhere it really works 😃

zbycz:
HI @TomPohys, it seems that most of sourceLayers are fixed now, but there are still some returning the ID without type. I tried it both in CZ and US, it seems that cache isn't the issue now. Do you think you can look at those please?

Here is the list (i will edit if i find more):
- `sourceLayer:place` - click any name of a city or part
- `sourceLayer:building` - can't find example now
- `sourceLayer:poi, class:castle` - here [palác](https://osmapp.org/50.09030,14.39722)
- `sourceLayer:poi, class:garden` - here [botanická](https://osmapp.org/50.11872,14.41567)

-------

[...]

Issue #241, opened 2017-05-05
Missing Building Names


bricechart:
While setting up a style I realized the building names are not listed in the schema.  Could these be added?

lukasmartinelli:
> While setting up a style I realized the building names are not listed in the schema. Could these be added?

They could be added as anonymous POIs I am not sure whether it makes sense to add a name column to building.

bricechart:
> They could be added as anonymous POIs I am not sure whether it makes sense to add a name column to building.

How would I go about doing this?  I'm designing a style using the maputnik.com/editor/.  Any help is greatly appreciated.

garza:
the building should have both a name and ref, common use case would be several buildings on a university campus or large corporation

claustres:
We would like to know if you would accept a PR on this if we'd like to propose it. It seems that the openmaptiles team has some reluctance to include it because of the amount of data. According to your experience what is the expected additional size to include these labels ?

MA-MacDonald:
Any change on this? It would be nice to see building names for large campuses like Universities, Hospitals, etc..

<img width="756" height="813" alt="Image" src="https://github.com/user-attachments/assets/9703093e-7a5c-4f58-b7a3-9812eaeeb21b" />

amenk:
Claude and I would have a draft here:

https://github.com/OpenPlaceGuide/openmaptiles/commit/eecfaeae8032b383e38f640394097438deafb1f2

It works well in my tests.

Shall I make a pull request? Does this have chances to be included?
And the OpenStreetMap.org default style also includes building names, so it seems legit to include them in the Schema, too.

PR #1809, opened 2026-07-25
Encode the OSM object type in building layer feature ids

amenk:
The poi, place and water_name layers publish `osm_id * 10 + type_digit`, do so in the building layer, too

See #1587

### DIFF
--- a/layers/building/building.sql
+++ b/layers/building/building.sql
@@ -56,7 +56,7 @@ CREATE OR REPLACE FUNCTION layer_building(bbox geometry, zoom_level int)
 AS
 $$
 SELECT geometry,
-       osm_id,
+       CASE WHEN osm_id < 0 THEN -osm_id * 10 + 4 ELSE osm_id * 10 + 1 END AS osm_id,
        render_height,
        render_min_height,
        COALESCE(colour, CASE material


TomPohys:
Hi, the `poi`, `place`, and `water_name` could be created from node/way/area. That's why there is a distinction in `osm_id`. Do you have some examples where the building ID (`osm_id` attribute) is not correct? 

I would like to avoid creating these breaking changes.

amenk:
I assumed that building layer can be created from a node and a way, so causing ambiguity - but I am not sure. If it's always ways, we can probably close this and just use a little bit of extra code when parsing IDs.

TomPohys:
In the `tilestats` in `tiles.json` for the `building` layer, there are only `polygon` features. So I assume that the `mvt_id` value is 1:1 with the OSM, as there are no points or linestrings.

amenk:
You are right - there can not be nodes here, because we query `building_polygon`

But maybe we can have relations (type digit = 4)

My motivation for the PR was consistency with the other layers, where we have the type-digit.

If we only have relations (if even) and ways here, they are well distinguishable via negativity / positivity of the ID which is fine, but not so consistent.

I am okay with both.

But still we should document the encoding per layer.

TomPohys:
From the relation is taken the `member id`, so there should really be pure polygons with their OSM ID. I am in favor of documentation. I just do not know the perfect place for it. It could be part of the README of each layer, e.g., https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/README.md. 


---

https://github.com/openmaptiles/openmaptiles/blob/a8a0eefb8263709b719b55616e350c812986bd1f/layers/building/mapping.yaml#L152-L153

https://github.com/openmaptiles/openmaptiles/blob/a8a0eefb8263709b719b55616e350c812986bd1f/layers/building/building.sql#L12


It is also mentioned here: https://github.com/openmaptiles/www.openmaptiles.org/issues/85

amenk:
Okay, so we can close this for now and add documentation "later" :-)

Motivation was https://github.com/openmaptiles/openmaptiles/issues/241 because when adding names, the buildings should also be clickable in interfaces like OsmApp

PR #1810, opened 2026-07-26
Add shop=farm to POI classification

### LINKED ISSUE #1796: Add support for `shop=farm` in the POI classification
Hello,

I would like to propose adding support for the OpenStreetMap tag `shop=farm` in the POI schema.

### Context

I am a regular user of [OsmAPP](https://osmapp.org/):

I recently opened a feature request there to display farm shops (local producers selling directly to consumers):
https://github.com/zbycz/osmapp/issues/1462

The maintainer explained that this is currently not possible because the upstream vector tile providers do not include `shop=farm` in their tiles.

Because many applications rely on the OpenMapTiles schema, the absence of this tag at the schema level prevents these locations from being displayed in many maps.

### About `shop=farm`

`shop=farm` is a documented and widely used OpenStreetMap tag for:

* farm shops
* local producers selling directly
* short food supply chains

Documentation:
https://wiki.openstreetmap.org/wiki/Tag:shop=farm

These places are increasingly important for local food systems and sustainable agriculture.

### Proposal

Add `shop=farm` to the POI mapping so it appears in the `poi` layer, for example as:

* class: `shop`
* subclass: `farm`

or mapped to a similar retail category if preferred.

Even if map styles do not display it by default, including it in the tiles would allow downstream applications to decide whether to render it.

### Why this matters

Many OSM-based maps depend on the OpenMapTiles schema. If a tag is not included here, it effectively disappears from a large part of the mapping ecosystem.

Including `shop=farm` would help make local producers visible on modern vector maps.

Thank you for considering this addition.

yannisgu:
Fixes openmaptiles/openmaptiles#1796


TomPohys (approved):
Awesome! Thanks a lot!

PR #1811, opened 2026-07-29
Minor rail on bridges

### LINKED ISSUE #1788: No rail on bridges
There are no rails on bridges:

Image

These can be even verified on the OSM main website: https://www.openstreetmap.org/#map=19/49.999879/8.259534&layers=V

Using other styles, it looks correct with rails: https://www.openstreetmap.org/#map=19/49.999879/8.259534

I found a fix by removing two lines containing **bridge** from layers/transportation/style.json:

https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/style.json#L5756
https://github.com/openmaptiles/openmaptiles/blob/master/layers/transportation/style.json#L5801

Image

but I'm not sure about side effects.

If this accepted as fix, I can provide an PR.

petraduriancikova:
Solves https://github.com/openmaptiles/openmaptiles/issues/1788

- Adds two new layers for Minor rail on bridges
- This is to ensure proper rendering on top of the road network and also to comply with the layers of "Major rail bridge" and "Major rail bridge hatching" that are already in style and were created in a similar way
- Order numbers fixed

### DIFF
--- a/layers/aerodrome_label/style.json
+++ b/layers/aerodrome_label/style.json
@@ -71,7 +71,7 @@
         "text-halo-color": "hsla(0, 0%, 100%, 0.8)",
         "text-halo-width": 1
       },
-      "order": 240
+      "order": 242
     }
   ]
 }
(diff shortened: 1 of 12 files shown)


TomPohys (approved):
Awesome! Thanks a lot!

